PR #219, opened 2013-12-06
BUG FIX: FFMPEG Producer. When using LOABG with AUTO, and a SEEK or LENGTH

TomKaltz:
The frame count calculation was incorrect and could result in 0.

http://casparcg.com/forum/viewtopic.php?f=3&t=1609


TomKaltz:
@HellGore I am working on porting this PR and #239 to 2.1.0 branch.  Let me know if it's in progress on your end and I will wait.

PR #253, opened 2014-02-27
Allow CasparCG compilation with VS++ 2010 Express

gfto:
Without these patches CasparCG can't be built using Visual Studio 2010 Express.


maceyldn:
Not required. Sorry

gfto:
Since this would not be merged, at least the CasparCG wiki should be updated to specify that VS 2013 (?) is required for compilation and that VS 2010 is not supported, so we don't waste peoples time.

PR #254, opened 2014-02-27
Consumer improvements (enables streaming output!)

gfto:
These patches enable disk consumer to act like a streaming output consumer. Afterr the patches are applied the following works as expected (adds stream output with the chosen settings):

> ADD 1 FILE output.ts -f mpegts -o udp://239.78.78.78:5000?pkt_size=1316&reuse=1 -vcodec libx264 -crf 24 -preset ultrafast -tune fastdecode -b:v 512k -s cif -r 25 -acodec aac -b:a 96k

The extra settings needed can also be added in the configuration file using <extra-params></extra-params>, for example to have streaming output by default add this in caspar.config:

      <consumers>
        <file>
          <path>output.ts</path>
          <codec>libx264</codec>
          <separate-key>false</separate-key>
          <extra-params>-f mpegts -o udp://239.78.78.78:5000?pkt_size=1316&amp;ttl=1&amp;reuse=1 -vcodec libx264 -crf 24 -preset ultrafast -tune fastdecode -b:v 512k -s cif -aspect 16:9 -acodec aac -b:a 96k</extra-params>
        </file>
      </consumers>

### DIFF
--- a/modules/ffmpeg/consumer/ffmpeg_consumer.cpp
+++ b/modules/ffmpeg/consumer/ffmpeg_consumer.cpp
@@ -75,39 +75,6 @@ extern "C"
 
 namespace caspar { namespace ffmpeg {
 	
-int av_opt_set(void *obj, const char *name, const char *val, int search_flags)
-{
-	AVClass* av_class = *(AVClass**)obj;
-
-	if((strcmp(name, "pix_fmt") == 0 || strcmp(name, "pixel_format") == 0) && strcmp(av_class->class_name, "AVCodecContext") == 0)
-	{
-		AVCodecContext* c = (AVCodecContext*)obj;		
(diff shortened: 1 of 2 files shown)


maceyldn:
@gfto  What features are in here that aren't in #275 ?
Could you elaborate..

gfto:
One thing is that it allows the streaming output to be configured using the config file, also more ffmpeg options are parsed. #275 is may be generic solution but it doesn't support everything that this simple patches add. #275 is more like complete rewrite of the ffmpeg producer.

TomKaltz:
Features for the 2.0.x branch have been locked.  All further development is focused on the 2.1.0 git branch.  All future pull requests should be submitted against the 2.1.0 branch.

Issue #264, opened 2014-03-05
Playing progressive content on PAL channel with 2.0.7beta is buggy


gfto:
Guys, I think I've found a problem in Caspar 2.0.7beta. I'm trying to play 1080p50 content on SD PAL channel (ignore the obvious idiocy for now, I'm just testing).

In 2.0.4 the content is played perfectly, Caspar OSC reports 25fps output and increases current frame by 25 each second, the SEEK command is working as expected.

In 2.0.7beta the content is played jerky (like every second frame is missed, the audio is fine), Caspar OSC reports 50fps output and increases current frame by 50 each second. Also requesting Caspar to play 1000 frames (which should be 20 seconds) it plays only 10 seconds. The SEEK command seeks to entirely different position that 2.0.4 (and it is the wrong position).

The content I'm using is 1080p version of this: https://www.youtube.com/watch?v=Cj6ho1-G6tw

Here are logs from 2.0.4 and 2.0.7 playing the same content:

2.0.4 (played the material, waited to be over and then issued clear 1, notice that full 10 seconds were played):

```
[2014-03-05 14:15:53.393] [04BE4E30] [info]    ############################################################################
[2014-03-05 14:15:53.394] [04BE4E30] [info]    CasparCG Server is distributed by the Swedish Broadcasting Corporation (SVT)
[2014-03-05 14:15:53.395] [04BE4E30] [info]    under the GNU General Public License GPLv3 or higher.
[2014-03-05 14:15:53.395] [04BE4E30] [info]    Please see LICENSE.TXT for details.
[2014-03-05 14:15:53.395] [04BE4E30] [info]    http://www.casparcg.com/
[2014-03-05 14:15:53.395] [04BE4E30] [info]    ############################################################################
[2014-03-05 14:15:53.395] [04BE4E30] [info]    Starting CasparCG Video and Graphics Playout Server 2.0.4.60e4e25 Stable
[2014-03-05 14:15:53.395] [04BE4E30] [info]    on Windows 7 Ultimate 
[2014-03-05 14:15:53.395] [04BE4E30] [info]    Intel(R) Xeon(R) CPU E3-1230 V2 @ 3.30GHz Physical Threads: 8
[2014-03-05 14:15:53.395] [04BE4E30] [info]    X9SAE
[2014-03-05 14:15:53.442] [04BE4E30] [info]    Decklink 9.8
[2014-03-05 14:15:53.507] [04BE4E30] [info]     - DeckLink Mini Monitor [1]
[...]

PR #265, opened 2014-03-06
Fix regression in playing progressive clips.

gfto:
This fixes regression introduced in a95f6d84f. The problem is
that using avg_frame_rate causes fps to double for progressive
clips which is not what we want.

This patch removes the only place where avg_frame_rate was used
and fixes issue #264.


maceyldn:

gfto:
Any news on that? It is a major regression.

ronag:
the old ffmpeg stuff is not maintained and is being removed in 2.2.0.

PR #309, opened 2014-10-01
WASP3D  PRODUCER

WASP3D:
These are the first committed changes to incorporate WASP3D producer for CasparCG.


maceyldn:
@HellGore, Could we do whats required to give Manish an update please? If we need to do a code review could yourself or @ronag do it.

Cheers
Andy

WASP3D:
Hello HellGore,

We have updated our code and  sync the changes with master branch. Please guide us regarding merge with master branch.

 https://github.com/WASP3D/Server.

 We have done code review and made the required changes to improve code quality. We would be happy if you can update us regarding any more changes to be done or provide any guidelines for merge with master branch.


Regards
Manish Narang

elanvrt:
Hello,

I finally solved a nasty problem with audio mixing that has been haunting me for years.

When mixing sounds near 0dB, integer overflow occured, a very ugly crackling sound that made our sound engineers go mad at me.

This fix adds proper clipping instead of allowing overflow to happen.
It was already fixed in the 2.1 branch, but I really need this in 2.0.7 because 2.1 is far from production-ready (big audio troubles with Bluefish cards...) 
I adapted the lambda function in boost::range::transform, based on the code in 2.1 
Works fine with stereo sounds.

Code is here: (lines 247 -> 251)
https://github.com/elanvrt/Server-1/blob/master/core/mixer/audio/audio_mixer.cpp

It would be nice to merge this in the master branch, so everyone can benefit.

Regards,
Chris

HellGore:
@elanvrt what has this got to do with the WASP3D producer?

Furthermore isn't it better to simply backport relevent parts from https://github.com/CasparCG/Server/commit/60ebdf70b04e380ea273fae9f8c8e19f5fc0addd ?

elanvrt:
Hi Hellgore,

I am not experienced with github (and have nobody to give me some advice) and this is the first time I actually made a fork, so I guess i didn't use the correct procedure. :-/

No, it has nothing to do with WASP3D. Since there are no more pull requests allowed on this branch, I just clicked on this "view #309" button and this thread showed up. I found no better way to get it under attention since my commit on my own branch would probably go unnoticed by anyone. 

Yes I based my fix on the code in 60ebdf7, but it did not compile on the vs2010 I used for this version. Got this "error C3499: a lambda that has been specified to have a void return type cannot return a value", so to resolve this I squeezed it in a single return statement. 

I never used lambda functions, so I was a bit lost with that kind of syntax.
I finally found that I should have explicitly added the return type, so this compiles fine. Probably a bit more efficient: 
```
	boost::range::transform(result_ps, std::back_inserter(result), [&](double sample) -> int32_t
	{
		if (sample > max_amplitude)
		{
			return max_amplitude;
		}
		else if (sample < min_amplitude)
		{
			return min_amplitude;
		}
		else
			return static_cast<int32_t>(sample);
	});
```
Also the diagnostics line `graph_->set_tag(diagnostics::tag_severity::WARNING, "audio-clipping");` did not compile. I did not find it important enough to look into that graph code, so I skipped the clipping diagnostic alltogether.

Chris

HellGore:
>  I found no better way to get it under attention since my commit on my own branch would probably go unnoticed by anyone.

A separate pull will not get unnoticed. It is the correct way to contribute.

> Probably a bit more efficient:

The same efficiency, just clearer.

> I did not find it important enough to look into that graph code

I agree.

PR #317, opened 2014-10-17
fixed server terminating after pressing Ctrl+C in console

jaskie:
It's simple solution to block terminating server with Ctrl+C


maceyldn:

ronag:
@jaskie: can you rebase this on 2.2.0?

PR #318, opened 2014-10-17
Fixes image trembling when intrelaced video is paused

jaskie:
This is probably simplest solution to avoid shaking image when image is paused. When interlaced video is paused (from ffmpeg producer - when last frame is repeatedly displaying, and when layer's PAUSE command is invoked), only the 2nd (lower) field displays. This reduces vertical resolution by two, but eliminates this strange effect. 
Theoretically deinterlacing the last frame would be better solution, but is much more complex, because deinterlaced image should originate from time of second field (so this field should be unchanged, and only first field should be processed).
Not tested with lower field first (LFF) files, nor other producers than FFmpeg and blackmagic.

### DIFF
--- a/core/mixer/image/image_kernel.cpp
+++ b/core/mixer/image/image_kernel.cpp
@@ -323,7 +323,7 @@ struct image_kernel::implementation : boost::noncopyable
 
 		// Setup drawing area
 		
-		ogl_->viewport(0, 0, params.background->width(), params.background->height());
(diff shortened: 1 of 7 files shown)


maceyldn:

HellGore:
2.0.7 has been closed for new features, and this functionality actually already exists in the 2.1.0 branch

TomKaltz:
House keeping.

PR #319, opened 2014-10-17
Frame-accurate seeking files with long-GOP

jaskie:
Ensures correct seeking and A/V synchronization using packet's PTS values. It decodes all frames of video from keyframe (where file position is after avformat_seek_file) and discard if they are before requested frame number. This, of course, requires additional processing time. All calculations are in FFmpeg's AV_TIME_BASE units.

### DIFF
--- a/core/mixer/mixer.cpp
+++ b/core/mixer/mixer.cpp
@@ -72,7 +72,7 @@ public:
 		: ogl_(ogl)
 		, format_desc_(format_desc)
 	{
-		mipmapping_ = false;
+		mipmapping_ = env::properties().get(L"configuration.mixer.mipmapping_default_on", false);
 	}
 
 	void set_mipmapping(bool mipmapping)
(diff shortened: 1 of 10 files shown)


maceyldn:

jaskie:
This particular patch should be reviewed to work under 2.07. I recommend suspension - I’ll do it next month, I expect.

TomKaltz:
All work is current being done on the 2.1.0 branch.  Any chance you can take a stab at porting this PR to that branch?

@HellGore I am not quite sure about the ffmpeg producer in 2.1.0.  Would a PR like this against the 2.1.0 branch be useful?

jesperstarkar:
@ronag @5opr4ni Is this something that should be considered for the 2.1 sprint?

ronag:
@jesperstarkar: It will be fixed in the re-write.

PR #346, opened 2015-04-13
Enable remote debugging in CEF3 from the config file

waitforit:
This may not be the best way to do this, but I wanted you guys to take a look. 
Enabling remote debugging would be invaluable to anyone using the html producer .

The idea is to add the port to the config file and pass it to CEF3. 
Example config settings:
   <CefSettings>
        <remote-debugging-port>9022</remote-debugging-port>
    </CefSettings>


jesperstarkar:
This should be merged

Issue #171, opened 2013-09-16
Native AMCP over RS232/RS422 Serial Interface


premultiply:
Server process should open a configurable serial port like the TCP socket for AMCP protocol.

http://casparcg.com/forum/viewtopic.php?t=1704

yapus:
From a perspective of redundancy and constant reaction time serial interface thread on the same machine as CasparCG server is essential. This is because TCP stack in Windows7 can show "miracles" form time to time, especially when you have lot's of network interfaces from different subnets and want to play media from a shared network drive.

We're going with @HellGore 's number 2 - implement it directly in the CasparCG server.
For the constant reaction time - separate high-priority thread inside CasparCG process. 
For the redundancy - we plan to use two (or three - for the quorum rule) serial interfaces receiving almost synchronous sets of commands.

yapus:
#358 pull request contains some work on adding "serial" controller with PBUS protocol. I suggest there should be protocol "pbus" in controllers.serial, rather than making PBUS one and only default serial controller.

PR #358, opened 2015-06-23
Pbus

ronag:
Not entirely clean PR. Needs some love. But the basics are there.


TomKaltz:
@ronag Can you explain a little how PBUS feature works?  From reading the code it seems it triggers macros from a new special directory?

ronag:
Yes, you create regular text files in a "macro" directory. PBUS commands will cause the server to read in the text file and execute all the commands, There is also a special token "#WAIT" that will pause the execution of the macro.

TomKaltz:
file naming convention/mapping from PBUS register to macro?

yapus:
related to enhancement #171.
But i suggest that we not make PBUS one and only serial protocol available. Would be great if you add controllers.serial.protocol = "PBUS" configuration setting to choose PBUS protocol so other people can extend <serial> controller with other protocols

dimitry-ishenko:
In case anyone is interested, I am in the process of writing a simple PBus -> AMCP translator app:

https://github.com/dimitry-ishenko-casparcg/pbus2amcp

The only things left to do (as far as I am concerned) is the ability to save and load list of registers.

5opr4ni:
I will follow and watch.

dimitry-ishenko:
@5opr4ni it is now fairly usable. In fact you don't have to use the PBus part and can use it as a super simple CCG playout GUI:

snapshot

dimitry-ishenko:
Maybe I should post on in the forums, as I've totally hijacked this thread now :blush:

PR #360, opened 2015-06-28
Adds side-by-side fill/key to screen consumer

loveolsson:
Adds side-by-side fill/key to screen consumer for screen outputs like DualHead2Go, in surround configuration, or just for debugging.

It's enabled by adding `<sbs-key>true</sbs-key>` to screen consumer configuration. It overrides `<key-only>`

It creates a output that looks like this:
sbs_key_fill


coral:
For me this is a very welcome addition to the CasparCG server project for smaller productions on a budget or just using a MacBook Pro with dual outputs as a render.

hummelstrand:
Nice idea, good job!

elanvrt:
In the latest 2.1.0 build this feature is not present. I wonder why it is not included in the main code. I inserted the sbs code line by line (to not miss other code changes) and it compiled fine and seems to work. 
However I just dicovered that the key part always shows fully white on the screen consumer, even with all layers cleared, even when using the key-only option (never turned that on before) instead of sbs and also on pre-built and older CasparCG versions (2.0.7). Key output works correctly on Decklink and Bluefish outputs.
Seems to be a problem with my graphics card and openGL. Yes, I do use 32bpp display mode...

PR #362, opened 2015-07-16
ask for user to input the number of threads/cores for compilling when…

deedos:
the hard coded 12 threads was crashing my i5.
cheers


dimitry-ishenko:
This will break unattended builds. IMHO it would be better to do something like:
`time make -j${BUILD_PARALLEL_THREADS:-12}`

Then you can do:
`BUILD_PARALLEL_THREADS=4 ./set-variables-and-build-linux.sh`

PR #377, opened 2015-09-15
Revert "Crossdomain calls in 2.0"

ronag:
Reverts CasparCG/Server#376

Needs to be managed and tested by core team.


maceyldn:
Sorry bud! Thought you asked me to do it at IBC.

ronag:
There is a difference between doing a pull request and merging it. You shouldn't really have had access to do the merge so don't worry about it.

maceyldn:
Wheres the *like* button ;)

Zebiolo:
No one need to apologise.

The change is already in the 2.1 branch. If we never stop adding stuff to 2.0, we will never get the 2.1 out of the door.

Issue #410, opened 2016-02-01
Compiling CCG on Visual Studio 2015


elanvrt:
I finally found the time to attempt building CasparCG. No specific coding targets yet. :-)

I used the 85a7520 commit.

Since I had to install a new Visual Studio anyway, I used VS2015 version 14.0.24720.00 Update 1
After installing cmake, I could get started. I used the set-variables-and-build-windows.bat, only to see that the build failed on an error C4592.
The first one happened in core\video-format.cpp line 30

Since I'm a total noob on this code (and gnu open-source projects in general) I tried a quick 'n dirty fix.

Found some info on that error on stackoverflow.com: 
http://stackoverflow.com/questions/34013930/error-c4592-symbol-will-be-dynamically-initialized-vs2015-1-static-const-std

Okay, it looks like I should have stuck to the 2014 version. It seems to be a bug in the new VS2015 that occurs in some cases of initializer lists. 
Disabling the warning seems to be safe in this case. But where?

So I started inserting some pragma's to disable this error: `#pragma warning( disable : 4592)`
A logical place seemed to me to put one in every stdafx.h I could find, but the build kept showing the warning (which was treated as error), leading to failure.

I Had to insert the pragma in some other headers too: modules\image\consumer\image_consumer.h and modules\image\producer\image_producer.h

I think there should be a more elegant workaround, but it did the trick for me and I finally got my first home-built CasparCG server.

HellGore:
I am using Visual Studio 14.0.23107.0 D14REL.
I will revisit this issue when I have got time to potentially break my development environment by upgrading Visual Studio. At this time I can't afford that.

Julusian:
Build has moved to latest VS2017
@dotarmin

PR #415, opened 2016-02-29
Fix compilation in VS2015 version 14.0.24720.00 Update 1

drakmor:
Fixed issue #410


HellGore:
According to this: http://stackoverflow.com/questions/34013930/error-c4592-symbol-will-be-dynamically-initialized-vs2015-1-static-const-std this is a bug which Microsoft has fixed, and the fix will be available in the next update.

HellGore:
If you disable the warning in `common/compiler/vs/disable_silly_warnings.h` instead I will merge this

elanvrt:
Ah, that disable_silly_warnings.h was the place I didn't know about...

drakmor:
move fix to disable_silly_warnings.h

PR #417, opened 2016-03-02
Updated FFMPEG 2.7 -> 3.0 (only Windows)

drakmor:
I try update FFMPEG to latest and need help mirgate API to latest.

2014-08-07 - bb78903 - lsws 2.1.3 - swscale.h
  sws_getContext is not going to be removed in the future.

2014-10-15 - 17085a0 / 7ea1b34 - lavc 56.7.100 / 56.5.0 - avcodec.h
  Replace AVCodecContext.time_base used for decoding
  with AVCodecContext.framerate.

2015-10-29 - lavc 57.12.100 / 57.8.0 - avcodec.h
  xxxxxx - Deprecate av_free_packet(). Use av_packet_unref() as replacement,
           it resets the packet in a more consistent way.
  xxxxxx - Deprecate av_dup_packet(), it is a no-op for most cases.
           Use av_packet_ref() to make a non-refcounted AVPacket refcounted.
  xxxxxx - Add av_packet_alloc(), av_packet_clone(), av_packet_free().
           They match the AVFrame functions with the same name.     

2014-07-29 - 80a3a66 / 3a19405 - lavf 56.01.100 / 56.01.0 - avformat.h
  Add mime_type field to AVProbeData, which now MUST be initialized in
  order to avoid uninitialized reads of the mime_type pointer, likely
  leading to crashes.
  Typically, this means you will do 'AVProbeData pd = { 0 };' instead of
  'AVProbeData pd;'.


ronag:
Closed in favour of #636 #637

PR #427, opened 2016-04-11
Fix thumbnail saving on non-windows systems

krzyc:
Fix thumbnail saving on non-windows systems where function FreeImage_SaveU does nothing.


HellGore:
Good catch

PR #438, opened 2016-05-31
support for  UHD @ 50p and 60p

AurelienRevault:
This pull request is for supporting 50p and 60p in UHD
Plus some pragmas to allow compilation with visual studio 15
And a workaround at the initiliazation to make UHD @ 50p work everytime at startup.


ronag:
@AurelienRevault merged in https://github.com/CasparCG/Server/pull/544

PR #444, opened 2016-06-27
[ffmpeg_producer] Add relative and "from end" seeking

dimitry-ishenko:
This patch adds optional "REL" (relative) and "END" suffixes to the SEEK command to seek from the current position and from the end of the currently playing file.

So, one can do something like this:

`CALL 1-10 SEEK -30 REL`    to jump 30 frames back; or
`CALL 1-10 SEEK  60 REL`    to jump 60 frames forward

This essentially enables scrubbing. I ran the following:

    for i in `seq 1 100`; do casparcg-send "CALL 1-10 SEEK 30 REL"; done
    for i in `seq 1 100`; do casparcg-send "CALL 1-10 SEEK -30 REL"; done

and I saw that it was good. :)

Additionally, to jump to 30 seconds before the end of the video, one can do:

`CALL 1-10 SEEK 900 END` (assuming 30fps)

This is useful, when doing a run-through of an event and wanting to skip to the end of the video.

### DIFF
--- a/modules/ffmpeg/producer/ffmpeg_producer.cpp
+++ b/modules/ffmpeg/producer/ffmpeg_producer.cpp
@@ -391,7 +391,7 @@ public:
 	std::future<std::wstring> call(const std::vector<std::wstring>& params) override
 	{
 		static const boost::wregex loop_exp(LR"(LOOP\s*(?<VALUE>\d?)?)", boost::regex::icase);
-		static const boost::wregex seek_exp(LR"(SEEK\s+(?<VALUE>\d+))", boost::regex::icase);
+		static const boost::wregex seek_exp(LR"(SEEK\s+(?<VALUE>(\+|-)?\d+)(\s+(?<WHENCE>REL|END))?)", boost::regex::icase);
 		static const boost::wregex length_exp(LR"(LENGTH\s+(?<VALUE>\d+)?)", boost::regex::icase);
 		static const boost::wregex start_exp(LR"(START\\s+(?<VALUE>\\d+)?)", boost::regex::icase);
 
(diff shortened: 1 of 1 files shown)


dimitry-ishenko:
One gotcha with the SEEK END command is that it does not take into consideration the SEEK and LENGTH options given in the play command.

mauricev78:
Ok iḿ trying it out on the moment very pleased with this on first sight !

dimitry-ishenko:
This no longer works with the latest updates to the Server. :(

@HellGore I can update it, but need a bit of help understanding things inside ffmpeg_producer. I basically have 2 questions:
1. What function should I use to get the current frame number being played? I see `file_frame_number()` and `frame_number`. Is either one of these correct?
2. Is `input_.length()` the correct function to get the number of the last frame in the file being played?

HellGore:
> 1. What function should I use to get the current frame number being played? I see `file_frame_number()` and `frame_number`. Is either one of these correct?

Depends on what you want. `file_frame_number()` is for the current position in the file. `frame_numbers()` increases per frame during the lifetime of the producer, so if you are looping for example it increases indefinitely.

> 2. Is `input_.length()` the correct function to get the number of the last frame in the file being played?

No, `file_frame_number()` gets the current/last frame played in the file.

dimitry-ishenko:
@HellGore thank you! One last question: how do I get the total number of frames in the file?

HellGore:
`file_nb_frames()`

dimitry-ishenko:
@HellGore thank you sir. I've rebased the patch. Feel free to apply it any time :)

HellGore:
It works well, thank you.

Issue #407, opened 2016-01-07
ffmpeg_consumer


jlemanbeto:
Hi,
When using this command :
add 1 file folder/test20160107-1421.mov separate_key

Getting :
[2016-01-07 14:21:39.068] [6744]  [info]    [ffmpeg] Output #0, mov, to 'media/folder/test20160107-1421.mov':
[2016-01-07 14:21:39.156] [6744]  [info]    [ffmpeg] Output #0, mov, to 'media/test20160107-1421_A.mov':

Preferred to have : 'media/folder/test20160107-1421_A.mov'

Thanks for CasparCG.

Regards,

Jean

ronag:
2.2

PR #477, opened 2016-09-08
Separate Key video used to not be saved in same folder as the color video.

jlemanbeto:
Hello,

I've just isolate the commit that resolved the https://github.com/CasparCG/Server/issues/407

It works but please let me know if it's the right way to do and if it'll be fixed in the next release.

Thanks in advance

Jean


HellGore:
Fixed after refactoring of ffmpeg_consumer.

Issue #451, opened 2016-07-11
Audio decoding exception 44K for Decklink


Canigou77:
I have a question or an issue.
I do some tests with Decklink 4K Pro board and 4K files. BMD version is 16.6.8. 
I have only one CasparCG channel with dci2061p2400 and a stereo channelayout.
I've a test file : Sintel in 4K
When I used CasparCG build 228756a (8 of march) : playout is workfing fine:
[2016-07-11 08:19:17.063] [7364] [info]    ffmpeg[4Ksintel.mov|3840x1744i48.00|0/21312] [video-decoder] H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10
[2016-07-11 08:19:17.071] [7364] [info]    ffmpeg[4Ksintel.mov|3840x1744i48.00|0/21312] [audio-decoder] AAC (Advanced Audio Coding)

When I used last versions, it doesn't play:
[2016-07-11 08:21:39.641] [3680] [error]   [ffmpeg] [Parsed_abuffer_0 @ 0000000009A2A6A0] Changing frame properties on the fly is not supported.
[2016-07-11 08:21:39.689] [3680] [error]   E:\DEVS\CasparCG.2.1.0\Server-2.1.0.-juillet-2016\Server-2.1.0\modules\ffmpeg\producer\filter\audio_filter.cpp(219): Throw in function caspar::ffmpeg::audio_filter::implementation::push::<lambda_063064cb49531dd0dd163160a2a92831>::operator ()

Can you tell me why ?

For this file there is one audio stream: 
ID                                       : 2
Format                                   : AAC
Format/Info                              : Advanced Audio Codec
Format profile                           : LC
Codec ID                                 : 40
Duration                                 : 14mn 47s
Bit rate mode                            : Variable
Bit rate                                 : 181 Kbps
Maximum bit rate                         : 192 Kbps
Channel(s)                               : 2 channels
Channel positions                        : Front: L R
Sampling rate                            : 44.1 KHz
Frame rate                               : 43.066 fps (1024 spf)
Compression mode                         : Lossy
Stream size                              : 19.1 MiB (0%)
Language                                 : English
Default                                  : Yes
Alternate group                          : 1

Julusian:
I have just hit this issue too with various clips, including some native 44K and some transcoded as below.

[...]

PR #480, opened 2016-09-09
Support for playing rtmp video streams

Julusian:
This adds support for playing video streams (rtmp/http/udp etc) through ffmpeg.

Inspired by #131 

Note: Only works for streams with 48K audio, as normal video playback was failing when resampling is required (#451)

### DIFF
--- a/core/producer/frame_producer.cpp
+++ b/core/producer/frame_producer.cpp
@@ -368,22 +368,24 @@ spl::shared_ptr<core::frame_producer> frame_producer_registry::create_producer(c
 	auto& producer_factories = impl_->producer_factories;
 	auto producer = do_create_producer(dependencies, params, producer_factories);
 	auto key_producer = frame_producer::empty();
-	
-	try // to find a key file.
-	{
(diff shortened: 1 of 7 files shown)


baltedewit:
Compiling on Linux succeeds but trying to play back the following stream just gives me errrors: rtmp://vps48678.ovh.net/live/

't was running fine on 2.0.7

Julusian:
Hi @baltedewit, you appear to have missed off the end of the stream url so I cannot verify this, but I assume that the stream has 44.1K audio. 
If so then I believe that #451 is a relevent issue to you, which got me confused for a few hours when coding this.

baltedewit:
The right livestream is rtmp://vps48678.ovh.net/live/livestream

Sorry about that. I said nothing.

Julusian:
Thanks, I can see that it is not the issue I thought, but looks like some missing frame/packet in the stream. Ill see what I can do to make it be handled better.

HellGore:
This now works since the port from 2.0.7 that I just did.

PR #483, opened 2016-09-30
added delay option to decklink producer

AndreasWilladsen:
This makes CasparCG capable of delaying a decklink input by x amount of frames. In our situation it's used to get rid of a (expensive) hardware delay box.


premultiply:
Interesting, but who needs MORE delay? I can not imagine a use case for this yet.
The goal is usually too keep the delay/latency as low as possible.

baltedewit:
@premultiply hardware delay boxes are used to keep lipsync at times.

premultiply:
Yes but they are used to delay the (faster) **audio** to match the (large) video delay introduced by framesyncs and other videoprocessing.
Total delay of whole production chain should be as small as possible.

AndreasWilladsen:
@premultiply Our use case is for live sports broadcasting where the video signal has to be delayed in order to be in sync with an overlay we create that takes a little while to make (golf trajectory). Therefore we are using a hardware delay box which is rather expensive. There's no need for this box when the delay can be made directly in CCG.

premultiply:
Ok, special case.
But maybe you could increase the decklink consumer output buffering via config in that case.

AndreasWilladsen:
Our overlay is created by the html producer, so increasing the output delay wouldn't help as it would delay both the decklink- and html producer (all inputs in reality). We only need the decklink input to be delayed.

jesperstarkar:
Nice feature Andreas.

mauricev78:
Funny i am fighting the delay on the moment and you are adding it "LOL"

Can the "4" in line "140" be lowered as well ???
or is this balanced / tested out to be the best ???

grt
maurice

AndreasWilladsen:
@mauricev78 That would have to be @HellGore answering that question, as he has just increased it from 2->4.

tjeldflat:
Very nice!

HellGore:
`frame_buffer_.set_capacity(4)` does not increase the delay, as @mauricev78 and @AndreasWilladsen thinks. It only increases the headroom for when frames starts to get dropped. I raised that headroom for when for example 1080p6000 is input on a 50hz channel the `framerate_producer` might not invoke `recieve()` in the pace needed to drain `frame_buffer` enough each time to leave room for more input frames.

PR #493, opened 2016-11-02
[ffmpeg_producer] Constrain SEEK values

dimitry-ishenko:


dimitry-ishenko:
@HellGore does `file_nb_frames` function in `ffmpeg_producer` give accurate frame count? The reason I am asking is that I ran the following command:

    CALL 1-10 SEEK 0 END

which should have seeked (sought?) to the last frame and stopped the playback. However, the file froze as if the value was out of bounds.

premultiply:
Please keep in mind that maybe you can only seek to an I- or IDR-frame (GOP start) to get it decoded immediately.

dimitry-ishenko:
@premultiply good point, but I am playing DNxHD file in .mov container, so that shouldn't matter

Issue #351, opened 2015-04-27
Add information about compiling 2.1.0 source


krzyc:
I think we should add some info about compiling 2.1.0 server.
I have succesfully compiled it with VS 2013 Community but as my primary environment is linux it took me a while to figure all things out.
So why not put suggested workflow in readme.txt or at http://casparcg.com/wiki/Building_from_Source.
To generate VS 2013 project I have issued:
cmake -G "Visual Studio 12 Win64" .

HellGore:
Now there is a file called BUILDING describing this

Issue #342, opened 2015-02-26
FFMPEG audiostream bug


baltedewit:
When playing a raw mp3 audiostream I receive the following error:

```
[2015-02-26 21:05:03.596] [6024] [info]    Received message from Console: PLAY 1-10 http://radio.rtvslogo.nl:8000/live\r\n
[2015-02-26 21:05:05.150] [5136] [error]   [ffmpeg] [mp3 @ 0F356020] Header missing
[2015-02-26 21:05:05.161] [5136] [debug]   [audio-decoder] MP3 (MPEG audio layer 3) Selected channel layout STEREO
[2015-02-26 21:05:05.161] [5136] [info]    ffmpeg[live||0/0] [audio-decoder] MP3 (MPEG audio layer 3)
[2015-02-26 21:05:05.161] [5136] [info]    transition[empty=>ffmpeg[live||0/0]] Initialized.
[2015-02-26 21:05:05.161] [5136] [debug]   Executed command: PlayCommand
[2015-02-26 21:05:05.162] [5400] [info]    transition[html[file:///D:/Kleine%20schijf/Linkin%20Park%20-%20Living%20Things%20(2012)/01%20-%20Lost%20In%20The%20Echo.mp3]=>ffmpeg[live||0/0]] End Of File.
[2015-02-26 21:05:05.162] [5400] [error]   [ffmpeg] [mp3 @ 0F356020] Header missing
[2015-02-26 21:05:05.162] [5400] [error]   producer\audio\audio_decoder.cpp(135): Throw in function caspar::ffmpeg::`anonymous-namespace'::<lambda3>::operator ()
Dynamic exception type: class boost::exception_detail::clone_impl<struct caspar::ffmpeg::ffmpeg_error>
std::exception::what: Unknown exception
[struct boost::errinfo_api_function_ *] = avcodec_decode_audio4(codec_context_.get(), decoded_frame.get(), &got_frame, &pkt)
1094995529, "Unknown error"[struct caspar::tag_msg_info *] = Invalid data found when processing input
[struct caspar::tag_source_info *] = [audio_decoder]

```

The stream plays just fine in ffplay, so that doesn't appear to be the problem..

baltedewit:
Closing as this issue is resolved by the new ffmpeg producer.

Issue #216, opened 2013-12-05
crash while playing back y4m media


grmph555:
Hi,

I try to play back uncompressed yuv video without audio wrapped in y4m container.

Here are links to the files:
http://media.xiph.org/video/derf/y4m/720p5994_stockholm_ter.y4m
http://media.xiph.org/video/derf/y4m/ducks_take_off_422_720p50.y4m

These files can be played back with ffmpeg (ffmpeg-20131204-git-f65afef-win64-static)
ffplay.exe ducks_take_off_422_720p50.y4m

Nevertheless caspar Server crashes after PLAY OK. I attach the log below.
The crash happens everytime and the server application just disappears, no deadlock. Can be restarted immediately.

System:
HP Z200 Workstation
Intel Core i7 870 2.93 GHz
4GB RAM
Win 7 64 bit
GeForce 8400GS (Nividia Driver 9.18.13.1106)
Decklink Studio 2 (Decklink Desktop Video 9.7)
Flash 10.3.181.14

Why I want to play these files?
There are several widely known test sequences (foreman, mobile, riverbed) provided from SVT to compare Encoders and I want to use Caspar Server to feed different live encoders via SDI with these Sequences.

Log:
[12/04/13 13:36:27] [05EAE570] [info]    ############################################################################
[12/04/13 13:36:27] [05EAE570] [info]    CasparCG Server is distributed by the Swedish Broadcasting Corporation (SVT)
[12/04/13 13:36:27] [05EAE570] [info]    under the GNU General Public License GPLv3 or higher.
[12/04/13 13:36:27] [05EAE570] [info]    Please see LICENSE.TXT for details.
[12/04/13 13:36:27] [05EAE570] [info]    http://www.casparcg.com/
[12/04/13 13:36:27] [05EAE570] [info]    ############################################################################
[12/04/13 13:36:27] [05EAE570] [info]    Starting CasparCG Video and Graphics Playout Server 2.0.3.3590 STABLE
[12/04/13 13:36:27] [05EAE570] [info]    on Windows 7 Professional 
[12/04/13 13:36:27] [05EAE570] [info]    Intel(R) Core(TM) i7 CPU         870  @ 2.93GHz Physical Threads: 4
[12/04/13 13:36:27] [05EAE570] [info]    HP Z200 Workstation
[12/04/13 13:36:28] [05EAE570] [info]    Decklink 9.7
[12/04/13 13:36:28] [05EAE570] [info]     - DeckLink Studio 2 [1]
[...]

PR #497, opened 2016-11-05
[EXPERIMENTAL] Speed control

dimitry-ishenko:
This series of patches expands upon `CALL FRAMERATE SPEED` command and enables fast-forwarding at up to 16X speed.

When the speed is less than 2X the call is handled by the `framerate_producer` with frame blending and tween support.
When the speed is between 2X and 16X the call is handled by the `ffmpeg_producer` by discarding (`speed - 1`) frames and audio samples and only showing `speed`-th frame.

Examples:

`CALL 1-10 FRAMERATE SPEED 0.5` - slomo at 1/2X
`CALL 1-10 FRAMERATE SPEED 8` - fastfo at 8X

### DIFF
--- a/modules/ffmpeg/producer/audio/audio_decoder.cpp
+++ b/modules/ffmpeg/producer/audio/audio_decoder.cpp
@@ -59,6 +59,9 @@ struct audio_decoder::implementation : boost::noncopyable
 
 	std::queue<spl::shared_ptr<AVPacket>>	packets_;
 
+	int										discard_ = 0;
+	bool									release_ = true;
+
 	std::shared_ptr<SwrContext>				swr_				{
 																	swr_alloc_set_opts(
 																			nullptr,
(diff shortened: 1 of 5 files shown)


HellGore:
I think the `ffmpeg_producer` should not use `framerate speed` as parameter but something like `framestep` or similar. The `framerate_producer` is not for use exclusively in combination with `ffmpeg_producer` so the speed 2.0 limit seems a bit specific to this use case.

It is fine if a client needs to have a little bit of logic to distinguish between `framerate speed` and `framestep`

dimitry-ishenko:
@HellGore that makes sense. I will redo the patches accordingly.

As I've mentioned in my other post, I am planning to use seeking for both forward and reverse. Do you think it's better to implement it in `ffmpeg_producer` by calling (asynchronous) `input_.seek` function or should it be done (synchronously) from within the `ffmpeg::input` class?

I tend to think the latter would be the better approach, as it would have less delay, but I wanted to get your opinion. Thank you.

HellGore:
Start with using the existing method and see if that works.

dimitry-ishenko:
@HellGore here you go. This is using the discard method (audio and video decoders only decode every n-th frame and discard the rest). It currently only supports fastfo (going forward at high speed) at up to 16X.

Implementing higher speeds and rewind, will require seeking. Please lemme know what you think

dimitry-ishenko:
@HellGore on line [#342](https://github.com/CasparCG/Server/blob/2.1.0/modules/ffmpeg/producer/input/input.cpp#L342) in `queued_seek()` you initialize `flags` variable to `AVSEEK_FLAG_FRAME` and then set it to `AVSEEK_FLAG_BYTE` for VP6 codec on line [#351](https://github.com/CasparCG/Server/blob/2.1.0/modules/ffmpeg/producer/input/input.cpp#L351).

However, the variable doesn't get used anywhere else. Was it meant to be used in the call to `avformat_seek_file`?

dimitry-ishenko:
@HellGore another question: the `frame_number_` variable gets incremented on line [#216](https://github.com/CasparCG/Server/blob/2.1.0/modules/ffmpeg/producer/input/input.cpp#L216) after successful call to `av_read_frame`. However, it doesn't get adjusted after seek operations, so it just keeps increasing monotonically.

I was thinking of using it for frame stepping, but wasn't sure, if it serves some other purpose.

I see it being used in `is_eof` function, and if I read it correctly, there might be a bug in `frame_number_ >= length_` comparison (as the `frame_number_` is never subtracted from after seek backwards).

dimitry-ishenko:
I've gotten my questions answered, so I will close this and open another PR once I redo it using `queued_seek` function. Thank you all for reading :)

Issue #492, opened 2016-11-02
[2.1.0 Beta 1] Audio-stream overflow when seeking


dimitry-ishenko:
I have two very similar files. One about 3.5min long and the 2nd -- about 1hr 15min long. I tried the following test with both files:

1. Restart server.
2. Play file.
3. Issue SEEK 0 command

The 1st file (short) seeks to the beginning and continues to play as expected. The 2nd file (long) barfs with "audio-stream overflow" exception. (Seeking to any other point causes same behaviour.)

Attached are logs for both files.

dimitry-ishenko:
[caspar1.txt](https://github.com/CasparCG/Server/files/565452/caspar1.txt)
[caspar2.txt](https://github.com/CasparCG/Server/files/565453/caspar2.txt)

dimitry-ishenko:
@HellGore on a slightly separate note, it seems that the SEEK command does not check if the seek value is out of bounds. If you seek to a negative value or past the file end, the playback just freezes. (Version 2.0.6 ignores negative value and stops playing the file in the latter case.)

jesperstarkar:
@dimitry-ishenko I did notice the SEEKing past the boundaries of the file as well, but never got to report it. I think it will be correct to clamp the SEEKing values to the first- and last frames of the file. I can't see any other way of doing it.

dimitry-ishenko:
@jesperstarkar I've submitted PR #493

jesperstarkar:
Nice one!

dimitry-ishenko:
This is really bizarre. I ran the file through this command:

    ffmpeg -i in.mov -codec copy out.mov

which essentially just "re-containers" the file without any encoding. And seeking in the resulting file works on both servers. Tried the original file again, and it only works on 2.0.6, but still throws exception on 2.1.

The file was made in our studio on the west coast and I believe they used Avid to export it. Still it seems strange that (a) 2.0.6 handles seeking just fine; and (b) if I just play the file all the way through without seeking it works fine on both servers.

@HellGore any insight on this?

HellGore:
Is it possible for you to give me access to the files?

dimitry-ishenko:
@HellGore I thought you would ask that. Unfortunately, the file contains some content, which I cannot give out... :( And I realize this makes things almost impossible to debug. But maybe if you can guide me a bit, I can do some tests here locally?

HellGore:
[...]

PR #500, opened 2016-11-17
Increase stream buffer size limit to 120

dimitry-ishenko:
This addresses the problem I've described in issue #492


dimitry-ishenko:
This addresses the problem I've described in issue #492

dimitry-ishenko:
Here is another interesting case that I've run into that's making me rethink this.

I kept getting video-stream overflow error on one of the files. It turns out audio track on the file was shorter than video by about 10 seconds. There's copyright notice that was attached to the end of the file and that portion has no audio samples.

Both VLC and MPlayer play the file all the way (although the counter freezes), but both CasparCG 2.0.6 and 2.1.0 Beta 1 stop early without playing the copyright notice.

Issue #499, opened 2016-11-14
LENGTH command


dimitry-ishenko:
This is a continuation of my question regarding the `frame_number_` variable in PR #497. After some digging through the code, I think I understand now how it works.

This variable has to do with the `LENGTH` command (as in `PLAY ... LENGTH X` or `CALL ... LENGTH X`) and it counts how many frames have been played so far and indicates EOF after the count exceeds `X`. So, for example, if one were to issue this command:

    PLAY 1-10 sample/video SEEK 10 LENGTH 50

it will play 50 frames between 10 and 60 and then stop. This works out fine and aligns with the idea or marking IN and OUT points for a given clip, so that it plays between those 2 points.

The problem, however, arises if any sort of seeking is done during playback. For example, if after the above command another one is issued:

    CALL 1-10 SEEK 10

and executed after 40 frames have been played, it will seek back to frame 10, play 10 more frames and will stop at frame 20. While one may argue that this is by design, it is not very intuitive.

What I would like to propose is this:

1. In `PLAY ... SEEK X LENGTH Y` command rename `SEEK` to `IN`. `SEEK` will be retained until next major revision and these commands will be identical: `PLAY ... SEEK X` and `PLAY ... IN X`.

2. In `CALL ... START X` command rename `START` to `IN`. `START` will be retained until next major revision and these commands will be identical: `CALL ... START X` and `CALL ... IN X`.

3. Change meaning of the `LENGTH` command to represent time point (OUT) in the clip as a distance from the start point (IN). In other words, `PLAY ... SEEK 10 LENGTH 50` command will be guaranteed to play frames 10 to 60 regardless of any seeking.

4. Add `OUT` command to `PLAY` and `CALL` as an alternative to `LENGTH`. In other words, the following commands will be identical: `PLAY ... IN 10 LENGTH 50` and `PLAY ... IN 10 OUT 60`, as well as `CALL ... LENGTH 50` and `CALL ... OUT 60`.

HellGore:
> While one may argue that this is by design, it is not very intuitive.

I view it as a bug and not by design.

The only change needed is really your proposal point 3. I see no need to introduce the parameters IN and OUT.

mauricev78:
Well i find IN and OUT more practical then IN and LENGTH to be honest. 
This because it matches better with the production environment and the "terms" used there.
I think you normal want to play from a point to a point, not from a point for a certain time.
Next to that this saves a calculation steps / routine's on the client side

LS

[...]

PR #501, opened 2016-11-18
Add support for IN and OUT parameters in PLAY, CALL and SEEK commands

dimitry-ishenko:
This patch adds support for the following `PLAY` command syntax:

    PLAY 1-10 folder/clip IN 10 LENGTH 50
    PLAY 1-10 folder/clip IN 10 OUT 60 (equivalent to the above)

(Use of `SEEK` instead of `IN` is still supported.)

Following `SEEK` command syntax is now supported:

    CALL 1-10 SEEK 123 (seek to absolute frame 123)
    CALL 1-10 SEEK REL 10 (seek 10 frames forward from current position)
    CALL 1-10 SEEK REL -50 (seek 50 frames back from current position)
    CALL 1-10 SEEK IN (seek to IN point)
    CALL 1-10 SEEK IN 10 (seek to 10 frames after IN point)
    CALL 1-10 SEEK IN -50 (seek to 50 frames before IN point)

Additionally, `OUT` and `END` can be specified in place of `IN` to seek relative to the OUT point or the end of the file respectively.

Following `CALL` command syntax is now supported:

    CALL 1-10 IN 10 (set IN point to 10 frames)
    CALL 1-10 START 10 (still supported; equivalent to the above)
    CALL 1-10 OUT 60 (set OUT point to 60 frames)
    CALL 1-10 LENGTH 50 (set OUT point to IN + 50 frames)

As I've mentioned in Issue #499 the behaviour of the `LENGTH` parameter is now modified to be no longer affected by seeking during playback. In other words, when the `LENGTH` command is issued, it is used to calculate the OUT point relative to the current IN point. (NB: If the IN point is modified afterwards, the OUT point is not adjusted to keep the same length.)

### DIFF
--- a/modules/ffmpeg/producer/ffmpeg_producer.cpp
+++ b/modules/ffmpeg/producer/ffmpeg_producer.cpp
@@ -95,8 +95,6 @@ struct ffmpeg_producer : public core::frame_producer_base
 	std::unique_ptr<frame_muxer>						muxer_;
 
 	const boost::rational<int>							framerate_;
-	const uint32_t										start_;
-	const uint32_t										length_;
 	const bool											thumbnail_mode_;
 
 	core::draw_frame									last_frame_;
@@ -112,21 +110,18 @@ public:
 			const std::wstring& url_or_file,
 			const std::wstring& filter,
 			bool loop,
-			uint32_t start,
-			uint32_t length,
+			uint32_t in,
+			uint32_t out,
 			bool thumbnail_mode,
 			const std::wstring& custom_channel_order,
 			const ffmpeg_options& vid_params)
 		: filename_(url_or_file)
 		, frame_factory_(frame_factory)
 		, initial_logger_disabler_(temporary_enable_quiet_logging_for_thread(thumbnail_mode))
-		, input_(graph_, url_or_file, loop, start, length, thumbnail_mode, vid_params)
+		, input_(graph_, url_or_file, loop, in, out, thumbnail_mode, vid_params)
 		, framerate_(read_framerate(*input_.context(), format_desc.framerate))
-		, start_(start)
-		, length_(length)
 		, thumbnail_mode_(thumbnail_mode)
 		, last_frame_(core::draw_frame::empty())
-		, frame_number_(0)
 	{
 		graph_->set_color("frame-time", diagnostics::color(0.1f, 1.0f, 0.1f));
 		graph_->set_color("underflow", diagnostics::color(0.6f, 0.3f, 0.9f));
@@ -212,6 +207,12 @@ public:
 			CASPAR_THROW_EXCEPTION(averror_stream_not_found() << msg_info("No streams found"));
 
 		muxer_.reset(new frame_muxer(framerate_, std::move(audio_input_pads), frame_factory, format_desc, channel_layout, filter, true));
+
+		if (auto nb_frames = file_nb_frames())
+		{
+			out = std::min(out, nb_frames);
+			input_.out(out);
+		}
 	}
 
 	// frame_producer
@@ -313,12 +314,12 @@ public:
 			                   // therefore no seeking should be necessary for the first frame.
 		{
 			input_.seek(file_position > 1 ? file_position - 2: file_position).get();
-            boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
+			boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
 		}
 
 		for (int i = 0; i < NUM_RETRIES; ++i)
 		{
-            boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
+			boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
 
 			auto frame = render_frame();
 
@@ -338,7 +339,7 @@ public:
 				{
 					CASPAR_LOG(trace) << print() << L" adjusting to " << adjusted_seek;
 					input_.seek(static_cast<uint32_t>(adjusted_seek) - 1).get();
-                    boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
+					boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
 				}
 				else
 					return frame.first;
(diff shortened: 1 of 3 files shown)


modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -313,12 +316,12 @@ struct ffmpeg_producer : public core::frame_producer_base
 			                   // therefore no seeking should be necessary for the first frame.
 		{
 			input_.seek(file_position > 1 ? file_position - 2: file_position).get();
-            boost::this_thread::sleep_for(boost::chrono::milliseconds(40));
+			std::this_thread::sleep_for(std::chrono::milliseconds(40));
```
HellGore: Why have you changed this? This removes support for `boost::thread::interrupt()`.

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -405,74 +408,77 @@ struct ffmpeg_producer : public core::frame_producer_base
...
+				else if(boost::iequals(value, L"end"))
+					seek = nb_frames;
+				else
+					seek = boost::lexical_cast<int64_t>(value);
 
-			if(value < 0)
-				value = 0;
-			else if(value >= total)
-				value = total - 1;
+				if(params.size() > 2)
```
HellGore: Is it intended to be optional? If no, then it is better to let the out of range exception propagate here as it is handled by https://github.com/CasparCG/Server/blob/2.1.0/protocol/amcp/AMCPCommandQueue.cpp#L129 instead

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: OK fixed

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: Yes, I made it optional by intent. My idea was that you can simply do `CALL 1-10 SEEK OUT` to seek e.g. to OUT point.

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -405,74 +408,77 @@ struct ffmpeg_producer : public core::frame_producer_base
...
+			std::wstring cmd = params.at(0), value;
+			if(params.size() > 1) value = params.at(1);
 
-			if(boost::iequals(whence, L"REL"))
-				value = file_frame_number() + value;
-			else if(boost::iequals(whence, L"END"))
-				value = total - value;
+			if(boost::iequals(cmd, L"loop"))
+			{
+				if(value.size())
```
HellGore: Use `!value.empty()` instead, it better demonstrates intent.

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -405,74 +408,77 @@ struct ffmpeg_producer : public core::frame_producer_base
...
 	{
-		uint32_t file_nb_frames = 0;
-		file_nb_frames = std::max(file_nb_frames, video_decoder_ ? video_decoder_->nb_frames() : 0);
-		return file_nb_frames;
+		return video_decoder_ ? video_decoder_->nb_frames() : 0;
 	}
 
 	std::future<std::wstring> call(const std::vector<std::wstring>& params) override
 	{
-		static const boost::wregex loop_exp(LR"(LOOP\s*(?<VALUE>\d?)?)", boost::regex::icase);
```
HellGore: I am not too keen on using regexp either so I welcome this change.

modules/ffmpeg/producer/input/input.cpp
```diff
@@ -364,6 +358,8 @@ struct input::implementation : boost::noncopyable
 			static_cast<int64_t>((target / fps * stream->time_base.den) / stream->time_base.num),
 			std::numeric_limits<int64_t>::max(),
 			0), print());
+		
+		file_frame_number_ = target;
```
HellGore: Good catch

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: OK, makes sense

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -405,74 +408,77 @@ struct ffmpeg_producer : public core::frame_producer_base
...
-		boost::wsmatch what;
-		if(boost::regex_match(param, what, loop_exp))
-		{
-			auto value = what["VALUE"].str();
-			if (!value.empty())
-				input_.loop(boost::lexical_cast<bool>(value));
-			result = boost::lexical_cast<std::wstring>(input_.loop());
-		}
-		else if(boost::regex_match(param, what, seek_exp))
+		if(params.size())
```
dimitry-ishenko: Tracking

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -405,74 +408,77 @@ struct ffmpeg_producer : public core::frame_producer_base
...
 		{
-			auto value = boost::lexical_cast<int64_t>(what["VALUE"].str());
-			auto whence = what["WHENCE"].str();
-			auto total = file_nb_frames();
+			std::wstring cmd = params.at(0), value;
+			if(params.size() > 1) value = params.at(1);
```
HellGore: I prefer to always have a space after keywords and you should do the assignment on a new line.

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -608,27 +612,28 @@ struct ffmpeg_producer : public core::frame_producer_base
 void describe_producer(core::help_sink& sink, const core::help_repository& repo)
 {
 	sink.short_description(L"A producer for playing media files supported by FFmpeg.");
-	sink.syntax(L"[clip,url:string] {[loop:LOOP]} {SEEK [start:int]} {LENGTH [start:int]} {FILTER [filter:string]} {CHANNEL_LAYOUT [channel_layout:string]}");
+	sink.syntax(L"[clip,url:string] {[loop:LOOP]} {IN [in:int]} {OUT [out:int] | LENGTH [length:int]} {FILTER [filter:string]} {CHANNEL_LAYOUT [channel_layout:string]}");
```
HellGore: `{IN [in:int]}` should be changed to `{IN,SEEK [in:int]}`

modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -405,74 +408,77 @@ struct ffmpeg_producer : public core::frame_producer_base
...
-				input_.loop(boost::lexical_cast<bool>(value));
-			result = boost::lexical_cast<std::wstring>(input_.loop());
-		}
-		else if(boost::regex_match(param, what, seek_exp))
+		if(params.size())
 		{
-			auto value = boost::lexical_cast<int64_t>(what["VALUE"].str());
-			auto whence = what["WHENCE"].str();
-			auto total = file_nb_frames();
+			std::wstring cmd = params.at(0), value;
```
dimitry-ishenko: Fixed

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: Fixed

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: I figured. I've fixed it everywhere and will resubmit in a minute

dimitry-ishenko:
All fixed and resubmitted

dimitry-ishenko:
@HellGore in case you've missed it, I had made changes based on your review and resubmitted the patch

dimitry-ishenko:
@HellGore since you are on a roll, any chance of this being commited? :)

jesperstarkar:
Oh, it's hard to keep up with all the great changes these days! 😄

5opr4ni:
Keep it rolling

dimitry-ishenko:
Thank you

PR #507, opened 2016-11-23
[EXPERIMENTAL] Speed control (v2)

dimitry-ishenko:
This is version 2 of the speed control patches. This version adds `CALL ... FRAMESTEP X` command to the ffmpeg producer and works independent of the `CALL ... FRAMERATE SPEED` command.

Usage examples:

    CALL 1-10 FRAMESTEP 16 (fastfo at 16X)
    CALL 1-10 FRAMESTEP -64 (rewind at 64X)
    CALL 1-10 FRAMESTEP 1 (restore normal 1X playback)

    CALL 1-10 FRAMESTEP -1
    CALL 1-10 FRAMERATE SPEED 0.5 (slomo rewind at 0.5X)

This version uses `queued_seek()` function and is capable of achieving high speed fast forward and rewind (I've tested up to 64X).

TODO: Flush audio and video buffers on speed changes.

NOTE: 1X rewind (`CALL 1-10 FRAMESTEP -1`) and slomo rewind (combined with eg, `CALL 1-10 FRAMERATE SPEED 0.5`) doesn't look smooth.

NOTE: These patches include my IN/OUT patch (PR #501) currently under review by @HellGore.


modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -637,7 +652,8 @@ void describe_producer(core::help_sink& sink, const core::help_repository& repo)
 	sink.example(L">> PLAY 1-10 v4l2:///dev/video0", L"to use a web camera as video input on Linux.");
 	sink.para()->text(L"The FFmpeg producer also supports changing some of the settings via ")->code(L"CALL")->text(L":");
 	sink.example(L">> CALL 1-10 LOOP 1");
-	sink.example(L">> CALL 1-10 START 10");
+	sink.example(L">> CALL 1-10 IN 10");
+	sink.example(L">> CALL 1-10 OUT 60");
 	sink.example(L">> CALL 1-10 LENGTH 50");
 	sink.example(L">> CALL 1-10 SEEK 30");
 	core::describe_framerate_producer(sink);
```
HellGore: After `framerate_producer` has appended its help you could have some combination of `framerate speed` and `framestep` examples. You can check with `HELP producer ffmpeg producer` to see if they combine well.

HellGore:
> NOTE: 1X rewind (CALL 1-10 FRAMESTEP -1) and slomo rewind (combined with eg, CALL 1-10 FRAMERATE SPEED 0.5) doesn't look smooth.

I guess it doesn't look smooth without `framerate speed 0.5` either?

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: No problem, I will do it in my original PR #501.
I should have waited, really, with this PR until #501 is merged, as this PR depends on #501. I will tweak my patch and once #501 is in, will rebase this PR on top of 2.1.0 branch.

HellGore: > No problem, I will do it in my original PR #501.

No, examples for `framestep` was what I was talking about here. And primarily the interaction between `framestep` and `framerate speed`. Those descriptions should not be in #501.

dimitry-ishenko:
@HellGore yeah, -1X by itself also doesn't look good. Probably because seek takes too long. (IIRC I was seeing underflow warnings in the log.)

dimitry-ishenko:
I was thinking, maybe the `ffmpeg_producer` should keep a large queue of un-decoded packets, so they can be quickly passed to the decoders if/when the speed/direction changes. Maybe like 128 packets in each direction (or 256 packets total). With Avid DNxHR 444 at 4096x2160 10-bits 4:4:4 (3730Mb/s) 256 packets will require about 2GB of RAM, which is not too bad.

@HellGore if you have any ideas/directions, please let me know. I will try to tackle this, when I get a moment.

modules/ffmpeg/producer/ffmpeg_producer.cpp
dimitry-ishenko: Oh, I see! Yeah, I can do that.

HellGore:
Are you having problems with an SSD or a spinning disk? Backwards seeking will be much worse on a spinning drive.

Regarding 2GB of memory I don't think it is acceptable without the user explicitly saying that they intend to rewind. That behavior could never become the default, since a user might have a large number of FFmpeg producers on many layers/channels at the same time.

Furthermore I urge you to find the combination of container and video codec that is best for backwards seeking, it might not be dnxhd as you have tried with.

dimitry-ishenko:
We've got RAID-0 (Linux RAID) striped across 3 SanDisk SSD drives. Supposed to be very fast. But now I remember some time in the past we've had problems playing 4 layers (normal speed). At the time I've assumed it was the graphics card that was bottlenecking. But now that you mention it, I will do some tests to make sure it is not the RAID, which is the problem.

As regards to 2GB limit, I was being overly pessimistic in my calculations. You will need 2GB per stream, if you are playing 4K video with 10-bit color and 4:4:4 chroma sampling. We are currently doing 1080i60 in our facility and the files are 440Mb/s (vs 3730Mb/s for above). This is almost 1/10th in size and would require somewhere around 200-300MB per stream.

Also, the number of packets in the queue can be made user settable in the config file or something. Just throwing some ideas out there.

PR #502, opened 2016-11-20
Linux hdv dv camera support

walterav1984:
To complement the v4l2 patch https://github.com/CasparCG/Server/commit/c4e32e6af54000d4ef22b74e4c29529151f695e1 by adding (H)DV FireWire 1394 video device support.

Since I couldn't find a good place to describe the ffmpeg changes, I just added it to the dependencies64/ffmpeg/README.txt see comments about "--enable-libiec61883" and required packages for compiling ffmpeg.

Hope this is in time for the 2.1.0 beta 2?

PR #546, opened 2017-02-01
linux DV/HDV video device support

walterav1984:
Add ffmpeg linux build dependencies and build flag for DV/HDV video devices using iec61883.
Add ffmpeg producer support using the same URL parameter as v4l2 instead use iec61883 and simple usage example.

The old pull request #502 can/will be dropped if this is committed!


walterav1984:
Thanks!

Issue #428, opened 2016-04-17
Mac OSX build


7immy:
I've recently noticed that CasparCG can be built for linux Ubuntu machines. Seeing as Mac OSX is based off of linux I was wondering how likely it would be to be able to build CasparCG for a Mac?

I was just going to try it myself but I have next to no experience on compiling applications so was hoping I could get some advice on whether this is theoretically possible or not.

Cheers :)

7immy:
So I decided to have a quick go. I didn't really know what I was doing but I did this:

<img alt="screen shot 2016-04-17 at 13 46 16" src="https://cloud.githubusercontent.com/assets/3194029/14587056/1e993600-04a3-11e6-9b44-9fa400a53005.png">

Then in the 'build' folder I ran "make -j8".

This is a dump of the failed output:
http://pastebin.com/brqPPfKE

ronag:
This won't work out of the box... OSX is based on Unix not Linux.

7immy:
Ah well there we go then.

Would you know if there's a lot left out of the box or if its a case of including/ swapping in some packages and config files.

dimitry-ishenko:
Not that it helps in your effort, but here is a good read on history of Mac OS: http://unix.stackexchange.com/a/713
Theoretically, you should be able to build it.
- First, you should use the build scripts in the build-scripts directory and not run CMake directly
- The server comes packaged with a number of binary dependencies for Linux and Weendoze only, so
- You would need to pre-install them yourself and use this [fork](https://github.com/dimitry-ishenko-casparcg/Server/tree/system) to build against system installed dependecies.
- You should be able to install the needed dependencies through MacPorts and
- The Blackmagic SDK should be installed from from the Blackmagic Design's website
- You will probably need to tweak the system-build-linux.sh script and use it to disable some components
Good luck.

yapus:
Also CEF binaries are system dependant and Mac OS X version has to be downloaded from CEF web-site and manually added to the repository. I didn't try new CEF builds for Mac OS X yet, but when i tried last time (probably 2-3 years ago) there were some tricky parts with multi-threading inside CEF, so i ended up running single-threaded CEF inside my Mac OS X application, and that was slow

krzyc:
Mamy errors look like compiler related. I wonder whether CMAKE_COMPILER_IS_GNUCXX is set to true for Clang and can not find answer - can you create simple CMakeLists.txt with:
[...]

PR #547, opened 2017-02-02
Improved support for Bluefish hardware + new features

jamesdwise:
This updates and improves the bluefish module and hardware support, namely:
- Change the API interface to use the more recent BlueVelvetC
- Change the way the module links against the header and libs.
- Add support for the latest Bluefish hardware
- Add support for new features.

This work is conducted on behalf of Bluefish44 to ensure their 
products are correctly integrated and supported with CasparCG

Change the API interface to use the more recent BlueVelvetC:
	Created a BVCWrapper class that uses the Latest BlueVelvetC API. 
	This allows for caspar to support the most recent hardware and features.
	It also simplifies the build and link process.	

Change the way the module links against the header:
	The new BVCWrapper class just includes headers and attempts to load functions pointers from a DLL if found.
	This has allowed us to remove all files from the dependencies64/bluefish dir, reducing project size and dependencies.
	Created a new bluefish/interop dir that stores the updated and required header files.
	
Add support for the latest Bluefish hardware
	Latest verison of driver_p.h file has been used, this includes supoprt for the most recent Bluefish hardware
	There exist mutliple version of recent hardware that vary by what is supported, eg SDI i/o count, and presence of 		onboard scaler hardware or not.  
	Code has been added to support all these different configurations and variations,
 	and includes improved routing and configuration, required to support new features, below..

Add support for new features.
	The bluefish module now supports additional features: 
		Use of onboard embedded audio, 
		Selection of what output channel and type to use, ( ie. 4224, vs 422 )
		Support for onboard keyer, if supported in hardware.
	I have also updated the casparcg.config XML file, 
	Added example  of how to use the new features in the comments section.

Outstanding features and concerns.
1) The core BVCwrapper class can be used across all platforms, 
	but currently only windows is enabled and working in the code.
	Loading of function pointers still needs to be added for OSX and Linux.
	Other very minot OS specific issues will need to be addressed inside the BVCWrapper class.
2) CMakefile has been modified to ensure that bluefish module only gets built on win.
3) CMakefiles have been updated to remove the bluefish64 dir from the dependcies.
4) Further work will involve adding a Bluefish Consumer, and making the current module supported across all platforms.


hummelstrand:
Great to see Bluefish Tehnologies involved, thanks James!

modules/bluefish/CMakeLists.txt
```diff
@@ -3,18 +3,18 @@ project (bluefish)
...
 		util/blue_velvet.cpp
-
 		bluefish.cpp
 		StdAfx.cpp
 )
 set(HEADERS
+		interop/BlueDriver_p.h
+		interop/BlueHancUtils.h
+		interop/BlueVelvetC.h
+		interop/BlueVelvetCUtils.h
```
HellGore: I think these should still have been under dependencies64/bluefish/include as before since they are part of your SDK not really part of CasparCG. What do you think?

modules/bluefish/bluefish.cpp
```diff
@@ -78,21 +75,42 @@ std::vector<std::wstring> device_list()
 
 void init(core::module_dependencies dependencies)
 {
+
+	bool bCanBeConsumer = false;
```
HellGore: Please don't use camel case (only lower case with underscores) and also don't use hungarian notation.

modules/bluefish/bluefish.cpp
```diff
@@ -78,21 +75,42 @@ std::vector<std::wstring> device_list()
...
+				blue.Detach();
+			}
+		}	
 	}
 	catch(...){}
 
-	dependencies.consumer_registry->register_consumer_factory(L"Bluefish Consumer", create_consumer, describe_consumer);
-	dependencies.consumer_registry->register_preconfigured_consumer_factory(L"bluefish", create_preconfigured_consumer);
-	dependencies.system_info_provider_repo->register_system_info_provider([](boost::property_tree::wptree& info)
+	if (bCanBeConsumer)
```
HellGore: It is better if the consumer is registered even though there are no capable Bluefish cards in the computer, so for example

```
HELP CONSUMERS
```

...always includes Bluefish Consumer even though it is not possible to instantiate one based on the current system configuration.

modules/bluefish/bluefish.cpp
```diff
@@ -78,21 +75,42 @@ std::vector<std::wstring> device_list()
...
-		info.add(L"system.bluefish.version", version());
-
-		for (auto device : device_list())
-			info.add(L"system.bluefish.device", device);
-	});
+		dependencies.consumer_registry->register_consumer_factory(L"Bluefish Consumer", create_consumer, describe_consumer);
+		dependencies.consumer_registry->register_preconfigured_consumer_factory(L"bluefish", create_preconfigured_consumer);
+		dependencies.system_info_provider_repo->register_system_info_provider([](boost::property_tree::wptree& info)
+		{
+			info.add(L"system.bluefish.version", version());
```
HellGore: This is also an example of something that should be included even though Bluefish drivers are not installed on a system so that the startup log will show `bluefish.version Not found`

modules/bluefish/util/blue_velvet.h
```diff
@@ -33,25 +32,213 @@
...
 {
-	VARIANT variantValue;
-	variantValue.vt  = VT_UI4;
-	variantValue.ulVal = value;
-	return (pSdk->SetCardProperty(prop,variantValue));
+	return (pSdk->SetCardProperty32(prop, value));
 }
 
+template<typename T>
+int get_card_property(T& pSdk, ULONG prop, ULONG& value)
```
HellGore: Change to

```
int get_card_property(T& pSdk, int prop, unsigned int& value)
```

so the unnecessary lines below can be removed.

modules/bluefish/util/blue_velvet.h
```diff
@@ -33,25 +32,213 @@
...
-void blue_initialize();
-
-spl::shared_ptr<CBlueVelvet4> create_blue();
-spl::shared_ptr<CBlueVelvet4> create_blue(int device_index);
-bool is_epoch_card(CBlueVelvet4& blue);
-std::wstring get_card_desc(CBlueVelvet4& blue);
-EVideoMode get_video_mode(CBlueVelvet4& blue, const core::video_format_desc& format_desc);
 
 template<typename T>
 int set_card_property(T& pSdk, ULONG prop, ULONG value)
```
HellGore:
```
int set_card_property(T& pSdk, int prop, unsigned int value)
```

...is better.

modules/CMakeLists.txt
```diff
@@ -20,4 +19,5 @@ if (MSVC)
 endif ()
 
 add_subdirectory(image)
+add_subdirectory(reroute)
```
HellGore: What was the reason for this initialization order move?

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -56,10 +56,78 @@
...
+	}
+}
+
+bool get_videooutput_channel_routing_info_from_streamid(bluefish_hardware_output_channel streamid,
+	EEpochRoutingElements & channelSrcElement,
+	EEpochRoutingElements & sdioutputDstElement)
+{
+	switch (streamid)
+	{
+	case bluefish_hardware_output_channel::hardware_4224_channel_a:	channelSrcElement = EPOCH_SRC_OUTPUT_MEM_INTERFACE_CHA;
```
HellGore: There should be a newline directly after `case bluefish_hardware_output_channel::hardware_4224_channel_a:`

modules/bluefish/util/blue_velvet.h
```diff
@@ -33,25 +32,213 @@
...
+	bool				InitFunctionsPointers();
+	BLUEVELVETC_HANDLE	bvc;
+#if defined(_WIN32)
+	HMODULE				hModule;
+#else
+	void*				hModule;
+#endif
+
+//BlueVelvetC function pointers members
+	pFunc_bfcGetVersion bfcGetVersion;
```
HellGore: All of these function pointers should be initialized to `nullptr`. Especially `bfcFactory`.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -112,100 +276,274 @@ EVideoMode vid_fmt_from_video_format(const core::video_format& fmt)
...
+
 	return vid_fmt;
 }
 
-spl::shared_ptr<CBlueVelvet4> create_blue()
+spl::shared_ptr<BvcWrapper> create_blue()
 {
-	if(!BlueVelvetFactory4 || !encode_hanc_frame)
+	auto pWrap = new BvcWrapper();
+	if (!pWrap->IsBvcValid())
```
HellGore: This should be thrown in the constructor of `BvcWrapper` instead. In that way a `BvcWrapper` instance can never be "not valid"

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,231 @@
...
 
-void blue_hanc_initialize()
-{
+#if defined(_WIN32)
 #ifdef _DEBUG
-	std::string module_str = "BlueHancUtils64_d.dll";
+		hModule = LoadLibraryExA("BlueVelvetC64_d.dll", NULL, 0);
 #else
-	std::string module_str = "BlueHancUtils64.dll";
+		hModule = LoadLibraryExA("BlueVelvetC64.dll", NULL, 0);
```
HellGore: RAII should be used by changing the type of `BvcWrapper::hModule` to `std::shared_ptr<void>` using `FreeLibrary` as the custom deleter.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,231 @@
...
+#include <core/video_format.h>
+#include <bluefish/interop/BlueVelvetCUtils.h>
 
-#include <common/utf.h>
+#if defined(__APPLE__)
+	#include <dlfcn.h>
+#endif
 
-#include <core/video_format.h>
+#if defined(_WIN32)
```
HellGore: Throw exception using `CASPAR_THROW_EXCEPTION` instead of returning `false`.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,231 @@
...
-#ifdef _DEBUG
-	std::string module_str = "BlueVelvet64_d.dll";
-#else
-	std::string module_str = "BlueVelvet64.dll";
+	BvcWrapper::BvcWrapper()
+	{
+		bvc = nullptr;
+		hModule = nullptr;
+		InitFunctionsPointers();
+		if (bfcFactory)
```
HellGore: Currently this is undefined behaviour since `bfcFactory` is not initialized to `nullptr`. This should not be needed anyway if my other suggestions about throwing an exception during `InitFunctionsPointers()`

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,231 @@
...
+			bvc = bfcFactory();
```
HellGore: RAII should be used here as well with `bfcDestroy` as custom deleter of the `shared_ptr` constructor.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,231 @@
...
+	}
+
+	BvcWrapper::~BvcWrapper()
```
HellGore: With my other suggestions regarding RAII and `hModule` and `bvc` this destructor should be removed completely.

modules/bluefish/util/blue_velvet.h
```diff
@@ -33,25 +32,213 @@
...
+
+	BLUE_UINT32 EncodeHancFrameEx(unsigned int nCardType, struct hanc_stream_info_struct* pHancEncodeInfo, void* pAudioBuffer, unsigned int nAudioChannels, unsigned int nAudioSamples, unsigned int nSampleType, unsigned int nAudioFlags);
+	BLUE_UINT32 DecodeHancFrameEx(unsigned int nCardType, unsigned int* pHancBuffer, struct hanc_decode_struct* pHancDecodeInfo);
+
+	BLUE_UINT32 GetFrameInfoForVideoVode(const unsigned int nVideoMode, unsigned int&  nWidth, unsigned int& nHeight, unsigned int& nRate, unsigned int& bIs1001, unsigned int& bIsProgressive);
+	BLUE_UINT32 GetBytesPerFrame(EVideoMode nVideoMode, EMemoryFormat nMemoryFormat, EUpdateMethod nUpdateMethod, unsigned int& nBytesPerFrame);
+
+private:
+	bool				InitFunctionsPointers();
+	BLUEVELVETC_HANDLE	bvc;
```
HellGore: `bvc` Needs to be after `hModule` in order for the RAII I have talked about in my other comments to be done in the correct order.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,231 @@
...
+			GET_PROCADDR_FOR_FUNC(bfcUtilsGetFrameInfoForVideoMode, hModule);
+			GET_PROCADDR_FOR_FUNC(bfcUtilsGetAudioSamplesPerFrame, hModule);
+
+			bRes = true;
+		}
+
+		return bRes;
+	}
+
+	bool BvcWrapper::IsBvcValid()
```
HellGore: Should be removed completely after the changes in my other comments, `bvc` should never be `nullptr` if the `BvcWrapper` has been contstructed.

HellGore:
I have now tested the code with a "BlueFirmwareUpdate_Neutron_3o_V012.exe" firmware for 3 SDI out, and I get very bad performance with uneven channel pace. My config is:

```
  <channels>
    <channel>
      <consumers>
        <bluefish>
           <device>1</device>
           <sdi-stream>blue_422_channela</sdi-stream>
           <embedded-audio>true</embedded-audio>
        </bluefish>
      </consumers>
    </channel>
    <channel>
      <consumers>
        <bluefish>
           <device>1</device>
           <sdi-stream>blue_422_channelb</sdi-stream>
           <embedded-audio>true</embedded-audio>
        </bluefish>
      </consumers>
    </channel>
    <channel>
      <consumers>
        <bluefish>
           <device>1</device>
           <sdi-stream>blue_422_channelc</sdi-stream>
           <embedded-audio>true</embedded-audio>
        </bluefish>
      </consumers>
    </channel>
  </channels>
```

HellGore:
> Selection of what output channel and type to use, ( ie. 4224, vs 422 )

4224 here is what is known as external key in CasparCG.

> Support for onboard keyer, if supported in hardware.

I am guessing this is what we know as internal key in CasparCG.

If my above statements are correct perhaps we should use a similar terminology in the bluefish configuration to avoid confusion. Maybe something like:

```
        <bluefish>
           <device>1</device>
           <sdi-stream>a</sdi-stream>
           <keyer>external</keyer>
           <embedded-audio>true</embedded-audio>
        </bluefish>
```

...to indicate what currently would be:

```
        <bluefish>
           <device>1</device>
           <sdi-stream>blue_4224_channela</sdi-stream>
           <embedded-audio>true</embedded-audio>
        </bluefish>
```

modules/bluefish/CMakeLists.txt
jamesdwise: I was simply copying what was done with the blackmagic module - Which may or may not have been a good idea.  I am happy to move them to dependencies64/bluefish/include, As i personally feel that is probably the correct location for the files.

Personally i tend to agree that

modules/CMakeLists.txt
jamesdwise: Ahh no reason, I suspect this was left over from my (experimental) work in adding the producer code,   Will revert.

jamesdwise:
Sorry, @HellGore  I'm a bit confused by the changes you suggested to the syntax we use to configure the channels.  Could you possibly give you suggestion of how this configuration would end up looking

> <channel>
      <consumers>
        <bluefish>
           <device>1</device>
           <sdi-stream>blue_4224_channela</sdi-stream>
           <embedded-audio>true</embedded-audio>
        </bluefish>
      </consumers>
    </channel>
    <channel>
      <consumers>
        <bluefish>
           <device>1</device>
           <sdi-stream>blue_422_channelc</sdi-stream>
           <embedded-audio>true</embedded-audio>
        </bluefish>
      </consumers>
    </channel>


Are you suggesting that we use a, b, c, or d to indicate the desired channel?
and the keyer argument to determine if the channel is 422 vs 4224?

Otherwise I am working at integrating most of your other suggestions.
Thank you for being patient with me, this is my first time contributing to such a large project via github.

jesperstarkar:
If a-c are the outputs of the card, shouldn't that map to device-number by caspar convention?

jamesdwise:
maybe a quick high level overview first.

I'm not sure how familiar everyone is with the Bluefish444 hardware so perhaps it's best if I explain a bit about the hardware first.  Each bluefish device can have up to 4 inputs or 4 outputs, depending on
the hardware and what firmware is loaded.  All devices with an output channels support RGBA and are capable of doing 4224, ie. 2 SDI streams,  Fill and Key, from a single channel.
A small exception exists, that on cards with multiple output channels,
only channels A and C, can actually be configured to do the 4224 output.
(This is more a routing restriction that anything else)
Additionally some of the hardware supports a dedicated onboard hardware keyer. The Keyer in our instance works by taking the SDI input (Background) and combining this on the Bluefish444 hardware with the RGBA data (foreground) on the output channel (supplied by CasparCG). Some customers only require the 422 composited image and other require the composite and the Key data on the second SDI output.

Given that it is possible to have multiple bluefish devices in a machine,
I felt that it was appropriate to use the device number to indicate which
card / device is being used.  I think that using a <sdi-stream> a
[a|b|c|d] </sdi-stream>
is probably the best way to allow users to select which output
channel/stream they want to use.

I am unsure about the best way to provide users with the best balance of
simplicity and functionality from the rest of the config options.
My initial suggestions of the [ blue_4224_channela | blue_422_channela ]
combines the 422 vs 4224 option with the channel selection, and clearly illustrates that only certain hardware channels are capable of 4224 operation.
It also nicely reflects the way the hardware actually works, which makes
it simple for me to implement.   However I see that the simplicity of [a|b|c|d] for channel selection
would be very nice.  

I am open to suggestions of how to use the remaining options to offer
the functionality and flexibility we need.  One possibility would be to use the <keyer> tag with a number of options that allow us to configure, 422, 4224, and Hardware Keyer use.
Although combining all of these may be a bit confusing.
Could you maybe explain the meanings within CasparCG of the <Keyer> function and the arguments that you can use. For example "External" means that you will output a fill channel and a Key channel (4224) to be keyed externally. Or does this mean something else?

I do think that keeping the <hardware-keyer-audio-source>videooutputchannel
[videooutputchannel|sdivideoinput]</hardware-keyer-audio-source>
option is the correct way to go, but again i am open to changing the
actual text and value names here.

I have no Blackmagic hardware to play with so I am unfamiliar with what
the <keyer> options do on their hardware, perhaps
someone who has experience with decklink hardware can enlighten me on
what the various <keyer> options do?

modules/bluefish/util/blue_velvet.h
```diff
@@ -33,25 +32,196 @@
...
+	
+
+//	HMODULE					h_module_;
+//	BLUEVELVETC_HANDLE		bvc_;
+
+	std::shared_ptr<void>	h_module_;
+	std::shared_ptr<void>	bvc_;
+
+//BlueVelvetC function pointers members
+	pFunc_bfcGetVersion bfcGetVersion;
```
HellGore:
```
pFunc_bfcGetVersion bfcGetVersion = nullptr;
...etc
```

instead of doing all the assignments in `init_function_pointers()`

HellGore:
The Decklink cards are a bit different in that each separately addressable output on a card shows up as a separate device to the application so there each `<device />` in `casparcg.config` can be from the same physical PCIe card, but can also be from separate physical PCIe cards. I think that you should keep the meaning of `<device />` meaning actual physical card as you have done, because it better maps to your hardware.

My only suggestion is to separate the configuration about which output to use and whether to enable external key (4224 output), internal key (hardware keyer in your terminology) or no keying at all.

> Could you possibly give you suggestion of how this configuration would end up looking
> ```
>   <video-mode>1080i5000</video-mode>
>   <consumers>
>     <bluefish>
>        <device>1</device>
>        <sdi-stream>blue_4224_channela</sdi-stream>
>        <embedded-audio>true</embedded-audio>
>     </bluefish>
>   </consumers>
> </channel>
> <channel>
>   <video-mode>1080i5000</video-mode>
>   <consumers>
>     <bluefish>
>        <device>1</device>
>        <sdi-stream>blue_422_channelc</sdi-stream>
>        <embedded-audio>true</embedded-audio>
>     </bluefish>
>   </consumers>
> </channel>
> ```

Certainly:

```
  <consumers>
    <bluefish>
       <device>1</device>
       <sdi-stream>a</sdi-stream>
       <keyer>external</keyer>
       <embedded-audio>true</embedded-audio>
    </bluefish>
  </consumers>
</channel>
<channel>
  <consumers>
    <bluefish>
       <device>1</device>
       <sdi-stream>c</sdi-stream>
       <keyer>none</keyer>
       <embedded-audio>true</embedded-audio>
    </bluefish>
  </consumers>
</channel>
```

`<keyer />` should default to `external` when it is supported I think, because that is what most people expect.

modules/CMakeLists.txt
```diff
@@ -19,5 +19,4 @@ if (MSVC)
 endif ()
 
 add_subdirectory(image)
-add_subdirectory(reroute)
```
HellGore: Should not be removed just included at the position it was before.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -23,224 +23,266 @@
...
 	{
-		bvc = nullptr;
-		hModule = nullptr;
-		InitFunctionsPointers();
-		if (bfcFactory)
-			bvc = bfcFactory();
+		bvc_ = nullptr;
+		h_module_ = nullptr;
+		init_function_pointers();
+		bvc_ = std::shared_ptr<void>(bfcFactory(), bfcDestroy);
```
HellGore: Access violation occurs here because if `init_function_pointers()` returns `false` `bfcFactory` will be `nullptr`. You should instead throw the exception `if (!init_function_pointers())`. But also if `bfcFactory()` returns `nullptr`.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -23,224 +23,266 @@
...
 
-	BvcWrapper::BvcWrapper()
+	bvc_wrapper::bvc_wrapper()
 	{
-		bvc = nullptr;
-		hModule = nullptr;
-		InitFunctionsPointers();
-		if (bfcFactory)
-			bvc = bfcFactory();
+		bvc_ = nullptr;
```
HellGore: A default constructed `std::shared_ptr` is already `nullptr`. Goes for `h_module` as well

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -23,224 +23,266 @@
...
-		return bfcDetach(bvc);
+		return bfcDetach((BLUEVELVETC_HANDLE)bvc_.get());
 	}
 
-	BLUE_UINT32 BvcWrapper::QueryCardProperty32(const int iProperty, unsigned int & nValue)
+	BLUE_UINT32 bvc_wrapper::get_card_property32(const int iProperty, unsigned int & nValue)
 	{
-		if (bvc)
-			return (BLUE_UINT32)bfcQueryCardProperty32(bvc, iProperty, nValue);
+		if (bvc_)
```
HellGore: `bvc_` is never `nullptr` after the changes you have done in the constructor.

modules/bluefish/bluefish.cpp
```diff
@@ -75,41 +75,25 @@ std::vector<std::wstring> device_list()
...
-				unsigned int value;
-				blue.Attach(n);
-				blue.QueryCardProperty32(CARD_FEATURE_STREAM_INFO, value);
-				if (CARD_FEATURE_GET_SDI_OUTPUT_STREAM_COUNT(value))
-					bCanBeConsumer = true;
-
-				blue.Detach();
-			}
-		}	
+		bvc_wrapper blue = bvc_wrapper();
```
HellGore:
```
bvc_wrapper blue;
```

modules/bluefish/bluefish.cpp
```diff
@@ -57,15 +57,15 @@ std::vector<std::wstring> device_list()
 
 	try
 	{		
-		BvcWrapper blue = BvcWrapper();
+		bvc_wrapper blue = bvc_wrapper();
```
HellGore:
```
bvc_wrapper blue;
```

modules/bluefish/bluefish.cpp
```diff
@@ -42,8 +42,8 @@ std::wstring version()
 {
 	try
 	{
-		BvcWrapper blue = BvcWrapper();
-		return u16(blue.GetVersion());
+		bvc_wrapper blue = bvc_wrapper();
```
HellGore:
```
bvc_wrapper blue;
```

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -127,7 +127,7 @@ bool get_videooutput_channel_routing_info_from_streamid(bluefish_hardware_output
 
 struct bluefish_consumer : boost::noncopyable
 {
-	spl::shared_ptr<BvcWrapper>							blue_;
+	spl::shared_ptr<bvc_wrapper>							blue_;
```
HellGore: Maybe this should be held by value instead which would also remove the need for `create_blue()` entirely.

HellGore:
Could you also comment on my problem with performance? Has that been fixed in your recent commits?

jamesdwise:
I Have updated the config file and options to look like ...

>  <bluefish>
                <device>[1..]</device>
		            <sdi-stream>a[a|b|c|d] </sdi-stream>
                <embedded-audio>false [true|false]</embedded-audio>
                <channel-layout>stereo [mono|stereo|matrix|film|smpte|ebu_r123_8a|ebu_r123_8b|8ch|16ch]</channel-layout>
                <key-only>false [true|false]</key-only>
                <keyer>disabled [external|internal|disabled]</keyer>
                <internal-keyer-audio-source> videooutputchannel [videooutputchannel|sdivideoinput]</internal-keyer-audio-source>
            </bluefish>

Where, disabled gives 422 output, external gives a 4224 configuration, and internal uses the dedicated keyer on the bluefish hardware.  I have also updated the keyer audio source to better reflect that it is only for the internal keyer.
Also the sdi-stream selection is now just a,b,c,d, which makes thing s a lot clearer.

I do have a questions though...
Where is the correct place to tell users  about specific hardware restrictions, ie. the external keyer mode will only work on ch a or c?  

Lastly I have seen the performance issues you mentioned and i am looking into it.

HellGore:
> Where is the correct place to tell users about specific hardware restrictions, ie. the external keyer mode will only work on ch a or c?

```
<sdi-stream>a[a|b|c|d] (b and d never supports external key)</sdi-stream>
```

...or

```
<keyer>disabled [external|internal|disabled] (external never supported when sdi-stream is b or d)</keyer>
```

Or if you want to use the wording "only supported with a and c", but c is not supported on the card I have here for testing if I understand correctly because it only has 3 BNC in total.

jamesdwise:
Hey all,

So I am still looking into the performance issue,  and it appears (for me anyway) to go
away when I dont attach any screen consumers to the channels.
ie. only have bluefish consumers on each channel.

I am looking at a few different mechanisms to improve the performance inside the bluefish code as well.
However, it looks like caspar, despite launching the send_frame func inside an executor, 
is still waiting on some results from bluefish.     am trying to understand this mechanism better, 
but if anyone would like to share some knowledge that would be most appreciated.  
Eg. If i reduce the size of the video buffer that gets DMA'd to say 1/4 of its' original size,
everything works fine with 3 channels, of both screen and bluefish consumers.
indicating that caspar is still waiting on some result from the bluefish consumer?
Shouldn't caspar be insulated from the length of time of my DMA op? 


Additionally:
On some systems, I also get a large number of 
" [buffer] Performance warning. Buffer mapping blocked: 0.022"
messages, I am assuming this out out of my control ( or at least not related to bluefish)
and that dropping frames where i get this warning is acceptable?

HellGore:
CasparCG sends the frame to each consumer on a channel via the `send()` method and then waits for all `std::future<bool>` to return a result. This allows the consumers to keep the rendering pace of the video channel, because CasparCG itself doesn't have access to the black burst, but the consumers like bluefish and decklink do.

The problem here seems to be that sporadically the `bluefish_consumer::display_frame()` seems to take two frame durations to complete which leaves me to believe that it has something to do with `bfcWaitVideoOutputSync()` perhaps it should be done at the end of the method instead of the beginning or something.

I have the bluefish SDK on my hard drive but I can't find any API documentation about `bfcWaitVideoOutputSync()` only in examples.

If you write `diag` on the CasparCG command line you can see clearly in the graph that each bluefish consumer instance produces that when the performance problem occurs it seems like the red `sync-time` goes up to exactly a whole frame duration. `frame-time` is also close to an entire frame duration most of the time which is a bit alarming.

Theoretically `sync_time` plus `frame_time` should always add up to one frame duration, where `sync_time` is the "sleep" time needed to get the correct channel pace.

modules/bluefish/util/blue_velvet.cpp
```diff
@@ -20,67 +20,205 @@
...
-		LoadLibrary(u16(std::string(getenv("SystemDrive")) + "\\Program Files (x86)\\BlueFish444\\Driver\\" + module_str).c_str());
-	if(!module)
-		CASPAR_THROW_EXCEPTION(file_not_found() << msg_info("Could not find BlueHancUtils.dll. Required drivers are not installed."));
-	static std::shared_ptr<void> lib(module, FreeLibrary);
-	encode_hanc_frame	 = reinterpret_cast<decltype(encode_hanc_frame)>(GetProcAddress(module, "encode_hanc_frame"));
-	encode_hanc_frame_ex = reinterpret_cast<decltype(encode_hanc_frame_ex)>(GetProcAddress(module, "encode_hanc_frame_ex"));
-}
+	BLUE_UINT32 bvc_wrapper::attach(int iDeviceId)
+	{
+		return bfcAttach((BLUEVELVETC_HANDLE)bvc_.get(), iDeviceId);
```
HellGore: None of these casts to `BLUEVELVETC_HANDLE` should be needed since `bvc_.get()` returns `void*` anyway.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -277,32 +475,31 @@ struct bluefish_consumer : boost::noncopyable
...
 		}
 		else
 		{
-			blue_->system_buffer_write_async(const_cast<uint8_t*>(reserved_frames_.front()->image_data()),
+			blue_->system_buffer_write(const_cast<uint8_t*>(reserved_frames_.front()->image_data()),
 											static_cast<unsigned long>(reserved_frames_.front()->image_size()), 
-											nullptr,                 
-											BlueImage_DMABuffer(reserved_frames_.front()->id(), BLUE_DATA_IMAGE));
+											BlueImage_DMABuffer(reserved_frames_.front()->id(), BLUE_DATA_IMAGE),
+											0);
```
HellGore: It seems to be this `bfcSystemBufferWriteAsync()` call that takes almost linearly more time when more SDI outputs are used at the same time on the same card. So when I run 3 simultaneous channels this call takes 3 times as long on each channel as it does when only using one channel. They seem to be contending for some resource.

What does the `OVERLAPPED* pOverlap` parameter do? It is always set to `nullptr`. Does that mean that it is not asynchronous but blocking?

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -277,32 +475,31 @@ struct bluefish_consumer : boost::noncopyable
 						static_cast<int>(frame.audio_data().size()/channel_layout_.num_channels), 
 						static_cast<int>(channel_layout_.num_channels));
 								
-			blue_->system_buffer_write_async(const_cast<uint8_t*>(reserved_frames_.front()->image_data()), 
-											static_cast<unsigned long>(reserved_frames_.front()->image_size()), 
-											nullptr, 
-											BlueImage_HANC_DMABuffer(reserved_frames_.front()->id(), BLUE_DATA_IMAGE));
+			blue_->system_buffer_write(const_cast<uint8_t*>(reserved_frames_.front()->image_data()), 
+											static_cast<unsigned long>(reserved_frames_.front()->image_size()),  
+											BlueImage_HANC_DMABuffer(reserved_frames_.front()->id(), BLUE_DATA_IMAGE),
```
HellGore: Should `BlueImage_HANC_DMABuffer` be replaced with `BlueImage_HANC_DMABuffer` here as well or is it correct even though it is about video not sound?

jamesdwise:
Lots of good Q.'s there, 
I will try and answer them all.

1. The `BlueImage_HANC_DMABuffer`  usage is correct, we use the macro to say what buffers we are dealing with, and then the argument, ie. `BLUE_DATA_IMAGE` or `BLUE_DATA_HANC ` to indicate exactly what sort of DMA transfer is happening.
You will notice that in the case where we are not playing out audio/hanc we use the `BlueImage_DMABuffer` macro instead.

2. `bfcSystemBufferWriteAsync()`  is our DMA call, we only have a single DMA engine on the card, so it makes sense that the call takes up to 3 times as long to return, when you are doing 3 transfers.  Our driver allows us to make multiple calls from multiple threads, but due to the limitation of the hardware,  only 1 ever executes at a time. So the first call will return in the actual DMA time, but if you make 3 calls in close succession, then the 3rd one will only return once all 3 have completed. 
The `OVERLAPPED* pOverlap`  argument can be used for asynchronous transfers, by passing a `nullptr `  we force the operation to happen synchronously and blocking.  Given that the the send_frame happens on its own executor and that we still need to wait for the DMA to complete before making the present call, I didn't think it was worth the effort to make this a asynchronous operation.   - This is one of the performance improvements I am testing now

3. I have been using the `diag` tool already, but thank you for the explanation regarding what the `frame-time` and `sync-time` values are supposed to mean, it is always better than guessing.
`bfcWaitVideoOutputSync()`  is our basic Wait for sync operation it is the equivalent of the old `wait_output_video_synch` call.

I am currently looking at a couple of different ways to improve the performance,
- changing the DMA to happen Asynchronously
- modifying the code to our card in a `FIFO `mode instead of `FRAMESTORE`  which adds some latency but also buffering to the entire process.
- disconnecting the DMA from the critical path and returning to caspar asap.

Keep in mind our current hardware is only PCI-E  Gen 1 x4 lane, so the max possible bandwidth you can get to the card is between 700 and 800MB/s.  3 streams of 1080i50 RGBA is around 650MB/s  so whilst possible we are already close to the DMA limits of what is possible.  This means that when doing 3 streams we require almost the entire 40ms (actually around 35ms) available to transfer 3 channels worth  of data.   If we get the first of those frames even  5ms after the last wait for sync call completes there will always be a dropped frame.   Hopefully this provides some insight as to why you might be seeing the performance issues you are.   Of course adding some buffering and async support should help alleviate these issues.

HellGore:
It seems to me that the problem might be that calling `bfcWaitVideoOutputSync()` in times when waiting is unnecessary causing one extra frame duration to pass. Like for example if 39ms has passed since the last frame was shown and we call the `bfcWaitVideoOutputSync()` method the actual wait might slip over to the next frame duration causing 41ms to pass instead of 1ms. Is this something that might be the cause of the hickups?

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -196,30 +251,177 @@ struct bluefish_consumer : boost::noncopyable
...
+
+			keyer_control_value = VIDEO_ONBOARD_KEYER_SET_STATUS_ENABLED(keyer_control_value);
+			if (BLUE_FAIL(blue_->get_card_property32(VIDEO_INPUT_SIGNAL_VIDEO_MODE, inputVideoSignal)))
+				CASPAR_THROW_EXCEPTION(caspar_exception() << msg_info(" Failed to get video input signal mode"));
+
+			if (inputVideoSignal >= invalidVideoModeFlag)
+				keyer_control_value = VIDEO_ONBOARD_KEYER_SET_STATUS_ENABLE_OVER_BLACK(keyer_control_value);
+			else
+				keyer_control_value = VIDEO_ONBOARD_KEYER_SET_STATUS_DISABLE_OVER_BLACK(keyer_control_value);
+
```
HellGore: I have been playing with the Neutron 1i20 firmware a little bit now instead of 3o and noticed that the internal keyer expects RGB channels to be straight non multiplied. CasparCG's native output mode is to premultiply RGB channels with alpha. perhaps the `VIDEO_ONBOARD_KEYER_SET_STATUS_DATA_IS_PREMULTIPLIED` mask is needed as well?

HellGore:
> Also minor change to tell the Keyer that data is pre-multiplied when we receive it.

I don't see any changes that solves this problem. It is still multiplied twice, making the image darker when alpha is between 0.0 and 1.0

HellGore:
Now that the actual delay between a frame arriving at the consumer and it being presented on the SDI output has increased you will need to change the return value of `bluefish_consumer_proxy::buffer_depth()` accordingly. The `buffer_depth` of a consumer is how many frames worth of delay that the consumer itself adds. It is used by the `output` class in CasparCG to try to make all consumers play in sync with regards to presentation to eye and ears.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -237,15 +455,85 @@ struct bluefish_consumer : boost::noncopyable
...
-		sync_timer_.restart();
-		unsigned long n_field = 0;
-		blue_->wait_output_video_synch(UPD_FMT_FRAME, n_field);
-		graph_->set_value("sync-time", sync_timer_.elapsed()*format_desc_.fps*0.5);
-		
+		bvc_wrapper wait_b;
+		wait_b.attach(blue->device_index_);
+		EBlueVideoChannel out_vid_channel = get_bluesdk_videochannel_from_streamid(blue->device_output_channel_);
+		wait_b.set_card_property32(DEFAULT_VIDEO_OUTPUT_CHANNEL, out_vid_channel);
+		int frames_to_buffer = 3;
```
HellGore: I tried 1 here and I still got no dropped frames when running the 3o firmware with SDI A, B and C. Did you see a need for more than 1?

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -75,7 +133,14 @@ struct bluefish_consumer : boost::noncopyable
 			
 	unsigned int										vid_fmt_;
 
-	std::array<blue_dma_buffer_ptr, 4>					reserved_frames_;	
+	std::array<blue_dma_buffer_ptr, 4>					all_frames_;	
+	std::queue<blue_dma_buffer_ptr>						reserved_frames_;
```
HellGore: I think it would be much simpler and less error prone to use `tbb::concurrent_queue` or `tbb::concurrent_bounded_queue` for both `reserved_frames_` and `live_frames_`. With no need for manual mutex handling.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -75,7 +133,14 @@ struct bluefish_consumer : boost::noncopyable
...
+	std::mutex											reserved_frames_lock_;
+	std::queue<blue_dma_buffer_ptr>						live_frames_;
+	std::mutex											live_frames_lock_;
+	std::shared_ptr<std::thread>						dma_present_thread_;
+	bool												end_dma_thread_;
```
HellGore: This needs to be a `tbb::atomic<bool>` to be able to be guarantee that the change from `false` to `true` is memory visible by the DMA thread, otherwise a cached value of `false` could be read.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -237,15 +455,85 @@ struct bluefish_consumer : boost::noncopyable
...
+		wait_b.attach(blue->device_index_);
+		EBlueVideoChannel out_vid_channel = get_bluesdk_videochannel_from_streamid(blue->device_output_channel_);
+		wait_b.set_card_property32(DEFAULT_VIDEO_OUTPUT_CHANNEL, out_vid_channel);
+		int frames_to_buffer = 3;
+		unsigned long buffer_id = 0;
+		unsigned long underrun = 0;
+
+		while (!blue->end_dma_thread_)
+		{
+			if (blue->live_frames_.size() && BLUE_OK(blue->blue_->video_playback_allocate(buffer_id, underrun)))
```
HellGore: Calling `blue->live_frames_.size()` without lock on `live_frames_lock_` is not correct, right? Anyway, with the change to TBB queues instead this can be done in a different way.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -254,56 +542,59 @@ struct bluefish_consumer : boost::noncopyable
...
-		
+			live_frames_lock_.lock();
+			live_frames_.push(buf);
+			live_frames_lock_.unlock();
+
+			// start the thread if required.
+			if (dma_present_thread_ == 0)
+			{
+				end_dma_thread_ = false;
+				dma_present_thread_ = std::make_shared<std::thread>(&dma_present_thread_actual, this);
```
HellGore: Should be

```
dma_present_thread_ = std::make_shared<std::thread>([this]{dma_present_thread_actual();});
```

...is better because then you can change the `dma_present_thread_actual` function to be an instance member function instead.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -237,15 +455,85 @@ struct bluefish_consumer : boost::noncopyable
 		});
 	}
 
-	void display_frame(core::const_frame frame)
+	static void dma_present_thread_actual(void* arg)
```
HellGore: Was there ever a need for this argument to be `void*` instead of `bluefish_consumer*`? Anyway with the changes I suggested regarding the thread constructor this signature will be changed to `void dma_present_thread_actual()` with ordinary implicit `this` pointer.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -397,6 +401,10 @@ struct bluefish_consumer : boost::noncopyable
 			if (BLUE_FAIL(blue_->get_card_property32(INVALID_VIDEO_MODE_FLAG, invalidVideoModeFlag)))
 				CASPAR_THROW_EXCEPTION(caspar_exception() << msg_info(" Failed to get invalid video mode flag"));
 
+			// The bluefish HW keyer is going to pre-multiply the RGB with the A. 
+			// This setting results in the correct image coming through when using the 1080_test.tga image.
```
HellGore: It should really be `VIDEO_ONBOARD_KEYER_SET_STATUS_DATA_IS_PREMULTIPLIED` for CasparCG. Notice `IS` instead of `IS_NOT`.

>Hardware Keyer:
>Now using the correct Macro to setup the hw Keyer,
>Be aware the terminology can be confusing as the the macro refers to the source data, not the process to be applied.

The source data **is** premultiplied.

>This has been tested with the casper 1080_test.tga file and produces the correct output.
>Ie. we see the following text in white.
>"Alpha / Key correct: Premultipled
> Linear / Additive"

Unfortunately that file seems to be non useful for this test.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -237,15 +463,77 @@ struct bluefish_consumer : boost::noncopyable
...
+		wait_b.attach(device_index_);
+		EBlueVideoChannel out_vid_channel = get_bluesdk_videochannel_from_streamid(device_output_channel_);
+		wait_b.set_card_property32(DEFAULT_VIDEO_OUTPUT_CHANNEL, out_vid_channel);
+		int frames_to_buffer = BLUEFISH_HW_BUFFER_DEPTH;
+		unsigned long buffer_id = 0;
+		unsigned long underrun = 0;
+
+		while (!end_dma_thread_)
+		{
+			if (!live_frames_.empty() && BLUE_OK(blue_->video_playback_allocate(buffer_id, underrun)))
```
HellGore: Actually I am not sure if `tbb::concurrent_bounded_queue::empty` is guaranteed to give the expected result. It is based on `size()` and that doesn't seem that convincing in the api docs:
> Number of pushes minus number of pops. The result is negative if there are pop operations waiting for corresponding pushes. The result can exceed capacity() if the queue is full and there are push operations waiting for corresponding pops."

The correct way is to use the `try_pop()` method, which pops if an item is available and returns `true` otherwise `false`. You can see example uses of that method in other places in the CasparCG Server code.

modules/bluefish/consumer/bluefish_consumer.cpp
```diff
@@ -254,56 +542,54 @@ struct bluefish_consumer : boost::noncopyable
 		previous_frame_ = frame;
 
 		// Copy to local buffers
-		
-		if(!frame.image_data().empty())
+		if (!reserved_frames_.empty())
```
HellGore: Same thing here with `try_pop()` instead

HellGore:
I think this is shaping up nicely and with the few changes I proposed now I feel that it is almost ready.

Btw, how simple is it to add Linux support? The `dlopen/dlsym` stuff would be simple enough, is there any other things in your Linux SDK that differs much?

jamesdwise:
I have now applied to latest changes as per your comments.

Given the confusion I seem to be having with the correct HW keyer setting, and that apparently the test file I was using does not produce the correct output, could you possibly suggest a test that would help me to ensure i have the keyer setting correct?

As for Linux/otherOS support i was initially planning to do OSX support before Linux, but i have seen that the OSX build is not really in good working [order.....](https://github.com/CasparCG/Server/issues/428)
So perhaps i will focus on the Linux support first.
Adding Linux support should be pretty easy, we have a BlueVelvetC impl available on linux so it is just a matter of assigning all the function pointers inside the wrapper class, and everything should be good to go.

However in either case I was planning to add producer support before adding Other OS support.

HellGore:
A question I have is that when I exit CasparCG solid black is sent instead of fully transparent so when using the internal keyer I see whatever was playing in the foreground at exit fully opaque instead of only the background. The same happens with external key but then it is not full opaque but instead the last transparency when exiting CasparCG.

Is there something I can do in `BlueProps.exe` to affect what is being output when an application is not using the card? Or is the application responsible of showing transparent before exiting?

HellGore:
> Given the confusion I seem to be having with the correct HW keyer setting, and that apparently the test file I was using does not produce the correct output, could you possibly suggest a test that would help me to ensure i have the keyer setting correct?

Certainly! Given this being played as background on the SDI input:

background

...and this being played through CasparCG as the foreground using internal keyer (note seems completely white in browser but has non opaque alpha):

alpha_gradient

...the result should look like:

correct_compositing

If however the RGB values are multiplied twice with the A, which essentially is what happens when this is done incorrectly this happens:

incorrect_compositing

HellGore:
> As for Linux/otherOS support i was initially planning to do OSX support before Linux, but i have seen that the OSX build is not really in good working order.....
So perhaps i will focus on the Linux support first.

Yes, definately Linux support first. An OSX port of CasparCG is currently not on the road map at all.

jamesdwise:
Thanks for the Info about testing the various keyer settings.

As for the state of the device when exiting caspar, you can use BlueProps.exe to enable the bypass relay(s)
that will forcibly pass the video from the input, through to the input, without any interaction with the card.

Additionally when we exit caspar we display a blackframe, via the VIDEO_BLACKGENERATOR property,
which i think explains why you are seeing the results you when configuring the keyer in external mode.
Assuming that you are using an external keyer, us generating black on fill and key, will result in your external keyer displaying the background video.

This is different to when configuring the keyer as internal, we still turn the black Generator on, but this happens at the end of the chain, therefore blocking the incoming SDI signal.

Does the above explanation make sense?
If you have recommendations regarding the correct behavior on exit
I am open to modifying the code,  

One possible solution would be to,  DMA a black frame upon exit, instead of enabling the VIDEO_BLACKGENERATOR, this would result in the input being displayed on the input when internal keyer is set on, and the same / current behavior should be maintained when using an external keyer.

(upon exit the card will automatically cycle the last buffer displayed, but no audio)

Hope this helps..

HellGore:
> Additionally when we exit caspar we display a blackframe, via the VIDEO_BLACKGENERATOR property,
which i think explains why you are seeing the results you when configuring the keyer in external mode.
Assuming that you are using an external keyer, us generating black on fill and key, will result in your external keyer displaying the background video.

But it is not displaying black on the key SDI output, but rather the last transparency used when playback was active in caspar. So if i for example played a fully opaque `PLAY 1-10 RED` before exiting, the FILL SDI indeed goes black (i guess that is the `VIDEO_BLACKGENERATOR` property), but it actually still shows full white on the KEY SDI signal so the background will be blocked when the downstream keying is performed on the mixer. My suggestion here is to ensure that the KEY SDI also goes black, if that is possible.

jamesdwise:
Hey All,

Just wondering if there is anything else required from me for this pull request to be accepted?
Would you prefer I withdraw the original request and make a new one?
Or is there some more changes that you would like me to make before accepting the pull request?


- James

(hope i'm not being pushy here, i just to keep progressing)

HellGore:
No, don't withdraw it and make a new one. I will have a look on monday again. I have been busy with other stuff.

jesperstarkar:
Applause

sfahadshahzad:
any update regarding OSX support.

Fahad

HellGore:
Do you mean Linux?

HellGore:
There is no planned support for Max OS in CasparCG.

jesperstarkar:
@jamesdwise I'm testing Bluefish for the first time, and about to write a summary and send back to Bluefish about my experiences.

Would you be available/interested in implementing further on this? I'm thinking about a couple of things needed like reporting cards and capabilities in `INFO SYSTEM` calls.

I'm also thinking that a device with a given firmware/configuration should be internally mapped as different devices in CasparCG, so that a `INFO SYSTEM` would report the number of available channels/devices/streams available. This also affects how your configuration between devices and streams in the config file looks. 

Please let me know if your up for further discussions and potentially putting down some more work on this.

jamesdwise:
@jesperstarkar  Nice to hear your using bluefish, we would definitely be intersted to hear feedback regarding your experiences.

I am interested in doing more work on this, primarily input/ producer support and Linux support.
In fact I have already added input/producer support, and the code is currently on my github page, 
I am just doing some more testing before I make a pull request to get it pulled into the main branch.

Feel free to grab the branch from my page and test the input support as well.

As for you suggestions:
- Yes adding information about each devices capabilities would be a good idea,

- Currently each card/device is already addressed via a different deviceID in the config file already
(different to how the blackmagic cards work)   so i'm not exactly sure what you mean here?

Happy to discuss,
  James

jesperstarkar:
Hi James. Sorry about being a bit vague, I just wanted to get in contact and hear if you still invested time and development effort into this. 

I'll make a separate thread for discussing devices vs streams.

jesperstarkar:
Btw James - what are you thinking about implementing a proper watch dog. Not only tied to the CasparCG process, but relying on constantly getting pings from CasparCG to keep alive. Such a watch dog would enable relay bypass in case of software hangups. It would be a big advantage and sales argument for Bluefish over Decklinks - as they don't support that kind of relays.

jamesdwise:
Hi Jesper,
Implementing a proper watch dog might be possible, 
Obviously this will only be useful for people using input and output firmwares and/or the keyer.
Clearly this will be more useful once i have added the input/producer support.

We would need to configure this via an argument in the config file, and also include a setting to indicate the length of time to set before the bypass relay(s) is enabled. 

Eg. Adding a...
`<bypass> 40</bypass>`
to the config setting for a Bluefish channel, could configure things such that after 40ms of no response the card would automatically turn on the bypass relays.
a setting of 0 could or omitting the setting could be used to NOT set any watchdog.

I'm currently trying to think through the various workflow possibilities of this feature, so feel free to add in your expectations and thoughts.
Eg. What happens when Caspar crashes, and then the user wants to restart, but the bypass relay(s) have been enabled? When should we transition back to displaying the caspar output? ie. disable the bypass relay(s)

jamesdwise:
I have been looking at the possibility of a hw watchdog.
it is definitely possible.  However the length of time before it enables the bypass relay(s)
is measured in interrupts, not ms.
Would setting this to a constant of 2 missed interrupts initiates the watchdog be suitable?
or do you really want it to be a configurable amount?

jesperstarkar:
When it comes to going back from a bypass to CasparCG output, I think it should happen instantaneously, with no configuration.

I don't know what the correct number of interrupts in the threshold should be. It isn't hard to make this a config variable with `2` as the default value, is it? What is the largest number you accept?

jamesdwise:
I think that 2 would be the minimum, and the default.
I'm not even sure why you would want a larger number,
but we can support values up to at least into the thousands...

I think that if added it should be enabled by default,
and a special argument in the config file can be used to change the default number of interrupts to wait, OR to turn it off, eg. <hw-watchdog> <0>  using 0 to indicate we don't want it to run.

I'm not sure how open the casparCG community is to accepting custom config commands?

jesperstarkar:
I can only speak for myself. I mean, this value has a default, so it is opt-in to use the new parameter. And adding a new parameter either to a top-level bluefish-tag, or as a child of the consumer, will be fully backwards compatible. I don't think there's any resistance again adding a new parameter.

PR #586, opened 2017-03-27
Add *.ts extension to media whitelist

ovistr:
This is a pull request for adding a possibility to play MPEG-2 (*.ts) files from media folder.

I have tested some files with the commands used to manage these files (LOADBG, PLAY etc.), everything looks working fine.

### DIFF
--- a/modules/ffmpeg/producer/util/util.cpp
+++ b/modules/ffmpeg/producer/util/util.cpp
@@ -581,7 +581,8 @@ bool is_valid_file(const std::wstring& filename, bool only_video)
 		L".h264",
 		L".prores",
 		L".mkv",
-		L".mxf"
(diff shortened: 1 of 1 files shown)


HellGore:
Have you been having problems with `.ts` not being recognized?

ovistr:
Actually yes, files with this extension weren't recognized with LOAD(BG), PLAY or CLS commands. I think that's because they're not recognized because of not registered extension.

There was a possibility to play them in 2.0.7, like this way:

`PLAY 1-1 "file.ts"`

But this method not working for 2.1.0 version.

HellGore:
Actually `.ts` does not have to be in white list array for mpegts files to be recognized but it short circuits the function earlier.

Issue #619, opened 2017-08-18
Screen consumer ignoring <device> directive?


k4kfh:
In the 2.1.0 beta (and any builds in the 2.1.0 tree that I've tried) it seems the screen consumer ignores the ``<device>`` directive. Regardless of what number I put in, the consumer launches on the primary display. 2.0.7 works perfectly with the same configuration, so not really sure what's going on.

Here's a log:
```
[2017-08-07 12:30:00.142] [528] [info]    ############################################################################
[2017-08-07 12:30:00.143] [528] [info]    CasparCG Server is distributed by the Swedish Broadcasting Corporation (SVT)
[2017-08-07 12:30:00.143] [528] [info]    under the GNU General Public License GPLv3 or higher.
[2017-08-07 12:30:00.143] [528] [info]    Please see LICENSE.TXT for details.
[2017-08-07 12:30:00.143] [528] [info]    http://www.casparcg.com/
[2017-08-07 12:30:00.143] [528] [info]    ############################################################################
[2017-08-07 12:30:00.143] [528] [info]    Starting CasparCG Video and Graphics Playout Server 2.1.0.92e13a8 Beta 1
[2017-08-07 12:30:00.143] [528] [info]    on Windows 7 Professional Service Pack 1
[2017-08-07 12:30:00.143] [528] [info]    Intel(R) Xeon(R) CPU E5-2609 0 @ 2.40GHz Physical Threads: 8
[2017-08-07 12:30:00.143] [528] [info]    1095CF6
[2017-08-07 12:30:00.144] [528] [info]    Initialized reroute module.
[2017-08-07 12:30:00.145] [528] [info]    Initialized ffmpeg module.
[2017-08-07 12:30:00.145] [528] [info]    Initialized oal module.
[2017-08-07 12:30:00.518] [528] [info]    Initialized bluefish module.
[2017-08-07 12:30:00.518] [528] [info]    Initialized decklink module.
[2017-08-07 12:30:00.518] [528] [info]    Initialized screen module.
[2017-08-07 12:30:00.542] [528] [info]    Initialized html module.
[2017-08-07 12:30:00.542] [528] [info]    Initialized psd module.
[2017-08-07 12:30:00.673] [528] [info]    Initialized flash module.
[2017-08-07 12:30:00.673] [528] [info]    Initialized newtek module.
[2017-08-07 12:30:00.673] [528] [info]    Initialized image module.
[...]

Issue #487, opened 2016-10-17
postion non-windowed screenconsumer


mauricev78:
Hi all
I think it would be nice if we could position the screenconsumer
from the config file

i found line #219 en #220 in
/modules/screen/consumer/screen_consumer.cpp
are responsible for this

Julusian:
After poking around with this area for #619, I have some ideas about how this should be achieved, but have some points that might be worth discussing before it is implemented:

### What should happen if both a device and x,y,w,h values are specified?
Currently, the x,y,w,h values are taken from the specified screen(device) with the width and height being taken from the channel if not running fullscreen.
My thinking is that this behaviour should continue but any specified values should override the existing calculations. Perhaps the aspect ratio of the channel should be preserved if only one of width/height are specified. I am unsure on this because it might be a bit unpredictable if the environment changes.

### How should this fit into AMCP commands?
It is possible to add a screen consumer over AMCP via ```ADD 1 SCREEN 2 FULLSCREEN```. It would be good to be able to specify the x,y,w,h values over AMCP also, so how should that be formatted? Something else to consider is that even with #620, the device selection does not apply, so that number (2 in the example) is ignored. Maybe the manual position should replace that? xrandr uses the format: 1920x1080+0+0 (WxH+X+Y), so perhaps the following should be accepted ```ADD 1 SCREEN "1920x1080+0+0" BORDERLESS``` Also, when these values are specified the fullscreen and borderless modes become very similar, it would be worth considering whether there is value in having both options.

Any input on the above is appreciated, this should be pretty quick to implement once decisions have been made on the above points.

mauricev78:
I can only speak for my situation off course.
I normally try to work 1:1 in pixels.
So the channel should be runnend in its native resolution.
I all ready figured out how to make custom resolution channel and place it on its place by changing the earlier mentioned Lines. I find my self compiling the channels on a per job basis in the shop. Wich is fine but the position often needs to shift on site a pixel or 2 where this request came from.

For me X and Y are only important. But having the W and H  would be Nice.

What i understand is that there is a difference between Xserver and Windows. 
As far as i understand In Linux/Xserver the device number is a don't care as it handles screens laying on a big canvas wich is for the os one display and so it appears to CCG. I think in Windows X and Y should be in respect to the selected display. Linux will figure this out when Well documented

Julusian:
[...]

PR #620, opened 2017-08-20
Re-enable screen consumer display selection on windows

Julusian:
Fixes: https://github.com/CasparCG/Server/issues/619

Currently, this only enables screen selection on windows as libX11 does not present a method of determining the position of screens. xrandr appears to calculate by looking at the relative positions of the screens, and adding this logic to CasparCG feels wrong. An alternate solution would be to implement https://github.com/CasparCG/Server/issues/487

Some docs have been added to indicate that the property only works on windows, including a warning to the console if used on linux.

This also fixes issues:
- the size of a fullscreen consumer would use the dimensions of the primary screen when on a secondary screen
- a windowed consumer would always be positioned on the primary screen (this required removing the close button from the window)


k4kfh:
Can confirm that this works without issue on Windows.

mauricev78:
I hope this will be merged soon it will help me out on a paticular job

ls

ronag:
@Julusian PR 2.2.0?

PR #611, opened 2017-07-10
[html] Upgrade to CEF3239 (Chromium 63)

Julusian:
This upgrades the HTML producer to a more modern version of Chromium (CEF3239 / Chromium 63).

This builds on the work from @AndreasWilladsen and @sesse. Any additional tweaks or tests to run are appreciated. I have labelled this PR as a WIP as performance is not yet perfect and some frames appear to be being lost, however the same is true of the current version

Improvements:
- Newer chromium. Gives new features and makes template dev much easier at it matched current versions of chrome
- Font smoothing
- 3D CSS

Missing:
- WebGL (Pending [CEF #1006](https://bitbucket.org/chromiumembedded/cef/issues/1006))

Forum post: https://casparcg.com/forum/viewtopic.php?f=3&t=4681&p=29526

Builds: https://caspar.julusian.co.uk/cef-upgrade/

PR #644, opened 2017-10-28
[html] Allow running templates from an arbitrary url

Julusian:
Currently html templates must be stored in the templates folder, but any webpage can be played not as a template. This change allows for running templates from any url ending with .html (or any other allowed extensions defined in future)

I found this to be an annoying limitation while developing some templates that are built using webpack.


jesperstarkar:
What features are you missing, and how would you have it implemented?

I use webpack with HMR for development, that can be deployed both as CG templates and ran as HTML pages with INVOKE interfaces.

What are you missing?

Julusian:
I have done this in the past by putting just the html files in the templates folder, and making them load scripts from webpack. The issue with that is it means my source files are split and I have to remember to copy them back to my source directory to commit changes, and remember to ignore the temporarily changed script tag.

With this change, I can leave all of the html and js to be served by webpack so they are already in my source directory and have not been modified for development. Once I am finished I will most likely be compiling them down so they can go in the templates folder, so this is currently just a developer environment improvement for me.

After thinking about it, I can see some value in being able to host the templates separately from caspar anyway. If the graphics require talking to a standalone server for their data, they may as well be served up by that server too. The advantage of this is that it keeps the assets together.

jesperstarkar:
I run it in developer mode from the Webpack server, as HTML pages. Then compile/build as standalone files for CG Template usage.

jesperstarkar:
(for assets, change Webpack assets path to "./" instead of "/"...)

Julusian:
I would run them as pages, but I need them running as templates currently as I am testing them being triggered via play/next/stop over amcp as part of larger sequences.

jesperstarkar:
Oh, that's no problem. Just add play/next/stop/update as public functions, and they will function even as a page.

And, just realised this is a PR and not a issue... I'm not against this feature to be added.

silid:
I vote for this too.

ronag:
@Julusian is this included in https://github.com/CasparCG/Server/pull/611?

Julusian:
No, this is not in #611

PR #649, opened 2017-11-07
[image_consumer] Un-multiply frame colours before writing png to file

### LINKED ISSUE #648: Image consumer wrongly calculated alpha values
Semi transparent alpha values are wrongly produces in Image consumer. I think this is matter of calculating on premultiplied vs straight values.

**Example file: Text on transparent background. Notice the darkened outline effect in the anti-aliasing area around the text.**
20171107t174539

**Demo: Compositing this on top of a white solid in Photoshop (or Nuke)**. Upper half of the stage is a PRINT from CasparCG where the same text layer is compsited by MIXER on top of a white solid (inside of CasparCG).
screen shot 2017-11-07 at 14 44 33

**Example file: 10% transparent white background (still HTML producer)**
20171107t181850

**Demo: Same here. The 10% white on top of solid white should become 100% white.**
screen shot 2017-11-07 at 18 21 15

I have done the same examples as QTRLE movie clips and checked in similar comps inside of Nuke and After Effects, and the File consumer doesn't have the same problems. Which leads me to think that this is only related to Image consumer, or perhaps only the PNG handling, since @Julusian has noted similar problems in the Image producer as well: #614

Julusian:
Fixes #648

https://drive.google.com/open?id=1huOSoaoawhlPin_D_gtlac-_LMu5Qlvt

### DIFF
--- a/modules/image/consumer/image_consumer.cpp
+++ b/modules/image/consumer/image_consumer.cpp
@@ -47,6 +47,7 @@
 #include <algorithm>
 
 #include "../util/image_view.h"
+#include "image/util/image_algorithms.h"
 
 namespace caspar { namespace image {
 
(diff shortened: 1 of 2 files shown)


jesperstarkar:
Wow @Julusian that was quick! And perfect. Just testet it, 100% perfect :)

screen shot 2017-11-08 at 09 57 24
screen shot 2017-11-08 at 09 57 19

PR #666, opened 2018-01-07
[mixer] Directional blur

### LINKED ISSUE #585: Scene Producer - Blur Filter
I alreday posted this in the forum, but here is probably the better place. I played around with the new scene producer for a while. A little thing, that I noticed is missing (or I did not found it) is:

A blur filter to cast a soft shadow or similar stuff.

Julusian:
Potentially fixes #585 

This adds direction blurs to caspar. Intended to be used as motion blur. Available as MIXER BLUR and from the scene producer.

I used https://john-chapman-graphics.blogspot.co.uk/2013/01/what-is-motion-blur-motion-pictures-are.html as a reference for doing a blur in GLSL. 
I had to composite the layers into a new texture before applying the blur as otherwise if it was applied to a transformed image, it would not blur beyond the bounds of the image, or for the text producer would have some bleed from other characters in the source texture.

This may not be the most optimised of solutions as I haven't done much OpenGL development before.

### DIFF
--- a/accelerator/ogl/image/image_kernel.cpp
+++ b/accelerator/ogl/image/image_kernel.cpp
@@ -248,6 +248,7 @@ struct image_kernel::impl
 
 		shader_->use();
 
+		shader_->set("blur",			false);
 		shader_->set("post_processing",	false);
 		shader_->set("plane[0]",		texture_id::plane0);
 		shader_->set("plane[1]",		texture_id::plane1);
@@ -282,6 +283,10 @@ struct image_kernel::impl
 		else
 			shader_->set("chroma", false);
 
+		// Check if the blur post step needs to be performed
+		const int blur_x = static_cast<int>(params.transform.blur[0]);
+		const int blur_y = static_cast<int>(params.transform.blur[1]);
+		const bool has_blur = blur_x > 0 || blur_y > 0;
 
 		// Setup blend_func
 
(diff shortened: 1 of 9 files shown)


Julusian:
As this can be quite gpu intensive, I am tempted to add a quality option so that the sampling frequency can be reduced for when quality matters less or will be less noticeable. This is currently tuned to look good on a still frame, and so may be overkill for some animations.

premultiply:
Looks fine!

sesse:
Your overall approach is fine. A few comments:

* I'm unsure if you're allowed to send sampler2D objects through functions or not (NVIDIA possibly allows it, but I'd check with the standard).
* If you want more speed with heavy blurs, a standard trick is to work with a downsampled version of the image, although it adds more complexity in this case (and for directional blurs, you'd probably need to rotate before downsampling).
* You can move some work out of the loop; in particular, precalculate 1.0/(nSamples - 1) so you can multiply instead of dividing. This kind of shader is usually heavily biased towards bandwidth use, so it might be that the divide gets hidden anyway, but it will depend on the GPU.

Julusian:
Thanks for the feedback.
* As it is the same texture every time, I have removed passing it as a parameter just to be safe
* I've had another play, and reduced the sampling frequency. I can't see any difference, I'm not sure why I thought I could before. This means that I can apply a blur of (850,850) or (0,1200) with my GTX1060 being able to keep up at 85% usage. So I think performance should be ok without needing to add complexity with downsampling. Also, at a more reasonable strength of 50 (the second screenshot above) the blur uses 7% of my gpu, which I think is an acceptable amount.
* That is a good suggestion. I have pulled out as much of the maths from inside the loop as possible, so that it is just a couple of add operations.

sesse:
If you're still worried about performance, just do the blur recursively. Do 3–4 samples spread out appropriately, render to a texture, use that texture as the input to the next blur phase, etc.

mistergiantmedia:
Is there a build with mixer blur somewhere?  I want to blur a decklink input to use the changing lighting conditions in the input as add effect in caspar. So the scene in caspar changes with the lighting conditions in the studio.

Issue #640, opened 2017-10-19
AMCP heartbeat


ronag:
provide a heartbeat against the AMCP processor without spamming the logs

jesperstarkar:
Will this beat agains all connected clients, or is it a ping that clients can send to the server?

ronag:
from client to server

ronag:
the other way would also be useful both not very

premultiply:
I suggest to introduce a PING command in AMCP which will only be logged in very high log level or never be logged at all:

`PING {[token:int]}`

Where token may be any (random) value choosen by the client. It will just be returned as sent.

Example:
```
>> PING 73487965
<< 201 PING OK
<< 73487965
```

or without token:

```
>> PING
<< 202 PING OK
```

jesperstarkar:
Do we need the token? A client only gets its own responses back, doesn't it? And the REQ/RES protocol gives you the ability to track commands.

premultiply:
Yes, you are right. REQ/RES provides the same function so the simple variant is enough.

jesperstarkar:
I like the idea of the token, though. Don't mind it being added, however, just wanted to remind about REG/RES and the fact that only a single client receives responses.

jesperstarkar:
@ronag I think this can be closed?

PR #667, opened 2018-01-07
AMCP heartbeat

Julusian:
#640

Sending a PING to the server causes it to respond with PONG. Additional tokens can be added which will be returned with the PONG.
This command does not support the REQ/RES syntax as in order for it to be excluded from the logs the command has to be filtered out before that handling is applied.


premultiply:
#640

ronag:
@Julusian could you do a new PR based on 2.2.0?

Issue #456, opened 2016-07-18
Bug/unexpected behaviour: configurable AMCP commands, wrong response. 


jesperstarkar:
Mixer blend-modes are disabled by default. This bug/unexpected behaviour is valid in both default configuration and by explicitly disabling Mixer blend-modes.

If blend-modes are disabled globally, the mixer will still "set" a blend mode and respond with a positive 200-status.

If you query the layer for its blend mode it will report the blend-mode previously set.

I expect AMCP to respond negatively to setting blend-modes if these are disabled through config, and I expect AMCP not to respons a false-positive blend-mode in this case.

jesperstarkar:
Same goes for `straight_alpha_output` and `channel_grid` (channel_grid config actually not being implemented at all #455).

Julusian:
What response code would you expect for the 'failed' set? 
`600 [command] FAILED	[command] not implemented` looks to be the closest match. 
Perhaps adding `601 [command] FAILED	[command] disabled` would be better.

Would you also want them to be hidden from the help listings? Leaving them there could help users discover them, but it also may be a bit confusing as they won't work.

jesperstarkar:
601 disabled is a perfect fit, imo.

Julusian:
Ill raise a PR once 2.2.0 branch has stabilised. For now, you can see the changes at:
https://github.com/Julusian/CasparCG-Server/tree/456-amcp-disabled-commands

I ended up going with 600 as the error code as that was already being used for this purpose with thumbnail generation.
channel_grid hasnt been done as it operates on the last channel (as mentioned by hellgore in #455). Both straight_alpha_output and blend_modes now return this error though.

jesperstarkar:
Thank you Julian! Nice and important little fix! Will test it later.

PR #678, opened 2018-01-18
[protocol_amcp] Attempting to set MIXER BLEND or MIXER STRAIGHT_ALPHA_OUTPUT when disabled

Julusian:
#456

Attempting to set MIXER BLEND or MIXER STRAIGHT_ALPHA_OUTPUT when they are disabled in config now returns a 600 error, like the thumbnail commands.


Julusian:
@ronag am I right in thinking that blend modes and straight alpha can't be disabled in the config now?

ronag:
Nope. They can not.

dotarmin:
> @ronag Nope. They can not.

But they're still available? 
How do we choose between premultiplied alpha and straight alpha by this?

/Armin

ronag:
@dotarmin they should be enable by default. The reason they were possible to disable was because they used a OpenGL extension. However, since we now require OpenGL 4.5 that doesn't make sense anymore.

dotarmin:
Thanks for clearing this out Robert! 👍

jesperstarkar:
Just to clearify. It is the `mixer->straight_alpha` config that gets removed, and the `channel->straight_alpha_output` remains both as config and runtime mixer command?

Thank's for your help with this @Julusian. A shame it was a waste of your time.

Julusian:
@jesperstarkar It would appear that straight alpha output has been removed completely. https://github.com/CasparCG/server/commit/947818a041fcb4f986adbe86f9b84531e471d0c1

jesperstarkar:
Oh. That's not good. I look forward to seeing branches stabilise and start compiling a list of breaking changes to get an overview over functionality again.

ronag:
> Oh. That's not good. I look forward to seeing branches stabilise and start compiling a list of breaking changes to get an overview over functionality again.

Please do!

ronag:
> @jesperstarkar It would appear that straight alpha output has been removed completely. 947818a

It on my list to re-add. The implementation was very slow with both an extra render pass + a texture barrier which can cause problems on some setups.

jesperstarkar:
@ronag My strategy is to start from the amcp-implementation, and then dive into producers/consumers to see what functions can be invoked or config.

ronag:
@jesperstarkar yes, since a lot of people assume functionality in 2.1 that's been removed we decided to leave 2.1 as is and do a version bump.

PR #682, opened 2018-01-22
Cleanup for 2.2.0

k0d:


Julusian:
Why has the deploy stuff been removed?
I think that having some sample assets are useful to have, at least in any release versions. Is it intended to host those files elsewhere instead, or are they simply being scrapped.

Also, I believe the flash producer wont work without the templatehost files (cg20.fth.*) so removing those will likely cause some users issues

k0d:
I believe they will be hosted somewhere else. I hope to have some more sample templates etc, but they should be in a separate repository in my opinion.

Ok...I didn’t realise the .fth files are needed, but just for windows. I will add them back!

dotarmin:
> Ok...I didn’t realise the .fth files are needed, but just for windows. I will add them back!

Needed only for Windows because we cannot play flash templates on Linux.
Good job Mark!

k0d:
This pull request is now ready in my opinion.

ronag:
@k0d can u fix the last merge conflicts and then I'll merge

Julusian:
Im having some issues because shell/LiberationSans-Regular.ttf isnt committed to LFS when git says it the gitattributes says it should have been. This means my working tree is always marked as dirty and is stopping me from editing commits and making rebases harder to do
@ronag can you fix this please

Julusian:
Additionally, the windows build scripts are broken in a few ways after this PR. 
There are a couple of incorrect paths in the bat files due to them being moved which are easy to solve, but then cmake throws some errors.
```
CMake Error at CMakeModules/LegacyBuild.cmake:101 (include):
  include could not find load file:

    CMake/PrecompiledHeader.cmake
Call Stack (most recent call first):
  CMakeLists.txt:14 (INCLUDE)


CMake Error at accelerator/CMakeLists.txt:39 (add_precompiled_header):
  Unknown CMake command "add_precompiled_header".


-- Configuring incomplete, errors occurred!
```

k0d:
CMake/PrecompiledHeader.cmake has moved to CMakeModules/PrecompiledHeader.cmake

dotarmin:
> Im having some issues because shell/LiberationSans-Regular.ttf isnt committed to LFS when the gitattributes says it should have been. This means my working tree is always marked as dirty and is stopping me from editing commits and making rebases harder to do
@ronag can you fix this please

I'm having the same issues.

ronag:
Fixed

dotarmin:
Thanks @ronag!

PR #693, opened 2018-01-28
Fix building with vs2015

Julusian:
A bunch of issues I encountered when attempting to build via the scripts.

### DIFF
--- a/CMakeModules/LegacyBuild.cmake
+++ b/CMakeModules/LegacyBuild.cmake
@@ -98,7 +98,7 @@ if (POLICY CMP0045)
 	cmake_policy(SET CMP0045 OLD)
 endif ()
 
-include(CMake/PrecompiledHeader.cmake)
+include(CMakeModules/PrecompiledHeader.cmake)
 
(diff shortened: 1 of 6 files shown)


sesse:
FWIW, changing std::shared_ptr<uint8_t[]> to std::shared_ptr<uint8_t> invokes undefined behavior when the buffer is freed (attempt to call delete on a buffer allocated with new[], instead of delete[]). If you can't use the array version, you'll need to supply a custom deleter.

ronag:
@Julusian @sesse is right. It should be:

```
auto result = std::shared_ptr<uint8_t>(reinterpret_cast<uint8_t*>(malloc(format_desc.size), free);
```

Julusian:
Ah, Ill raise another PR to correct that then.

@ronag After an update there is another build issue, which I assume affects both windows and linux. You removed the definition of flatten from common/future.h, but it is still being used by ogl/image_mixer.cpp and core/producer/stage.cpp. I can readd it to solve that, unless you plan to update the usages later on?

ronag:
@Julusian yes, i pushed  something that wasn’t ready. My mistake will fix it tomorrow.

Issue #692, opened 2018-01-27
Incompatible boost version on Ubuntu 16.04/17.10


Julusian:
I am attempting to build 2.2.0 on 16.04 using the docker images, but the build looks to have been broken by the change fdcf8c32898bc60524228c3492ef1a3c3a98c51f.

The cause looks to be because Ubuntu 16.04 uses boost 1.58, but the project is relying on a header added in 1.66. Ubuntu 17.10 has the same issue as it uses boost 1.62
```
[ 25%] Building CXX object accelerator/CMakeFiles/accelerator.dir/ogl/image/image_kernel.cpp.o
In file included from /source/accelerator/ogl/image/image_kernel.cpp:31:0:
/source/accelerator/ogl/image/../util/device.h:29:37: fatal error: boost/asio/io_context.hpp: No such file or directory
compilation terminated.
accelerator/CMakeFiles/accelerator.dir/build.make:73: recipe for target 'accelerator/CMakeFiles/accelerator.dir/ogl/image/image_kernel.cpp.o' failed
make[2]: *** [accelerator/CMakeFiles/accelerator.dir/ogl/image/image_kernel.cpp.o] Error 1
```

ronag:
@k0d

k0d:
I'll upgrade to Boost 1.66 for the linux builds. I have to compile it from source, but will link statically so runtime dependencies should all still be available in the distro repositories.

ronag:
http://www.boost.org/doc/libs/1_66_0/more/getting_started/unix-variants.html

k0d:
I've compiled it before inside the Docker image....so it's quite a simple task :)

Issue #350, opened 2015-04-24
[REQUEST] HTML supporting MP4 and MP3


baltedewit:
Out of the box CEF3 does not support mp3 nor mp4 because they are not open standards.

It is pretty much a standard in broadcasting world however, so it would be very nice if support could be added. Most video editors don't support .ogg, thus are incompatible with cef3. 

This is mostly an issue because cef3 can't communicate with caspar directly, thus caspar cannot be used for video playback in a template.

premultiply:
Is this fixed now with new CEF in 2.2? Please test.

baltedewit:
@Julusian would know best.

Julusian:
It is not enabled in the official builds, so this will not be working currently.
I would like to enable it, but didn't get an answer about potential licensing issues when I raised it in the PR

baltedewit:
> but didn't get an answer about potential licensing issues when I raised it in the PR

@dotarmin what do you think about enabling this? we already enable these in ffmpeg, so I don't see any issues license wise. technically, I don't think we need to change anything in the code, just a different cef binary.

Julusian:
> we already enable these in ffmpeg, so I don't see any issues license wise.

That is my thinking.

> I don't think we need to change anything in the code, just a different cef binary.

Yes, all that will need to change is for us to use a custom build of CEF. I have experience with doing that, the upgrade PR was custom builds until it became Chromium 63.

dotarmin:
> @baltedewit I don't think we need to change anything in the code, just a different cef binary.

> @Julusian Yes, all that will need to change is for us to use a custom build of CEF. I have experience with doing that, the upgrade PR was custom builds until it became Chromium 63.

I think it would be great of having this BUT I'm a bit sceptical. Now when we use Nuget for some of the packages (including CEF) and aim to continue with that in the future with the rest of packages that are not nugetified today. I think going back to custom builds has a potential risk of putting us in a situation as we where with CEF and CasparCG Server 2.0.7 and 2.1.0, making it hard to maintain and update.

> @baltedewit so I don't see any issues license wise
The decision is not really on me or my employer, it's more like a community thing to discuss. I think also that person making the pull request (if done) has to take responsibility for not breaking any licensing terms, which is a quite common thing in open source projects.

baltedewit:
[...]

PR #695, opened 2018-01-28
WIP: [html] Upgrade to CEF3239 (Chromium 63)

Julusian:
Rebase #611 and #644 for 2.2.0

This upgrades the HTML producer to a more modern version of Chromium (CEF3239 / Chromium 63).

This builds on the work from @AndreasWilladsen and @sesse. Any additional tweaks or tests to run are appreciated. I have labelled this PR as a WIP as performance is not yet perfect and some frames appear to be being lost, however the same is true of the current version

Improvements:
- Newer chromium. Gives new features and makes template dev much easier at it matched current versions of chrome
- Font smoothing
- 3D CSS

Missing:
- WebGL (Pending [CEF #1006](https://bitbucket.org/chromiumembedded/cef/issues/1006))


Julusian:
This isn't currently ready for merging until #692 is resolved as I suspect this might be missing some linux so files, as they were previously being included.

Julusian:
@dotarmin @ronag I think I asked before but don't remember getting an answer, what are your thoughts on using a CEF build with proprietry_codecs enabled. This will give h264 and mp3 playback. Its disabled for royalty licensing reasons, but as we already support them via ffmpeg I don't see any issue there. #350

Punkley:
I would also like to see propriety_codecs enabled, would make things a bit more functional.

Kingsley

ronag:
@Julusian this doesn't build on windows :(

ronag:
I'm reverting this until it does build.

Julusian:
Thats odd. Im sure it did build the other day... Ill have a look this evening

ronag:
@Julusian how's the windows libs going? The default builds won't work since the static link to runtime.

Julusian:
I made a start on trying this out a couple of days ago and what I had built on windows, but I havent gotten around to checking it on linux yet, and there have been a lot of commits since so I ought to recheck it. 
Ill quickly check if it still builds and runs then raise another PR. I can deal with linux in another PR, unless you want it to be done in one?

ronag:
Linux in another PR is perfect. I'm developing on Windows.

PR #696, opened 2018-01-28
Use best available video quality when there are multiple video streams

pkeuter:
When playing a HLS or DASH stream which contains multiple video streams (like https://bitdash-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8), currently the first available quality is picked. 

It is better to check which video stream has the best quality (by comparing height pixels) and select that stream for playing.


ronag:
This has a problem with CasparCG 2.2.0, which will always use the second stream as an alpha stream. @jesperstarkar @5opr4ni.... thoughts?

5opr4ni:
Hm! Is it possible to flag a stream in when composing the second alphastream in the BoffSolution.
If that flag don't exist 696 could activate. Just a thought. We really need suport for the BoffSolution.

Julusian:
Would that be an issue? If the second stream is an alpha channel, then surely it will be the same (or lower) resolution, in which case the first will always be chosen. This change only makes a difference if the first channel does not have the largest vertical resolution.

pkeuter:
I agree with @Juliusan, this shouldn't be an issue given that the vertical resolution won't be higher. @ronag Do you need me to rebase on the current 2.2 branch? I see there have been 'some' commits ;-)

ronag:
> Would that be an issue? If the second stream is an alpha channel, then surely it will be the same (or lower) resolution, in which case the first will always be chosen.

You misunderstand. If there are more than 1 video stream then the second stream will always be used as the alpha channel. This is in order to get around the problem with no video codec has proper alpha support.

i.e. caspar will not consider multiple video streams to be different qualities of the same video.

ronag:
> @ronag Do you need me to rebase on the current 2.2 branch?

Yes, please rebase. Makes it easier to discuss the change :).

Juliusan:
Guys... you quote the wrong person :D

pkeuter:
Sorry @Juliusan, I mentioned @Julusian (so close). I see what you mean now @ronag. Not so easy to rebase this :-)

pkeuter:
> i.e. caspar will not consider multiple video streams to be different qualities of the same video
Could we work around this by checking if the stream is a HLS or DASH stream, or even more generic HTTP for that matter?

ronag:
No, we actually use this feature for HLS streams :)

pkeuter:
Isn't that just going to give problems when using a "regular" HLS stream? I mean 9 out of 10 times an HLS manifest contains streams in multiple resolutions. If I understand correctly, this could be a "low cost" solution to get an alpha channel in a regular stream. But then wouldn't this break (or at least give unexpected results) playing out regular HLS streams (because of the difference resolutions)?

ronag:
>  mean 9 out of 10 times an HLS manifest contains streams in multiple resolutions.

Not in broadcast situations :). But I see your point. Will have to think about it.

ronag:
I've disabled auto alphamerge. If you want this feature in it should be possible. PR please.

pkeuter:
Revised PR in  #812

PR #698, opened 2018-01-28
[screen_consumer] Enable display selection on windows

Julusian:
Rebase of #620 for 2.2.0
Fixes: https://github.com/CasparCG/Server/issues/619

Currently, this only enables screen selection on windows as libX11 does not present a method of determining the position of screens. xrandr appears to calculate by looking at the relative positions of the screens, and adding this logic to CasparCG feels messy. An alternate solution would be to implement https://github.com/CasparCG/Server/issues/487

Some docs have been added to indicate that the property only works on windows, including a warning to the console if used on linux.

This also fixes issues:
- the size of a fullscreen consumer would use the dimensions of the primary screen when on a secondary screen
- a windowed consumer would always be positioned on the primary screen (this required removing the close button from the window)

### DIFF
--- a/modules/screen/consumer/screen_consumer.cpp
+++ b/modules/screen/consumer/screen_consumer.cpp
@@ -38,8 +38,6 @@
 #include <common/os/general_protection_fault.h>
 #include <common/scope_exit.h>
 
-//#include <windows.h>
-
 #include <ffmpeg/producer/filter/filter.h>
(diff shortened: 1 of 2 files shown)


ronag:
Merged in https://github.com/CasparCG/Server/commit/09dc703a625f228ec1c86543057bae08c516437d

PR #700, opened 2018-01-28
[osc] Update paths when swapping layers

Julusian:
Using the SWAP command is currently dangerous if OSC data is used.
Running the following will cause layer data to be emitted claiming to be for layer 10 while the video plays in layer 11.
```
PLAY 1-10 test.mp4
SWAP 1-10 1-11
```

Additionally, the link does get broken until either the layer or channel is cleared (stop doesnt fix it). So by also running the following commands there will be two sets of layer data emitted claiming to be for layer 10
```
CLEAR 1-10
PLAY 1-10 test.mp4
```


ronag:
Swap should preferably be noexcept. Could you do a get_path as to avoid the extra allocation?

PR #701, opened 2018-01-29
[BUGFIX] Fixes segfault issue when outputting stacktrace

k0d:


ronag:
We’re killing the whole existing stacktrace implementation

k0d:
This is interfering with testing of the current code. If the stacktrace implementation is going to go within a couple of days...great...otherwise it'd be nice to still have this fix.

PR #702, opened 2018-01-29
[html] Fix building

krzyc:
Add missing icudtl.dat for linux
Use CEF_RESOURCE_DIR in CMakeLists.txt


Julusian:
Oops, I must have missed that CMakeLists change during a rebase, I did do that in a branch somewhere...

icudtl.dat needs to be committed to git-lfs. You might need to do a `git lfs install` in your local clone, then update the commit. (an amend might trigger it?) I don't know how I missed that file, I did a cherry-pick and git didn't complain about any merge issues..

ronag:
@Julusian is this mergable or waiting for LFS?

Julusian:
This is waiting for lfs. It really needs to be done before merging, otherwise it will take another history rewrite to properly correct it

krzyc:
I don't get this lfs stuff. Of course I have installed git lfs, checked that 'git lfs track' lists '*.dat' and then added this 'icudtl.dat' - still 'git lfs ls-files' don't show this file - it does not show windows 'icudtl.dat' file either. LFS shows only this files as LFS:
c43c61ae50 * dependencies64/cef/linux/Release/libcef.so
882686174f * dependencies64/cef/win32/Release/cef_sandbox.lib
2f23182ec6 * dependencies64/cef/win32/Release/d3dcompiler_43.dll
f183f5b1a2 * dependencies64/cef/win32/Release/libcef.dll

Julusian:
Huh... so in my commit only a couple of the files went into lfs. Looks like it found all the large files >10mb, and a couple of smaller ones, but missed a lod of the others. Looks to total ~100mb filesize.

Im trying to figure out why and how to make lfs see the files it has missed, but its not going well...

@ronag this looks that may want cleaning up, but that would require a force push of some kind.

krzyc:
After clean checkout file was added as LFS - I do not understand where the problem was, but I will make another pull request.

Julusian:
In short, the lfs inclusion check only gets run when doing a git add to a file. This means that cherry-pick will not normally trigger the check, and rebase or amend will only do so if the relevent files are modified.

So because I cherry-picked, only the conflicted the large file archive contents got pushed to lfs.
@ronag do you want to rewrite history to fix this (again) or live with the 100mb bloat?

Im going to have a quick look into if there are any services that will do a git-lfs validation on incoming PRs, to help flag this in future

PR #735, opened 2018-02-04
Modifies the build scripts for linux, doesn't complete a compile yet...

k0d:


Julusian:
@k0d I think I saw a commit removing the dependency on rxcpp

k0d:
It's true :) I'll remove that now...when I wrote that part of the code it was still needed.

PR #744, opened 2018-02-05
2.2.0 newfolders

k0d:
Still to test in Windows...


ronag:
Done. Probably needs a bit more restructuring. i.e. moving packages out and the cmake stuff?

PR #751, opened 2018-02-07
Build-script cleanup, static boost linking

k0d:


k0d:
Wait with this...I'm adding some more...

Julusian:
Is the recommended way of running casparcg on linux now within docker?

k0d:
@Julusian no...not really...It's just that's as far as I am with the building instructions/build-scripts. I'm going to write more soon that explains how you 'extract' CCG with all the needed deps.

ronag:
@Julusian that's my hope for the future... but it's still very much an open discussion. For now we're using docker to make it easier to build.

BUILDING.md
```diff
@@ -33,11 +33,8 @@ Linux
...
-1. `git clone --single-branch --branch 2.2.0 https://github.com/CasparCG/server`
-2. build-scripts/ubuntu-17.10/build-docker-image
-3. build-scripts/ubuntu-17.10/launch-interactive
-4. cmake /source
-5. make -j8 (or however many cores you want to use)
-6. /source/build-scripts/ubuntu-17.10/package
-
-If all goes to plan, a package has been created in /build/products which should be accessible outside of the Docker container
+1. `git clone --single-branch --branch 2.2.0 https://github.com/CasparCG/server` casparcg-server-2.2.0
+2. `cd casparcg-server-2.2.0`
```
ronag: Just leave it with the default clone name?

k0d: Sometimes I have multiple clones...and I don't like it just to say 'Server'....I'd prefer a new clone name if it's ok?

Julusian:
@k0d ok good. If you could also include a section on a manual build process (for incremential builds during dev) that would be great.

@ronag While I can see there are benefits to that, I would definitely prefer a discussion before deciding anything. I do agree it does make sense for building though.

k0d:
I don't believe that running inside of Docker will be the only officially supported way to run CasparCG Server. I'm building it outside of Docker during development now...and will give instructions for that asap.

Julusian:
@k0d Whilst I remember, the CEF build I had included previously for linux was a custom build. I was getting a crash during the opengl mixer initialization, and by building the exact same revision and following their build steps myself it solved the issue.
Let me know if you want a build matching the version now being used on windows. I have a vm setup for building it that I can fire back up.

k0d:
@Julusian I will look at CEF a bit later, there are 2 higher priority issues I need to fix first.

PR #756, opened 2018-02-08
Bug fix and improvements to how audio is handled by bluefish producer

jamesdwise:
- Fix bug in audio decoding, where only 2 channels of audio was ever being decoded.
- change default layout  from 2 channel ( stereo) to 8 channel.


ronag:
@jamesdwise would you mind re adding bluefish to the 2.2.0 branch? The only thing that needs fixing is that 2.2.0 no longer has “channel_layout” it always does 8ch pass through.

jamesdwise:
I am definitely planning to do so, 
was waiting until the code base stabilizes a bit, but will get to it soon.

PR #757, opened 2018-02-08
Reformat changelog to use markdown

Julusian:
Means it can be easily read on github. eg 
https://github.com/Julusian/CasparCG-Server/blob/patch-1/CHANGELOG.md vs https://github.com/CasparCG/server/blob/2.2.0/CHANGELOG.md


ronag:
Thanks!

Issue #766, opened 2018-02-08
Fresh 2.2.0 build fails during "copy dependencies"


wlcx:
I followed the instructions in building.md, and the build fails in "copy dependencies" due to not being able to find packages and boost headers. Apologies if I am doing something stupid!

Also, regarding the instruction
> Open ./src/CasparCG Server.sln
for me this file ends up in the project root, not `src`. 

OS: Windows 10 Ent
Git: 2.16.1
VS: Community 2017

Full build log:
https://gist.github.com/anonymous/13497e39b944601aa0a27c7419c0bd6d

ronag:
did nuget properly restore packages? What do you have in "C:/Users/bob/server/src/packages/FFmpeg.Nightly.20180130.1.0/build/native/bin/x64"?

wlcx:
I did try nuget but wasn't sure if I was doing it right.
`Right Click>Restore NuGet Packages` on the root solution in solutions explorer results in 
`All packages are already installed and there is nothing to restore.` I note that nuget doesn't seem to actually see any dependencies via `Manage Dependencies`...

Here is the contents of my packages folder
![image](https://user-images.githubusercontent.com/3065381/35988311-0012b42e-0cf6-11e8-9885-54966ea3cef1.png)


I am not familiar with VS and NuGet, is there an additional step I need to take first?

ronag:
That should have worked. I will look into it when I get the time.

wlcx:
Thanks, glad it's not just me! Is it worth having the nuget step in `building.md` too?

ronag:
visual studio should do it automatically when building the solution.

ronag:
@wlcx if you want alpha builds join us in the slack channel. We're testing and discussing there.

wlcx:
Would love to - though can't find any links to join?

On 8 Feb 2018 18:03, "Robert Nagy" <notifications@github.com> wrote:

@wlcx <https://github.com/wlcx> if you want alpha builds join us in the
slack channel. We're testing and discussing there.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/CasparCG/server/issues/766#issuecomment-364197132>, or mute
the thread

ronag:
@hummelstrand

dotarmin:
@wlcx Have you access by now? If not, let me know and I can help you.

wlcx:
@dotarmin I don't yet no! Email's on my profile if that helps :)

[...]

PR #770, opened 2018-02-08
Windows cmake to /build directory & cef include nuget

Julusian:
This currently gives me an error building the image module, I havent looked into why. Thought it best to share what I have in case someone wants to look into it as I wont have time until next week.

This also moved the cef includes into nuget. Might be cherry-pickable, but it also might rely on the previous commit to work properly


Julusian:
@ronag I am happy with this now, after your image producer fix it is building for me.
The only bit I am uncertain about is that if a new nuget package gets added, the matching files in /src will likely need manual updating, but that wont happen often so I think it is fine.

The build docs have been updated to reflect this.

Fixes #766

ronag:
@k0d please take a look

ronag:
@Julusian could you resolve last conflicts and then I'll merge. Thanks!

ronag:
Does version.h work with this?

Julusian:
Sure, ill rebase this when I have some time later today.

I'm not sure if version.h works, but I shall check

Julusian:
@ronag This is rebased on latest 2.2.0

ronag:
@Julusian where should the sln file end up?

Julusian:
It will end up in whatever folder cmake is run from. The build docs have been updated to say to run it from within a /build folder, but it should also work as before when cmake is run in /src

ronag:
@Julusian I seem to have some problems with release build

ronag:
```

15>libcef_dll_wrapper.lib(navigation_entry_ctocpp.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '2' doesn't match value '0' in main.obj
15>libcef_dll_wrapper.lib(navigation_entry_ctocpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MDd_DynamicDebug' doesn't match value 'MD_DynamicRelease' in main.obj
```

ronag:
@Julusian probably something to do with the cef changes.

Julusian:
strange, I thought I checked that... Ill have a look now

PR #805, opened 2018-02-14
return support for SBS key

polardegausser:
Return the support for side-by-side key (for budget builds with dual HDMI output)


ronag:
@polardegausser: Thanks! Unfortunately this will have to wait until next release. Sorry.

ronag:
@polardegausser could we do this using 2 render calls instead of the memshfl? We are planning to skip the device->host->device round trip in future releases which would break this implementation.

Julusian:
@polardegausser this was closed unintentionally as part of a branch rename. Please open a new pr against the master branch of you want it to be considered for merging :)

Andols0:
Is this abandoned or on hold?

PR #812, opened 2018-02-14
check for best videoquality before selecting stream

pkeuter:
This is a revised PR based on the latest changes in the 2.2.0 branch. This supersedes https://github.com/CasparCG/server/pull/696

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -220,8 +220,8 @@ struct Filter
 					input->streams[n]->codecpar->height > best_height) {						
 						best_height = input->streams[n]->codecpar->height;
 						force_index = n;
-					}
-                    count += 1;
+				}
+                count += 1;
             }
             //if (count > 1) {
             //    filter_spec = "alphamerge," + filter_spec;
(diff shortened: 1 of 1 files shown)


pkeuter:
To test the difference, try to play this stream: https://bitdash-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8

Before: 
image

After: 
image

ronag:
It should only do this if number of filter video inputs is equal to 1.

pkeuter:
@ronag I've changed this now, although I don't really understand why. Can you explain?

ronag:
Because if I have a alphamerge filter in my graph it should work as expected :)

ronag:
It needs to go inside here https://github.com/pkeuter/Server/blob/8b988c9165c9d58c303aea289c463fcbb537c996/src/modules/ffmpeg/producer/av_producer.cpp#L212

pkeuter:
@ronag, sure although that makes it necessary to  add a force_index flag. I've done that now, I assume this is what you mean? Else please elaborate.

ronag:
Good enough.

PR #859, opened 2018-02-21
Add support for Hw/ application watchdog to Bluefish consumer

jamesdwise:
Add support for an application level watchdog, that talks to the bluefish device driver and can enable bypass relays if the application ( casparCG) crashes,

Watchdog only works on the first sdi- stream.  
watchdog is enabled by default, but can be disabled by setting a value of in the config file for the given channel.


jesperstarkar:
Thank you James!

ronag:
LGTM

PR #863, opened 2018-02-21
Make sure DIAG window shows play channel and layer of HTML producer

pkeuter:
This is currently not working because the graph sink is being created within a CEF context. I've fixed this by creating the sink while still in a Caspar context and then passing it.

This fixes the issue that the DIAG window wouldn't show the channel and layer where it's playing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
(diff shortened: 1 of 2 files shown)


ronag:
I don't like this graph context thing but that's for another day. Thanks!

PR #880, opened 2018-02-26
Config: allow channel with no consumers tag

krzyc:
Now Caspar ends with exception when there is no consumers tag on channel.


ronag:
Someone else tried this?

pkeuter:
Just did. LGTM 👍

PR #886, opened 2018-02-27
Build Scripts for Ubuntu 16.04 / 17.10

maceyldn:
All,

I've created 2 build scripts, for 16.04 and 17.10, this is to install all the required dependencies and compile boost/ffmpeg.

Be aware that on 16.04 - GCC 7.2 will be installed from backports required for building CCG.
This will become the default Compiler

Andy

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -50,6 +50,19 @@ Extracting CasparCG Server from Docker
 
 You will then find a folder called `casparcg_server` which should contain everything you need to run CasparCG Server.
 
+Development under Ubuntu
+-----------
+
+2 Scripts have been created to install dependencies and build boost/ffmpeg/cef.
+
+Ubuntu 16.04 (Xenial):
+From the base of the git repo, please run tools/linux/casparcg_build_dev_xenial.sh
+
+Ubuntu 17.10 (Artful):
+From the base of the git repo, please run tools/linux/casparcg_build_dev_artful.sh
+
+If all goes to plan, a folder called 'staging' has been created with everything you need to run CasparCG server.
+
 Development
 -----------
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,12 @@ INSTALLATION ON WINDOWS
 
 INSTALLATION ON LINUX
 ---------------------
-   
+
+DEVELOPMENT / BUILDING FROM SOURCE 
+---------------------
+Please refer to BUILDING.md
+
+
 RESOLVING COMMON ISSUES ON LINUX
 --------------------------------
 
--- a/tools/linux/casparcg_build_dev_artful.sh
+++ b/tools/linux/casparcg_build_dev_artful.sh
@@ -65,54 +65,65 @@ apt-get install -yq --no-install-recommends \
 		libglfw3-dev \
 		libsfml-dev
 
+
 mkdir build_deps
 cd build_deps
-#wget https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.gz
-#tar -zvxf boost_1_66_0.tar.gz
-#cd boost_1_66_0
-#./bootstrap.sh --prefix=/opt/boost
-#./b2 --with=all -j8 install
+
+echo "Build Boost"
+
+wget https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.gz
+tar -zvxf boost_1_66_0.tar.gz
+cd boost_1_66_0
+./bootstrap.sh --prefix=/opt/boost
+./b2 --with=all -j8 install
 
 
-#cd ../
-#wget https://github.com/FFmpeg/FFmpeg/archive/acdea9e7c56b74b05c56b4733acc855b959ba073.tar.gz
-#tar -zxvf  acdea9e7c56b74b05c56b4733acc855b959ba073.tar.gz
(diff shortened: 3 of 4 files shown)


Julusian:
Thanks for this.
I'm not sure how I feel about it being merged though. The current build script is done via docker, and this PR essentially replicates that process outside of docker, meaning there are now 3 almost identical linux scripts to update. 
At the very least it would be good if ffmpeg and boost can be extracted from the docker images, as they will rarely need to be changed and that cuts down on a lot of the complexity. That is assuming that those builds will work on Ubuntu16.04.
I believe that a docker version for 16.04 will be needed if any official builds are to be produced.

maceyldn:
Understood, These scripts do give you a simple build enviroment under Ubuntu though..

Upto you guys.
Andy

Julusian:
> These scripts do give you a simple build enviroment

Yeah, that does appeal to me so I am a bit undecided

baltedewit:
I am not against having a non-docker script as well.

I think that so long as we don't provide linux binaries, it'd be great to be able to run installation from docker. And even then, it'd be good to be able to compile for multiple targets.

I'd love to put docker images together for a debian build, but I'm a bit short on time unfortunately.

Julusian:
> I'd love to put docker images together for a debian build,

This is where this starts turning into a problem. In 2.1 it was easy, because the builds contained all the libs needed to run it meaning that the one build would run on anything. But now that it doesn't, it means there needs to be a seperate docker build script for each supported distro version, and potentially scripts to setup a dev environment. And then there will be people wanting to run on unsupported distros (I too would prefer to use debian), so the build scripts could very quickly become a hard to maintain mess unless there is a definite plan for them.

walterav1984:
It does build on ubuntu 16.04 and 18.04(using artful script) however I keep getting these errors while trying to run.sh about libsoftokn3.so related to nssdb?

```
MESA_GL_VERSION_OVERRIDE=4.5 ./run.sh
Type "q" to close application.
Logging [info] or higher severity to log/

[2018-03-02 11:38:32.938] [info]    ############################################################################
[2018-03-02 11:38:32.938] [info]    CasparCG Server is distributed by the Swedish Broadcasting Corporation (SVT)
[2018-03-02 11:38:32.938] [info]    under the GNU General Public License GPLv3 or higher.
[2018-03-02 11:38:32.938] [info]    Please see LICENSE.TXT for details.
[2018-03-02 11:38:32.938] [info]    http://www.casparcg.com/
[2018-03-02 11:38:32.938] [info]    ############################################################################
[2018-03-02 11:38:32.938] [info]    Starting CasparCG Video and Graphics Playout Server 2.2.0.3313 c37b017 Beta 1
[2018-03-02 11:38:32.940] [info]    Initialized ffmpeg module.
[2018-03-02 11:38:32.940] [info]    Initialized oal module.
[2018-03-02 11:38:32.940] [info]    Initialized decklink module.
[2018-03-02 11:38:32.940] [info]    Initialized screen module.
[0302/113832.976468:ERROR:nss_util.cc(650)] Error initializing NSS with a persistent database (sql:/home/ua/.pki/nssdb): libsoftokn3.so: cannot open shared object file: No such file or directory
[0302/113832.976596:ERROR:nss_util.cc(154)] Error initializing NSS without a persistent database: NSS error code: -5925
[0302/113832.976609:FATAL:nss_util.cc(156)] nss_error=-5925, os_error=0
Aborted (core dumped) 
```

Removing `~/.pki`, `~/.config/cef_user_data` or  `~/.config/chromium` didn't help any suggestions?

baltedewit:
Do you have the nss package installed?

walterav1984:
As in "libnss3" its installed, during the end stage when symlinking casparcg staging I see that there are some nss libs being linked.

ronag:
I don't like this and don't see the point. What's wrong with building in docker? Note, I haven't followed this discussion and I am not and will not maintain Linux stuff so I will leave it up to those that do (@Julusian ) to decide whether to merge this or not.

baltedewit:
docker runs a different distro inside a container, so the current builds will only on Ubuntu 17.10, which is EoL in July.
I don't know what the decision for 17.10 was. It makes more sense to have had it working on 16.04 or to focus on 18.04 immediately.

I personally don't care if it builds using docker or not, since it's easy enough to extract the required steps from the docker files. Regardless, Julusian's comment makes sense: pick something to support, make it work well and if someone wants to maintain a version for their own distro they are free to do so.

walterav1984:
@k0d 

Ubuntu 18.04 entered its feature freeze today, it comes with boost 1.65.1 and ffmpeg 3.4.2 maybe that lowers the requirements for external libs for casparcg instead of the 17.10 docker build, or focus on 16.04?

ronag:
I would much rather see this PR turned into making the docker build ubuntu 18.04

Julusian:
> Ubuntu 18.04 entered its feature freeze today

Yes, I think that now that it has, we should focus on 18.04 instead of 16.04. I don't know what the expected timeframe is for 2.3.0/3.0.0 but hopefully they will line up sufficiently well.

> it comes with boost 1.65.1 and ffmpeg 3.4.2 maybe that lowers the requirements for external libs

I don't think that will help as I believe we are using boost features that were added in 1.66, and we want the ffmpeg version to exactly match that in use on windows, so a non-system version is still best there.

ronag:
I don't think we're going to merge this as is. Please open a new PR for docker + 18 LTS

PR #903, opened 2018-03-04
Store HTML filename in state for OSC

pkeuter:
This will make OSC output at least some information about what is currently happening.


ronag:
I think the standard convention here is to put the full path (i.e. url) under `path` and just the filename under `name`.

PR #904, opened 2018-03-04
Implement HTML filename in OSC when available

pkeuter:
@ronag You were quick to merge. I was also adding the filename, when it's available 😄


ronag:
This doens't handle real http urls. 

I've been considering whether we in the future should deprecate the template folder once we deprecate flash. So http urls need to work properly.

ronag:
for http urls it should pick the "html" filename and if it is "index.html" it should pick the url path (i.e. exclidong proto, host etc...)

pkeuter:
As just discussed with @ronag. We're not adding the filename since it's almost never used anyway and the template folder will be removed in a future version.

PR #905, opened 2018-03-04
Don't let ffmpeg handle .html files

pkeuter:
This PR does a few things:

* Checks if the filename contains '.html' in the ffmpeg producer. And if so, returns an empty frame producer.
* Checks if the ffmpeg producer could be created and if not, returns an empty frame producer (currently it will swallow the request and if it fails, it stops. This gives other producers no change to handle the files)
* Removes the [HTML] prefix from the HTML producer. The HTML producer tries to handle input after the ffmpeg producer, so all HLS/DASH and other HTTP-like streams will be handled by ffmpeg in the first place. Because of the previous point, all other items that ffmpeg can't handle will then be passed to subsequent producers (including the HTML producer) and thus, the [HTML] prefix is unneccesary.

What I've tested:
* Normal file (PLAY 1-1 AMB)
* HLS stream (PLAY 1-1 https://bitdash-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8)
* HTML template (PLAY 1-1 graphics, graphics.html in template directory)
* PLAY explicit HTML file (PLAY 1-1 https://google.com/index.html)
* PLAY regular URL (PLAY 1-1 http://google.com)

About the last point: When ffmpeg does try to play the html page (e.g. http://google.com), it will log an error (just as now), but then the HTML producer will pick it up and it will play succesfully.

### DIFF
--- a/src/modules/ffmpeg/producer/ffmpeg_producer.cpp
+++ b/src/modules/ffmpeg/producer/ffmpeg_producer.cpp
@@ -320,10 +320,15 @@ spl::shared_ptr<core::frame_producer> create_producer(const core::frame_producer
     auto vfilter = boost::to_lower_copy(get_param(L"VF", params, filter_str));
     auto afilter = boost::to_lower_copy(get_param(L"AF", params, get_param(L"FILTER", params, L"")));
 
-    auto producer = spl::make_shared<ffmpeg_producer>(
-        dependencies.frame_factory, dependencies.format_desc, name, path, vfilter, afilter, start, duration, loop);
-
-    return core::create_destroy_proxy(std::move(producer));
+    try {
+        auto producer = spl::make_shared<ffmpeg_producer>(
+            dependencies.frame_factory, dependencies.format_desc, name, path, vfilter, afilter, start, duration, loop);
+        return core::create_destroy_proxy(std::move(producer));
+    }
+    catch (...) {
+        CASPAR_LOG_CURRENT_EXCEPTION();
+    }
+    return core::frame_producer::empty();
 }
 
 }} // namespace caspar::ffmpeg
--- a/src/modules/ffmpeg/producer/ffmpeg_producer.cpp
+++ b/src/modules/ffmpeg/producer/ffmpeg_producer.cpp
@@ -36,6 +36,7 @@
 
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/logic/tribool.hpp>
 
 #include <cstdint>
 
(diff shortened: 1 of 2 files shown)


dimitry-ishenko:
@pkeuter with all due respect `if (boost::contains(path, L".html")) {` is a rather naive way of determining if a file is an HTML page. What if I try to play a file called "Screenshot of page.html webpage.png"?

Also, what about .htm extension?

pkeuter:
@dimitri-ishenko, I agree that it's naive. But there are only so much use cases. This at least filters out the most common. We could do an ends_with comparison on .html or .htm? This would fix both your cases, but I don't think we can handle all use cases this way. We might keep the [HTML] alive though, just to hint the other producers that it's not for them... @ronag probably has an idea about this anyway.

ronag:
Use a regex, see www.regex101.com

ronag:
something like: `^https?:|\.html?$`

ronag:
Though, I'm not sure we need this. Doesn't ffmpeg probe return a failure?

ronag:
I think it's enough to just add `.html` to the invalid_exts array

ronag:
I would like to move away from supporting extension less file paths and urls in the future.

pkeuter:
@ronag I had to separate out the extension check out of is_valid_file because that function doesn't work for URLs. But this works. I also added the `[HTML]` prefix check.

pkeuter:
@Julusian, you are correct about the rule (which has been improved to check just the end and also htm-files by the way). But it would still have worked because ffmpeg would have aborted the creation of a producer, and because of that the HTML producer would have still created the actual producer. 

I do understand that you might want to directly go to the HTML producer though, so I've re-added the [HTML] specifying. The main advantage now is that every webpage you try to play without explicitly specifying [HTML] will not end up in an error (well it will log the error), but will play.

pkeuter:
Hah yeah, easiest way to use the undetermined state. Could have separated them into two functions too. Do you prefer that?

ronag:
No, it's fine. Just wanted to comment on a good solution 👍

Julusian:
@pkeuter Thanks for adding back an optional prefix, I think that is good to have as a way to force something to go to the correct producer when the auto-resolve can't figure it out. I do think that prefixes is something we should encourage, but that is a separate discussion for AMCP 3.0

I do have one more concern with this change, which is that ffmpeg is now aware of the html module in that it is avoiding files it thinks should be for that. This feels like something we should avoid doing, as we could easily end up adding more rules which will end up getting very complex and intricate. 
Saying that, I no longer object to this being merged.

pkeuter:
@Julusian. Thanks for the comments! Yeah, removing the prefix in the first place wasn't such a great idea. But in the end it gave me the possibility to remove some duplicate code :-). About your last point; that is not entirely the case. It just adds two file extensions which it should not even try to handle. I assume that a .html or .htm files (just like for example .gif or .swf files) are files you never want to feed to the ffmpeg producer. So the concerns you have about tight coupling shouldn't really be an issue here. Have you tested it by the way?

Julusian:
@pkeuter I wouldn't call this coupling very tight, and it is more of a general concern than with your changes. We already have a list of file extensions which are to be handled by other modules, which feels like it shouldn't need to exist, but again that is a seperate discussion. You are simply following what has already been done. 

@ronag I have tested this and am happy with it. My only concern is the exception logged by ffmpeg when it fails to play before passing the command onto the next producer. Perhaps if we are playing a url create_producer should log something simple like "Failed to play remote file" instead of using `CASPAR_LOG_CURRENT_EXCEPTION() ` in the catch block? I don't want to end up hiding too many exceptions though and make it harder to trace real issues.

ronag:
> My only concern is the exception logged by ffmpeg when it fails to play before passing the command onto the next producer.

It shouldn't send any exception here since .html is ignored, no? I think you are referring to another issue (which doesn't exists in GitHub yet).

ronag:
Has anyone also verified `CG ADD 1 http://localhost:3000/index.html 1`?

Julusian:
> It shouldn't send any exception here since .html is ignored, no?

It does if you try `PLAY 1-10 http://google.com` to essentially, any url without an extension. Although, that might be what it does already. Ill check

pkeuter:
@Julusian @ronag, yes currently it will log an error and stop. From this PR it will log the error and continue.

ronag:
Verify `CG 1 ADD 1 http://localhost:3000/index.html 1` and then I'm happy to merge.

Julusian:
> Verify CG 1 ADD 1 http://localhost:3000/index.html 1 and then I'm happy to merge.

Yep, that works as expected.

> yes currently it will log an error and stop. From this PR it will log the error and continue.

That behaviour is something new, so it would be nice to fix to reduce confusion, but it does work fine as it currently stands

pkeuter:
@Julusian That is one of the main reasons for this PR. It is good that if ffmpeg can't play a file, other producers get the change to try to play it.

Julusian:
@pkeuter I like the behaviour, but not that the exception is still being logged as it looks as though something went wrong, even though this is now expected behaviour

pkeuter:
@Julusian I know, and agree, but swallowing the exception isn't such a good idea anyway. What we could do is debug log the error and info log that ffmpeg producer couldn't handle it? I'm not sure what will happen when none of the producers are able to pick up the playrequest though.

ronag:
@Julusian @pkeuter this is something we should resolve, please create a new issue and move the discussion there.

Issue #828, opened 2018-02-15
Windows Build - Shell.rc error from zip


PeterAkakpo:
Building on windows with the current 2.2.0 source gives an error because "shell.rc" includes "version.h" which is always not in the right place. "version.h" is suppose to be in the shell folder but its not there. copying the "version.h" into the shell folder gives a successful build

ronag:
Did you follow the instruction in BUILDING.md exactly? Because depending on where and how you run cmake you can get this issue. @Julusian

PeterAkakpo:
I have realize that anytime i clone the repo (which is larger), build is successful. This issue occurs anytime i download the ZIP file (which is about 6MB). Is there any difference (in terms of code)  between cloning the repo and downloading the ZIP file?

ronag:
@PeterAkakpo I don't know. How exactly do you download?

PeterAkakpo:
![cg](https://user-images.githubusercontent.com/7926382/36265655-98441a14-1267-11e8-9c60-ab52a20f5229.png)

Julusian:
When you download the zip file, you won't be getting any of the files which are stored in LFS. So it looks as though cmake failed to run fully because those files were still the LFS pointers instead of full files.

I expect that downloading as a zip should be fine after #734, or when building for linux

PeterAkakpo:
Just "version.h"

ronag:
@PeterAkakpo "version.h" should be auto generated from "version.tmpl".

Julusian:
> We could remove LFS at this point I believe.

Everything still in /src/packages is still in LFS. <25mb according to the tree I am currently in.

ronag:
@Julusian I'm fine with converting to regular git. What do you think?

PeterAkakpo:
yes version.h is generated into a folder called "generated" instead of the "shell" folder.

ronag:
that doesn't sound LFS related

ronag:
@PeterAkakpo could you try with the zip again?

PeterAkakpo:
Ok, let me do that now

ronag:
ah! the problem is that version.h cannot be generated since the version is read from git.

ronag:
@PeterAkakpo I don't know how to fix this properly. I think we're leaving it as is for now. Until someone finds a solution...

PeterAkakpo:
ok this is my result. i just downloaded the zip and noticed that the size has increased by about 3M. 
[...]

PR #907, opened 2018-03-06
fix: Building from zip on Windows

Julusian:
Solves the issue with invalid CASPAR_REV and CASPAR_HASH fields in version.h when building from zip (no .git folder present)

With this I can now build from zip following the instructions in BUILDING.md, skipping just the git clone step.

#828


ronag:
perfect!

Issue #920, opened 2018-03-11
2.2.0 beta 1 performance


jesperstarkar:
### Expected Behaviour
- Better performance in Decklink producer playback.
- Performance that at least matches 2.0.7 and 2.1.0 beta 2.
- Near-linear growth of CPU, GPU and RAM consumption as load increases with new sources and layers.

### Current Behaviour
- 2-4x less performance in Decklink producer playback compared to 2.0.7 and 2.1.0 beta 2.
- Exponential growth in resource consumption.
- Unexpected crashes at some strange threshold when CPU/GPU maxes out.

### Steps to Reproduce 
- 4 channels 1080i5000 with no consumers
- 8 Decklink inputs available
- Fire one command at the time, measure CPU and RAM (Windows Task Manager) and GPU load (GPU-z):
```
play 1-0 decklink device 1 format 1080i5000
play 2-0 decklink device 2 format 1080i5000
play 3-0 decklink device 3 format 1080i5000
play 4-0 decklink device 4 format 1080i5000
play 1-1 decklink device 5 format 1080i5000
play 2-1 decklink device 6 format 1080i5000
play 3-1 decklink device 7 format 1080i5000
play 4-1 decklink device 8 format 1080i5000
```

### Environment
* CasparCG Server version: 2.2.0 beta 1 vs 2.1.0 beta 2 vs 2.0.7 stable
* Operating system: Windows 7

### Attachments
```
2.2.0 beta 1

Idle at 0% cpu and 0% gpu
1 input 7% cpu and 34% gpu 3,25gb ram
2 inputs 17% cpu and 36% gpu 3,53gb ram
3 inputs 33% cpu and 56% gpu 3,72gb ram
4 inputs 81% cpu and 93% gpu 4,06gb ram. Crashes after a few seconds.
```

```
2.1.0 beta 2

Idles at 3% cpu and 0% gpu 3,60gb ram

1 input 3% cpu and 30% gpu 3,72gb ram
2 inputs 4% cpu and 16% gpu 3,84gb ram
3 inputs 5% cpu and 24% gpu 3,98gb ram
4 inputs 5% cpu and 32% gpu 4,15gb ram
5 inputs 6% cpu and 34% gpu 4,42gb ram
6 inputs 7% cpu and 35% gpu 4,70gb ram
7 inputs 9% cpu and 40% gpu 4,98gb ram
8 inputs 9% cpu and 40% gpu 5,24gb ram
```

ronag:
[...]

PR #922, opened 2018-03-11
fix(decklink): guard against null video or audio frame from decklink

Julusian:
Fixes the crash in #920


ronag:
This might cause some other v/a sync issues but it's better than crashing.

ronag:
@Julusian it's better to have a:
```
if (!video || !audio) {
return;
}
```

in the beginning of the func

PR #928, opened 2018-03-13
fix: split osc packets

ronag:
I don't have time to do this properly. Hoping someone can take over finish up and test.

It should be good enough. Although room for improvement. But it needs test.

### DIFF
--- a/src/protocol/osc/client.cpp
+++ b/src/protocol/osc/client.cpp
@@ -73,6 +73,8 @@ struct client::impl : public spl::enable_shared_from_this<client::impl>
     std::map<udp::endpoint, int>             reference_counts_by_endpoint_;
     std::vector<char>                        buffer_;
 
+    int64_t time_ = 0;
+
     std::mutex                mutex_;
     std::condition_variable   cond_;
     boost::optional<core::monitor::data_map_t> bundle_opt_;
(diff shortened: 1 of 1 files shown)


src/protocol/osc/client.cpp
```diff
@@ -112,10 +114,14 @@ struct client::impl : public spl::enable_shared_from_this<client::impl>
                         continue;
                     }
 
-                    ::osc::OutboundPacketStream o(reinterpret_cast<char*>(buffer_.data()),
-                                                  static_cast<unsigned long>(buffer_.size()));
+                    // TODO: Should use server clock.
+                    time_++;
```
ronag: It's good enough

src/protocol/osc/client.cpp
```diff
@@ -126,13 +132,33 @@ struct client::impl : public spl::enable_shared_from_this<client::impl>
                         }
 
                         o << ::osc::EndMessage;
+
+                        if (o.Size() >= 1024) {
```
ronag: Limit is 1500 bytes... this an approximation.

ronag:
Helges implementation actually did this mostly correctly. It was just very complicated.

dotarmin:
Are you still working on this @ronag?

dedicatedbroadcastsolutions:
This appears to be superseded by commit [44ce8303e ](https://github.com/CasparCG/server/commit/44ce8303e2466b3b6a3503cff2d75a667407beda)("fix(osc): split OSC packets into time tagged bundles") by ronag, which implements the same fix more robustly (proper bundle time tag instead of a placeholder counter, and a cleaner size-based split). This should be ready to close this as already resolved on master.

PR #929, opened 2018-03-13
Build freeimage against vs2015

### LINKED ISSUE #926: Requires VC Redistributable 2013 package from specific commit
### Expected Behaviour
Able to run CasparCG Server 2.2.0 after having downloaded a build [casparcg.builds.com][1] with only requirement of having `VC Redistributable 2015` package installed.

### Current Behaviour
It requires `VC Redistributable 2013` to be able to run builds from commit 3024a11 and onwards.

### Steps to Reproduce 
1. Download a build from commit 3024a11 and onwards available at [casparcg.builds.com][1]
2. Extract the downloaded file
3. Start casparcg.exe directly or use the provided .bat file.
4. Error about `VCOMP120.dll` will be displayed.

[1]: http://builds.casparcg.com

Julusian:
Fixes #926 

I had to make my own nuget package for this.
Build scripts with instructions: https://github.com/Julusian/nuget-freeimage

@dotarmin do you want to transfer that repo to the caspar org?
If you create a casparcg nuget user, I can make you an owner of the nuget package too.

Alternatively, if it is too much work to maintain a nuget package, this could be put back into git as it is small enough and new versions arent released very often.

### DIFF
--- a/src/CMakeModules/Bootstrap_Windows.cmake
+++ b/src/CMakeModules/Bootstrap_Windows.cmake
@@ -135,10 +135,11 @@ casparcg_add_runtime_dependency("${NUGET_PACKAGES_FOLDER}/sfml-system.redist.2.4
 casparcg_add_runtime_dependency("${NUGET_PACKAGES_FOLDER}/sfml-system.redist.2.4.2.0/build/native/bin/x64/v140/Release/dynamic/sfml-system-2.dll")
 
 # FREEIMAGE
-set(FREEIMAGE_INCLUDE_PATH "${NUGET_PACKAGES_FOLDER}/native.freeimage.3.17.0/build/native/include")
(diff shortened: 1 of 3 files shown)


dotarmin:
> @Julusian do you want to transfer that repo to the caspar org?
> If you create a casparcg nuget user, I can make you an owner of the nuget package too.

I'm not sure, I'm a nuget fan and like the simplicity of it and benefits. I doesn't seem to be that much work to maintain it (or at least freeimage), especially if we help each other? If we go. this way we could then include other libs as well.

@ronag What do you think? If we go this path we have to make sure somebody (hopefully together?) maintain this and update versions when needed. 

It would be easier to update to new versions when of dependencies when needed.

ronag:
> @ronag What do you think? If we go this path we have to make sure somebody (hopefully together?) maintain this and update versions when needed.

If maintenance becomes a problem it's simple to revert back to inline deps.

dotarmin:
@Julusian We can talk about how to do this in best way at Skype! 😄

Issue #941, opened 2018-03-20
UNABLE to get BACKGROUND media via OSC or AMCP INFO


TondaKrist:
Expected Behaviour
Return snapshot state of specified layer, either via AMCP INFO or OSC

Current Behaviour
Unable to get info about background media in specified layer
and other details which were able to get via AMCP in <= 2.1

Julusian:
Related #923 #924

Punkley:
Background media information is a must to be able to get.

TomKaltz:
Regarding my previous comment https://github.com/CasparCG/server/issues/857#issuecomment-372438022
> I was tinkering and it seems that OSC will not get updated with the info from the background since it's not currently receiving frames since it's in the background. My work-around was to send the background producers print() information on every foreground frame receive but obviously this is a HUGE hack and delivers the background producer data in a weird string format.

@ronag do you think there is an easy solution to this problem or is my hack something that can be released as stop-gap measure until a proper fix is implemented?

ronag:
It's easy to just add another osc path.

Julusian:
#1063 #1080

PR #950, opened 2018-03-25
PR wip send background producer info via osc

TomKaltz:
This is all I've been able to figure out to get info about background producer.  The producer updates it's state in the ```receive()``` method which hasn't yet been called if it's in the background.  Currently this is a hack to call ```print()``` on the background producer and send it via osc.  @ronag can you give us some direction on this?

Related #941


Julusian:
Closing in favour of #1063

PR #966, opened 2018-04-15
Switch OpenGL code to use a Core 4.5 profile

### LINKED ISSUE #937: Target OpenGL Core profile
Depends on: #899 

Some code is using features only available on the compatibility profile, but Intel chips on Linux (not sure about windows) only support some of the required new feature extensions in the core profile.

From what I can tell, the following blocks of code all need rewriting. I checked all the opengl code I could find, so hopefully this is all of it.

https://github.com/CasparCG/server/blob/e7312587978273d16b3a80be43ee551e7ccd68b1/src/accelerator/ogl/image/image_kernel.cpp#L369-L395

https://github.com/CasparCG/server/blob/e7312587978273d16b3a80be43ee551e7ccd68b1/src/modules/screen/consumer/screen_consumer.cpp#L219
https://github.com/CasparCG/server/blob/e7312587978273d16b3a80be43ee551e7ccd68b1/src/modules/screen/consumer/screen_consumer.cpp#L329-L339

The diag graph will also need rewriting or removing as that uses SFML/Graphics, which is dependent on compatability profile features.

With SFML, it looks to be possible to create a core profile via this context constructor: https://www.sfml-dev.org/documentation/2.4.2/classsf_1_1Context.php#a2a9e3529e48919120e6b6fc10bad296c
There is also a similar constructor for window, which will be needed for the screen consumer.

Julusian:
Fixes #937 

The OpenGL code for the mixer and screen consumer has been updated to use a OpenGL 4.5 Core profile context. This has been tested on Debian 10 on an Intel HD5500. It should help on both windows and linux when the GPU claims to support OpenGL 4.5 but CasparCG doesn't detect that functionality.

Some of the code could do with review by someone who knows OpenGL. In particular I'm not sure what some of the bits I removed from the shader did, and it feels wrong to be creating buffers so often in the kernel during drawing, but I'm not sure where best to do so.

The diag window is still using a compatibility profile, but it is still runs even on linux.

I think this is best left to 2.3, as there is some risk of breaking things or decreased performance. Unless it gets sufficient testing before 2.2 is released

Windows: https://drive.google.com/open?id=1IXqVzTflx9-D1fsrCJ2ufl1XBMIqMUza
Ubuntu18: https://drive.google.com/open?id=1ialg921IzdVlWvpA0JZn6Tsh2SrS-JPd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@
 build
 CMakeFiles
 casparcg_server
+
+src/cmake-build-*
+src/.idea
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -310,10 +310,10 @@ struct image_kernel::impl
             double h = static_cast<double>(params.background->height());
 
             GL(glEnable(GL_SCISSOR_TEST));
-            glScissor(static_cast<int>(m_p[0] * w),
+            GL(glScissor(static_cast<int>(m_p[0] * w),
                       static_cast<int>(m_p[1] * h),
                       std::max(0, static_cast<int>(m_s[0] * w)),
-                      std::max(0, static_cast<int>(m_s[1] * h)));
+                      std::max(0, static_cast<int>(m_s[1] * h))));
         }
 
         // Set render target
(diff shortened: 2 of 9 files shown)


ronag:
merged

PR #993, opened 2018-05-09
feat(decklink) Add option to freeze on signal lost instead of going black

Julusian:


ronag:
@5opr4ni can we avoid having both behaviors? Is freeze the way to go always?

5opr4ni:
Hmmm, this is a tricky one. But I can't see any disturbances if we freeze the in if there is a signal-loss. As long as a reinit produces black if there is no signal. ;)

ronag:
Julusian can u make this the default behavior and remove the option?

hummelstrand:
May I humbly suggest that having this as an option achieves two things:
1. The current behaviour remains the default, avoiding a "breaking change".
2. Gives the user the choice to configure the software to suit her/his needs.

ronag:
I wouldn’t consider it breaking as it’s basically a bug.

I want to avoid unnecessary complexity. Is there any valid use case for the existing behavior? If yes then I’m fine with having options.

premultiply:
Yes there is the usecase that you may want to see if a signal is missing.
Maybe for a multiscreen (multiview) display on a secondary channel...

Julusian:
I agree that keeping the original behaviour is useful to be able to see signal loss in some scenarios. 
While it would be nice for this to be the default behaviour, it could be considered a breaking change, so I would be hesitant to change that (which is why I did not change it in this PR)

PR #994, opened 2018-05-10
Add producer for QtQuick

ServiusHack:
I've used QtQuick for several productions now. Its language (QML) is easier to write and more expressive than HTML. A lot of features, like animations, are easily accessible. After doing one event with CasparCG and HTML I missed the convenience of QtQuick so much that I had to try adding it to CasparCG. The result is this pull request.

The code is based on the HTML producer from 2.1.0 and [Qt's QQuickRenderControl example](http://doc.qt.io/qt-5/qtquick-rendercontrol-example.html). Unfortunately I'm not very knowledgeable in graphics programming which is why the code might not be that good. Very much looking forward to any comments on it.

~Please note that the code requires Qt 5.10 which might need to be installed manually. For cmake to pick it up set `CMAKE_PREFIX_PATH=/home/user/qt/5.10.0/gcc_64/lib/cmake/`.~ Qt is installed during the Docker build and instructions for Windows were added.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -19,17 +19,19 @@ Development using Visual Studio
 
 2. Install CMake (http://www.cmake.org/download/).
 
-3. `git clone --single-branch --branch master https://github.com/CasparCG/server casparcg-server-master`
+3. Install Qt (https://www.qt.io/download-qt-installer).
 
-4. `cd casparcg-server-master`
+4. `git clone --single-branch --branch master https://github.com/CasparCG/server casparcg-server-master`
 
-5. `mkdir build`
+5. `cd casparcg-server-master`
 
-6. `cd build`
+6. `mkdir build`
 
-7. `cmake -G "Visual Studio 15 2017" -A x64 ../src`
+7. `cd build`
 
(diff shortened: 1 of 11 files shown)


didikunz:
Sounds interesting.

ianshade:
This is an awesome idea. With some additional effort put into adjusting Qt Creator it could get better and easier for designers than Flash and HTML.
Do I understand correctly that currently each frame after being rendered goes from the GPU to the CPU and back to the GPU?

ServiusHack:
> Do I understand correctly that currently each frame after being rendered goes from the GPU to the CPU and back to the GPU?

Yes, this is also my understanding. I didn't know how else to implement it. Seems to me I always need to provide a `core::draw_frame` which contains the pixels of the frame.

dotarmin:
Hi

Thanks for your pull request, unfortunately it will not be merged because it's a really niched producer even if it's really cool.

Best regards,
Armin

didikunz:
@ServiusHack : You can try pulling it to the NRK version instead, but I would ask first, if they are interested to have it. At least then the work would not be wasted.

PR #996, opened 2018-05-14
fix(html) #991 Make url/filename validation less strict when using [HTML] prefix on a play command

### LINKED ISSUE #991: 2.2 [Linux] PLAY [HTML] 404
We are trying to play a locally stored HTML file. 
Command as follows:
PLAY 1-15 [HTML] "file:///PATH/TO/FILE/NAME.html?subtitle=test%20123&duration=3000"
Which used to work with 2.1, excluding the arguments (? and following) is of no avail

Julusian:
fixes #991


gizahNL:
Unfortunately it doesn't fix the 404 we are experiencing:
**Escape character is '^]'.
PLAY 1-15 [HTML] "file:///data/media/dusktv/subtitles.html?subtitle=test%20123&duration=3000"
404 PLAY FAILED
PLAY 1-15 [HTML] "file:///data/media/dusktv/subtitles.html"                                  
404 PLAY FAILED
**

PR #997, opened 2018-05-15
fix: #963 LOADBG with AUTO not working after seeking forward

### LINKED ISSUE #963: LOADBG with AUTO not working after seeking forward
Issue following commands:
```
    PLAY 1-10 clip1
    LOADBG 1-10 clip2 AUTO
    CALL SEEK 1-10 OUT -300 (or any command that seeks forward)
```
`Clip2` should start playing automatically once `clip1` comes to an end. It works fine so long as you don't reduce `clip1` playout duration (eg, using the `SEEK` command). In other words, if you seek forward so that the amount of time `clip1` was playing is less than the original duration of `clip1`, `clip2` will not start playing.

This is important for example for cue-to-cue pre-event run-through, where all you need to play are ins and outs to make sure transitions are OK.

It seems some sort of timer gets fired, which is tied to the duration of `clip1` and once the timer expires, `clip2` is transitioned from bg to fg.

### Expected Behaviour
Once `clip1` is over, the queued up `clip2` should start playing regardless of any seek commands.

### Current Behaviour
As described above, if you seek forward while `clip1` is playing, `clip2` will not auto play.

### Steps to Reproduce 
Issue following commands:
```
    PLAY 1-10 clip1
    LOADBG 1-10 clip2 AUTO
    CALL SEEK 1-10 OUT -300 (or any command that seeks forward)
```
Wait for `clip1` to end and `clip2` to not start playing.

### Environment
* CasparCG Server version: 2.1.0.472 dc2e94b Beta 2
* Operating system: Ubuntu 16.04 LTS

Julusian:
Fixes #963 

This fixes both cases raised in #963. Transition to the background layer now triggers on the correct frame at the end of the clip, at the end after a seek and with the clip length shortened

### DIFF
--- a/src/modules/ffmpeg/producer/ffmpeg_producer.cpp
+++ b/src/modules/ffmpeg/producer/ffmpeg_producer.cpp
@@ -107,13 +107,13 @@ struct ffmpeg_producer : public core::frame_producer_base
 
     std::uint32_t frame_number() const override
     {
-        return static_cast<std::uint32_t>(producer_->time() * format_desc_.fps);
+        return static_cast<std::uint32_t>(producer_->time());
     }
 
     std::uint32_t nb_frames() const override
     {
         return producer_->loop() ? std::numeric_limits<std::uint32_t>::max()
-                                 : static_cast<std::uint32_t>(producer_->duration() * format_desc_.fps);
+                                 : static_cast<std::uint32_t>(producer_->duration());
     }
 
     std::future<std::wstring> call(const std::vector<std::wstring>& params) override
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -916,6 +916,10 @@ struct AVProducer::Impl
 
     boost::optional<int64_t> duration() const
     {
+        if (duration_ != AV_NOPTS_VALUE) {
+            return av_rescale_q(duration_, TIME_BASE_Q, format_tb_);
+        }
+
         if (!input_.duration()) {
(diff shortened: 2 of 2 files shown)


src/modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -107,13 +107,13 @@ struct ffmpeg_producer : public core::frame_producer_base
 
     std::uint32_t frame_number() const override
     {
-        return static_cast<std::uint32_t>(producer_->time() * format_desc_.fps);
```
ronag: time is AV_TIME_BASE time, not frame number?

src/modules/ffmpeg/producer/ffmpeg_producer.cpp
```diff
@@ -107,13 +107,13 @@ struct ffmpeg_producer : public core::frame_producer_base
...
+        return static_cast<std::uint32_t>(producer_->time());
     }
 
     std::uint32_t nb_frames() const override
     {
         return producer_->loop() ? std::numeric_limits<std::uint32_t>::max()
-                                 : static_cast<std::uint32_t>(producer_->duration() * format_desc_.fps);
+                                 : static_cast<std::uint32_t>(producer_->duration());
```
ronag: duration is AV_TIME_BASE time, not frame number?

Julusian: No, they are both frame numbers. We are treating them as such in various other places too, such as when outputing osc state or in the CALL handling

PR #1002, opened 2018-05-18
fix(opengl) #989 #999 Fix transform, rotate and distort

### LINKED ISSUE #989: fill works as crop
With latest build using
MIXER 1-1 FILL 0.12 0 0.75 1 25 easeinsine
causes picture to crop instead of squeeze

### Expected Behaviour
picture should squeeze

### Current Behaviour
picture cropping occurs

load a picture on layer 1-1
send command
MIXER 1-1 FILL 0.12 0 0.75 1 25 easeinsine

### Environment
* CasparCG Server version: build 4 may 2018
* Operating system: Win7 -64

### Attachments

### LINKED ISSUE #999: Distort not working correctly since #966
Distort is not working correctly since after build 72943c537c9829c10c9491c929ae2b811e857f26

Might be linked to changes made in #966

Julusian:
Fixes #989 & fixes #999 

The key issue was no offset being passed in for the texture buffer data, so it was using the same coordinates as the vertex.

I have tested each of FILL, ROTATE and DISTORT individually and combined and in my testing they now behave the same as 2.1

### DIFF
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -378,11 +378,13 @@ struct image_kernel::impl
 
         // Draw
         switch (params.geometry.type()) {
-            case core::frame_geometry::geometry_type::triangles: {
+            case core::frame_geometry::geometry_type::quad: {
                 GL(glBindVertexArray(vao_));
                 GL(glBindBuffer(GL_ARRAY_BUFFER, vbo_));
 
-                GL(glBufferData(GL_ARRAY_BUFFER, static_cast<GLsizei>(sizeof(core::frame_geometry::coord)) * coords.size(), coords.data(), GL_STATIC_DRAW));
+                std::vector<core::frame_geometry::coord> coords_triangles { coords[0], coords[1], coords[2], coords[0], coords[2], coords[3] };
+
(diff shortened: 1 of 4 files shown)


src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -393,9 +395,9 @@ struct image_kernel::impl
                 GL(glEnableVertexAttribArray(tex_loc));
 
                 GL(glVertexAttribPointer(vtx_loc, 2, GL_DOUBLE, GL_FALSE, stride, nullptr));
-                GL(glVertexAttribPointer(tex_loc, 4, GL_DOUBLE, GL_FALSE, stride, nullptr));
+                GL(glVertexAttribPointer(tex_loc, 4, GL_DOUBLE, GL_FALSE, stride, (GLvoid*)(2 * sizeof(GLdouble))));
 
-                for (int i = 0; i < coords.size(); i += 3) {
+                for (int i = 0; i < coords_triangles.size(); i += 3) {
```
ronag: quads or triangles?

Julusian: frame_geometry is back to defining quads, as it makes the transform maths simpler as that is what it is already expecting.
but opengl doesnt support drawing quads anymore, so I am converting the quads to triangles after any transforms are calculated

ronag:
LGTM

PR #1016, opened 2018-06-05
fix: Count the commits only from the current branch

Julusian:
The version number scheme uses the number of commits as the fourth block. This was counting all the commits in the local tree, so it can vary massively between users or clones. This will now count only the commits in the current branch/ref and so should be more consistent.

@dotarmin It looks like Jenkins is doing a shallow clone, and so is always ending up with 4. Could you look at changing it to do a full clone of the branch to get this number to be correct?
If not then we may as well drop this number as it will be rather uninformative.


ronag:
Why are we using count and not just the rev?

Julusian:
I'm not against it being removed, but it is useful when someone reports an issue to quickly see how old their version is relative to another version or latest. It is much slower to check the revision hash

ronag:
Go with revision. A count is just confusing.

Julusian:
The count is removed.

I'm thinking that the tag currently being set to `Beta 1` is also a bit confusing/misleading, as it is hard to tell whether a build is actually the beta 1 build or it is a dev build from after beta 1. I think it should be left as dev and changed to something to produce a single build when a release is done.

PR #1020, opened 2018-06-16
fix(mixer) tweens dont run on uninitialized layers

### LINKED ISSUE #879: Mixer tweens don't work on empty layer
Not sure if it's a bug or feature but a mixer fill tween doesn't start until there is a producer on the layer.  Steps to reproduce...

MIXER 1 FILL .25 .25 .5 .5 60 easeinoutquad
PLAY 1 AMB

The tween will not start until the second command is sent.

Julusian:
Fixes #879 

Tested with FILL and OPACITY after CLEAR 1 and at startup.

I'm not 100% certain this is the best solution, but I think this is the better of the two solutions I came up with.

### DIFF
--- a/src/core/producer/stage.cpp
+++ b/src/core/producer/stage.cpp
@@ -69,11 +69,14 @@ struct stage::impl : public std::enable_shared_from_this<impl>
             std::map<int, draw_frame> frames;
 
             try {
+                for (auto& t: tweens_)
+                    t.second.tick();
+
                 // TODO (perf) parallel_for
                 for (auto& p : layers_) {
                     auto& layer     = p.second;
                     auto& tween     = tweens_[p.first];
-                    frames[p.first] = draw_frame::push(layer.receive(format_desc, nb_samples), tween.fetch_and_tick(1));
+                    frames[p.first] = draw_frame::push(layer.receive(format_desc, nb_samples), tween.fetch());
                 }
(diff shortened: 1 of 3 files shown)


src/core/producer/stage.cpp
```diff
@@ -106,8 +109,8 @@ struct stage::impl : public std::enable_shared_from_this<impl>
                 auto& tween = tweens_[std::get<0>(transform)];
                 auto  src   = tween.fetch();
                 auto  dst   = std::get<1>(transform)(tween.dest());
-                tweens_[std::get<0>(transform)] =
-                    tweened_transform(src, dst, std::get<2>(transform), std::get<3>(transform));
+                auto  index = std::get<0>(transform);
+                tweens_[index] = tweened_transform(src, dst, std::get<2>(transform), std::get<3>(transform));
```
ronag: DOes this actually change anything or is it just cosmetics?

ronag:
Looks good.

src/core/producer/stage.cpp
Julusian: Cosmetic. I meant to remove that bit as it is from the other way I tried. 
Do you want me to change it back?

ronag: Na, it's fine, just wanted to make sure

PR #1021, opened 2018-06-16
fix(mixer) tweens dont run on uninitialized layers

Julusian:
#1020

Fixes a build error due to missing parameter


Julusian:
@ronag I thought I had pushed that before I raised the previous PR...

PR #1023, opened 2018-06-19
Added error messages and solutions for running on Ubuntu 18.04

Fixy250185:
Added the solutions to some error messages I got while trying to run on Ubuntu 18.04 to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,15 @@ https://launchpad.net/ubuntu/+archive/primary/+files/libgcrypt11_1.5.3-2ubuntu4.
 Install central cgroup manager daemon (client library):
 sudo apt install libcgmanager0
 
+5. Error while loading shared libraries: libgconf-2.so.4
+Install GNOME configuration database system:
+sudo apt -y install libgconf2-4
+
+6. lib/libz.so.1: version `ZLIB_1.2.9' not found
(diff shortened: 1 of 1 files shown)


Julusian:
Thanks for doing this.

It is interesting that these have cropped up, especially the zlib issue. 
I know that the docker build images are using a beta build of ubuntu 18.04, so I am curious as to whether these issues would be fixed by updating those to newer versions

ronag:
@Julusian u ok this?

Fixy250185:
I've no idea about the docker images, I just thought i'd update the readme with some of the issues I got when I reinstalled Ubuntu 18.04 in case anyone else got the same problems

dimitry-ishenko:
Binary dependency hell... I think I've brought this up before, but IMHO CCG should have a proper .deb package

Julusian:
I have no issue this being merged, but I am wondering if this section of the readme would be better placed in the wiki? 

@Fixy250185 I didnt have any issues running on a fresh ubuntu18.04 install. Was the previous install of ubuntu18.04 up to date or had it not been updated since it was in beta perhaps? Im trying to understand how zlib could have gotten into this state. We package up the libs to avoid using the system installed ones, so it is weird that you had to do that step

Issue #1032, opened 2018-07-04
Feature request: Support for DataVideo TC–200


didikunz:
Support for DataVideo‘s [TC–200](https://www.datavideo.com/product/TC-200). It has been discussed earlier, but I can not find a feature request here. The TC–200 uses a unique way to pack the fill and key signal to a standard HDMI signal, by packing the U and V together into one color channel and use the spare one fpr the key. By adding this packing mechanism to the screen consumer we could add an option to play fill & key, with genlock (SDI) for standard laptops with only an HDMI output.

didikunz:
I would buy one of these and lend it for testing...

drakmor:
I have one, I can check something.

baltedewit:
Seems like this can be achieved with a simple shader.

Julusian:
@baltedewit I agree. 
The screen consumer isnt currently using shaders as it was simple enough to not need it, although I have been thinking that one would need to be added anyway to do key-only or sbs-key efficiently.

premultiply:
Does anybody have an exact specification how this is implemented?
There can be several ways. I think a YUV 4:4:4 signal is sent through HDMI where 4:2:2 is fill and the remaining "0:2:2 space" is recombined as key channel. So as result after remapping you will get 4:2:2 + 4:0:0 at SDI outputs.

drakmor:
You can download software https://www.datavideo.com/product/CG-200 and test it on second HDMI monitor.

Julusian:
This will be nice and easy with #1040 as it will just be a new option to use some colour conversion inside the shader.

didikunz:
Just to know: Will this be implemented in the 2.2 version?

baltedewit:
> Just to know: Will this be implemented in the 2.2 version?

Definitely not in 2.2 since that is pretty much in a feature freeze now. Will be 2.3 at earliest, given that there is a developer interested in implementing it.

Julusian:
This is done via #1178 in 2.3

PR #1040, opened 2018-07-13
feat(screen): Use shader for screen consumer, to allow gpu to perform key-only step

Julusian:
Primarily this avoids the aligned_memshfl() when set to key-only.

It also allows #805 and #1032 to be done more cleanly without any more work on the cpu


ronag:
Nice!

Issue #1008, opened 2018-05-30
Feature under development: Sting transition


Julusian:
CasparCG supports some basic transitions between producers on a layer. The range of options is rather limited and doing anything more complex requires running them manually in seperate layers and using mask layers which quickly gets complicated. 
This could be made a lot cleaner by extending the transition producer with a sting style. This would utilise a couple of FFmpeg Producers to generate the mask to apply to the old producer and an overlay layer.
Doing this would also allow the user to make their own wipe/reveal patterns as well as sting transitions.

Key features:
* Define a video clip to use as the mask doing the transition from foreground- to background clip. The mask clip will be luma-keyed and white fill to paint out the region for the upcoming clip to show.
* Optionally define a video clip with alpha to play as the sting-wiper overlay keyed on top of the transition.
* The duration of the transition should be taken as the duration of the mask file
* Specify the trigger point (the point during the transition to start the destination producer)

The motivation behind this to provide the sting transition similar to that on an ATEM, using a wipe between the producers, but with a soft edge (or any kind of shape).

PeterAkakpo:
@Julusian what is a sting transition, how does the visual look like

premultiply:
A stinger transition is an keyed animation which runs during/over a classic transition between two sources/clips. This can be seen often in sport events when switching from live camera to slowmotion replay and back to live again.

jesperstarkar:
This feature consists of two parts:
1) The mask transition: Instead of just mixing, wiping, sliding between two clips, you can now use a video file to mask the transition. Create the mask clip however you'd like. Black pixels show the current playing clip, white pixels show the next playing clip and values between do alpha blending. This way you can "paint" out the old clip with the new however you'd like. The new clip is playing in the background, and plays as it gets "painted in".

2) The wipe overlay: At the same time as this happens, a fourth clip can play at a new top layer, as decor covering up the ongoing transition. This is behaves like @premultiply describes.

PR #1042, opened 2018-07-16
feat: Sting transition #1008

Julusian:
This allows for doing more complex transitions than are currently supported.
It uses a video file as the mask (it will work with other producers if they report a length)
And another optional video file (or any producer) as the overlay artwork.

The length is dictated by the mask file, if the overlay is longer it will get cut off.
There is an optional parameter to delay the start of the destination producer. 

Example usage:
```
PLAY 1-10 AMB STING wipe_mask 0 wipe_overlay

explanation: 
STING (mask_filename) (dest_start_delay) (overlay_filename)
```

Here is not great looking demo of it in action, along with the source assets. Hopefully this will help others understand how it all fits together when used.

Demo: https://drive.google.com/open?id=1XDuZXug2kepzFLK9MGWdAeNsEKTTv6pM
Assets: https://drive.google.com/open?id=1cGvX8CfUwg89AbQ9PvU7iLW-g-YPp1wr

### DIFF
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -280,6 +280,7 @@ struct image_kernel::impl
         shader_->set("keyer", params.keyer);
 
         // Setup image-adjustements
+        shader_->set("invert", params.transform.invert);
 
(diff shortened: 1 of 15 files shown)


jesperstarkar:
Wiki needs to be updated with AMCP changes.

PR #1048, opened 2018-07-17
Cloud based builds

Julusian:
This is something that will want discussion, I think was back when 2.2 was being started, but I am finding the current build process to be a bit limiting. Primarily, it is very hard to notice build issues caused by a commit or PR without someone trying to build it or @dotarmin noticing the SVT build server threw an error.
This PR is something I was spent a couple of hours playing with the other day. I don't mind if it is decided to stick with the current Jenkins process, but thought it worth review. 

One solution for this is to use a cloud based ci provider for the building.
This PR is largly some linux script refactoring to make the version numbers be defined in one place, the actual build scripts for both platforms are pretty simple, and are also good for using locally to produce a build zip file.

The only free cloud ci I am aware of that supports windows is appveyor, so I have been playing with that. The total build time ends up being 20-25mins (~10mins per platform, free tier won't run concurrent)

Another benefit of doing builds like this is that it will add some status badges to commits, which makes it easy to see if a commit builds, and does the same for PRs too. (Example: https://github.com/Julusian/CasparCG-Server/pull/2 https://github.com/Julusian/CasparCG-Server/commits/appveyor)

Each build is also producing a zip file for each platform (eg https://ci.appveyor.com/project/Julusian/casparcg-server/build/2.2.15-appveyor2/job/o58acv3ribggwjkr/artifacts and https://ci.appveyor.com/project/Julusian/casparcg-server/build/2.2.15-appveyor2/job/o4a3hhf7t2h8shh8/artifacts) These can also easily be pushed to builds.casparcg.com via whatever method SVT prefer, without adding that to the repo (it can be defined in the appveyor ui)

On linux, there is also a step which is rebuilding the docker base images if they do not yet exist, and I will make it push the new images to dockerhub, and so will also remove the current manual step to rebuild those when the versions are bumped.

And continuing further, users will be able to easily enable this whole process for their own repos with no additional work, which I would really like to have for any experimental/testing branches I have

What do people think about using this? I think it will be a good step to being more open and make the build process more transparent.

A later step (which I shall do as another PR) I would also like to do is to add a step which will mark the commit/PR as failed if clang-format detects anything wrong with a source file, to enforce the common styling across the project and reduce churn caused by formatting being corrected in later commits

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 CasparCG Server
 ===============
 
+[![Build Status](https://dev.azure.com/julusian/casparcg/_apis/build/status/Julusian.CasparCG-Server?branchName=azure-pipelines)](https://dev.azure.com/julusian/casparcg/_build/latest?definitionId=1?branchName=azure-pipelines)
+
 Thank you for your interest in CasparCG Server, a professional software used to
 play out and record professional graphics, audio and video to multiple outputs.
 CasparCG Server has been in 24/7 broadcast production since 2006.
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,69 @@
+jobs:
+- job: ubuntu_1804_docker
+  displayName: 'Ubuntu 18.04'
+  strategy:
+    matrix:
+      Default: {}
+      Clang:
+        CC: clang
+        CXX: clang++
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - checkout: self
+    clean: false
+    fetchDepth: 1
+    lfs: false
+  - bash: |
+      export PROC_COUNT=$(nproc)
+      ./tools/linux/ensure-base-images
+      ./tools/linux/build-in-docker
+    #   ./tools/linux/extract-from-docker
+    displayName: 'build in docker'
+#   - task: ArchiveFiles@2
+#     inputs:
+#       rootFolderOrFile: 'casparcg_server'
+#       archiveType: 'zip' # Options: zip, 7z, tar, wim
+#       archiveFile: 'artifacts/casparcg_server_linux.zip'
+#     displayName: 'make archive'
+# TODO - publish zip to ftp/similar
+
+- job: windows
+  displayName: 'Windows VS2017'
+  pool:
+    vmImage: 'vs2017-win2016'
+  steps:
+  - checkout: self
+    clean: false
+    fetchDepth: 1
+    lfs: false
+  - task: CMake@1
+    inputs:
+      workingDirectory: "$(Build.BinariesDirectory)"
+      cmakeArgs: "-A x64 $(Build.SourcesDirectory)/src"
+    displayName: 'prepare build'
+  - task: NuGetCommand@2
+    inputs:
+      command: 'restore'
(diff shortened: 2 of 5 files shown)


petterreinholdtsen:
Is there any reason not to push this change into master, and fix any fallout if anything is discovered?

Julusian:
As no-one has said raised any concerns with this proposal, I am assuming that means t no-one has a problem with this being done?

It has been rewritten to use Azure Pipelines instead of appveyor, as it is a much better system and also supports mac building (in case support for that gets added).
This has also reduced the build time down to ~15mins, which matches what the SVT servers are doing. I haven't yet gotten the binaries uploading anywhere, but I shall continue to work on that once this is merged.

Demo PR at https://github.com/Julusian/CasparCG-Server/pull/5 it also shows the status being applied to the commit history. 
Build history on azure: https://dev.azure.com/julusian/casparcg/_build?definitionId=1

petterreinholdtsen:
[Julian Waller]
> Demo PR at https://github.com/Julusian/CasparCG-Server/pull/5 it also
> shows the status being applied to the commit history.

Very cool.  Can you extend the test to compile with both gcc and clang
on Ubuntu?  It would be good to know the code is portable across both
these compilers, as they are picky in different areas. :)

Btw, is there some method to scroll sideways in the build log to get to
see the end of long lines?

-- 
Happy hacking
Petter Reinholdtsen

Julusian:
> Can you extend the test to compile with both gcc and clang
on Ubuntu? 

Sure, Azure supports build matrix for jobs like Travis does. Is it just adding the CC and CXX variables to cmake to do it? And it can use the same version of boost/ffmpeg etc?

> Btw, is there some method to scroll sideways in the build log to get to
see the end of long lines?

I get a horizontal scrollbar. Its not very visible, but should be at the bottom of the screen

Julusian:
@petterreinholdtsen 
Commit adding a clang build on ubuntu18.04: https://github.com/Julusian/CasparCG-Server/commit/6a874ab4a145fe02a149bf81632bcf44a9f3f4e4

I will hold that commit back until the build passes, in case it takes a while to resolve

petterreinholdtsen:
The build status look great.  I hope this get integrated into the official repo soon. :)

petterreinholdtsen:
Is there anything wrong with this change, causing it to not be integrated into master?

Julusian:
@petterreinholdtsen not that I am aware of, I will merge it when I have a time. I will need to update urls to not point to my personal azure pipelines project, so is a small amount of work

petterreinholdtsen:
Very good.  I look forward to having the build status for each pull request available from https://github.com/CasparCG/server/pull/.

petterreinholdtsen:
@Julusian Hi, any progress on updating and merging this patch?

chrisspiegl:
Any movement here?

Julusian:
No, there has been no progress on this.
As there are already some automated builds being published at http://casparcg.com/builds/CasparCG%20Server/, there is no pressing need to do this change, so it has been hard to motivate.
I would still like this to be done, but I need to rework it to use github actions before I can try and convince the others.

Julusian:
The builds are now being produced by github actions

Issue #1007, opened 2018-05-30
Proposed feature for discussion: OSD Producer


jesperstarkar:
The main idea is to be able to natively render text overlays for timecode and statuses on separate layers (channels/outputs) based on the current status of media and playback on the main layer/channel/output. We’re starting by focusing on the FFmpeg Producer and video playback.

**Example usage:**

- Your main video playback layer of choice is 1-10
- You then operate 1-10 as normal
- You want to set 2-99 as your OSD layer

**Depending on the MODE of the OSD Producer you will automatically get either...**
... a preview (route) with information of the currently playing clip on 1-10 (foreground)
 
or 

... the first frame as a freeze-frame with information of the currently loaded(bg) clip of 1-10.

**The MODE flag can be set to** 
1. show foreground by default, and background only if you have no foreground
2. show foreground only
3. show background only

The text information and layout will in this first iteration not be configurable. Our suggestions are:
<img width="622" alt="screen shot 2018-05-30 at 09 51 10" src="https://user-images.githubusercontent.com/6236874/40706760-0623e8ec-63ef-11e8-997c-287c94bba680.png">
Where the VT-a is a user-defined label for the destination layer, TC 00:00:15:14 shows the duration of the loaded clip and 1-10 shows the destination layer.

<img width="614" alt="screen shot 2018-05-30 at 09 50 22" src="https://user-images.githubusercontent.com/6236874/40706732-f31dd3b6-63ee-11e8-919d-88aeb1445048.png">
Where the VT-a is a user-defined label for the destination layer, the TC 00:00:14:15 shows the current running TC of that clip and 1-10 shows the destination layer.

jesperstarkar:
This feature also solves another great thing other than text overlay: background preview. We want to enable the opeator to have an instant preview of a clip loaded in background (cued up) to be confident that the next clip is ready to go, and the correct clip is loaded.

didikunz:
It would be complete if it also shows the remaining time until the clip is finished. Either configurable or next to each other. Normally the operator is not too much interessted on the running timecode but wants to know hoe long it will run.

premultiply:
You may also think about using an OSD producer for subtitling.

Further possible improvements / additions to OSD overlay:
- Audio Level Meter Bars
- Countdown / remaining time
- Indication for Loop-Mode
- Visualisation of remaining time / progress with a bar

respectTheCode:
[...]

PR #1054, opened 2018-07-19
#1007 feat(route) route first frame of layer background producer

Julusian:
```
PLAY 2-10 route://1-10 BACKGROUND
PLAY 2-10 route://1-10 NEXT
```
This allows you to view the first frame of the background producer. Intended as a preview of the source to make it possible to visibly show to users that the clip is loaded and it is the correct clip.
BACKGROUND mode will only ever show the first frame of the layer background producer,
NEXT mode will show the first frame of the background clip if one is loaded, otherwise will show the foreground. So if you LOADBG then PLAY, the clip will continue playing.

@ronag I have tried to avoid calling receive on the background producers as much as possible, to avoid adding additional cpu usage for who doesn't use this feature. It would be a bit cleaner to always generate them though.

This also fixes a bug where CLEARing a layer which was being routed would cause that route to freeze on last frame until any operation was performed on that empty layer (even a CALL would cause it to unfreeze). This was from the layer no longer existing, so there was no frame to send to a route, when we should have been sending an empty frame.

### DIFF
--- a/src/core/mixer/mixer.cpp
+++ b/src/core/mixer/mixer.cpp
@@ -69,12 +69,12 @@ struct mixer::impl : boost::noncopyable
     {
     }
 
-    const_frame operator()(std::map<int, draw_frame> frames, const video_format_desc& format_desc, int nb_samples)
+    const_frame operator()(std::vector<draw_frame> frames, const video_format_desc& format_desc, int nb_samples)
     {
         for (auto& frame : frames) {
-            frame.second.accept(audio_mixer_);
-            frame.second.transform().image_transform.layer_depth = 1;
-            frame.second.accept(*image_mixer_);
+            frame.accept(audio_mixer_);
+            frame.transform().image_transform.layer_depth = 1;
+            frame.accept(*image_mixer_);
         }
 
         auto image = (*image_mixer_)(format_desc);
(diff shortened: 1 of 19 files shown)


Julusian:
> @ronag I don't understand what exactly this is fixing/doing?

Sorry, I missed out a chunk of the description. hopefully that helps explain the purpose?

TomKaltz:
This should also allow OSC to get background producer information correct?

Julusian:
> @TomKaltz This should also allow OSC to get background producer information correct?

If we always generate the first frame of any background producer then it should give us all the osc data wanted and it would be simple to then expose that. 
I didn't want to do that as part of this as I'm not 100% certain of the performance implications of always generating the first frame of the background producers, or if there would be any other issues arising from doing so

premultiply:
There is one thing to make shure when the background producers first frame is brought to the foreground: The first frame and audio has to be played again.

And maybe the "first frame" implementation is even usefull for generating thumbnail files for any producer/content.

Julusian:
@premultiply good point. I havent checked that, and suspect this code will not handle that correctly

TomKaltz:
meta #941

Julusian:
I have rebased and reworked this a little with some fixes.

I can confirm that the first frame is still played on the output after being shown via a background route.

Additionally, any real-time sources (decklink, html etc) will now show motion, as the route shows the first frame of their buffers. This is to ensure it is a true representation of what will be shown on PLAY rather than the true first frame of the producer which is likely very different.

The transition will always show the first frame of the destination producer instead of the transition, as the transition will almost always start with showing just the source

PR #1064, opened 2018-08-06
Append preceding nodes name to digit-only nodes

pkeuter:
See #1062 for more info.

> Number-only nodes are not allowed in the XML spec so the INFO-command is currently breaking some XML-parsers. I've changed this so that the preceding name will be appended to the number. For example <0></0> will now be <layer_0></layer_0>

The regex looks for `.name.<digit>.` and replaces it with `.name.name_<digit>.`. This makes the INFO output valid XML.

### DIFF
--- a/src/protocol/amcp/AMCPCommandsImpl.cpp
+++ b/src/protocol/amcp/AMCPCommandsImpl.cpp
@@ -59,6 +59,7 @@
 #include <memory>
 
 #include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/regex.hpp>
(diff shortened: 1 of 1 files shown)


pkeuter:
@Julusian @ronag Is there anything holding you guys back from merging this in? I see that more people are dealing with this issue and it does a simple regex replace to fix the output to conform to spec. If necessary, I can do some changes to it.

ronag:
I don't like it but don't have any better ideas. It's better than what is currently and I won't be spending more time on this. So I guess this solution is good enough :).

PR #1066, opened 2018-08-08
fix(build): Linux fails on boost::lexical_cast not being defined in monitor.h

Julusian:
@ronag any idea why it builds without this only on windows?


ronag:
pre compiled headers

PR #1076, opened 2018-09-05
Automatically detect video input

### LINKED ISSUE #714: Decklink Producer only supports 1:1 format
decklink producer currently needs to run in same format as channel

pkeuter:
This PR enables the subscription to input changes and sets the Decklink input accordingly. 

Resolves #714


ronag:
@Julusian comments?

pkeuter:
By the way there might be some improvements. Most importantly, the buffer drains quite often, but this PR doesn't affect that issue, it was there beforehand. Could be a sync thing? If someone has an idea I'd like to know!

Also, we probably want the `FORMAT` override in the future, but this can be a relatively easy addition.

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -122,10 +122,6 @@ struct Filter
                 }
             }
 
-            if (format_desc.field_count == 2) {
-                filter_spec += ",bwdif=mode=send_field:parity=auto:deint=all";
```
ronag: What about this?

pkeuter: It doesn't seem to do anything. It deinterlaces when the output format is deinterlaced, but since the pipeline is now fully progressive, correct? If you notice issues with it, I can put it back...

Julusian: This will still be needed when the input is interlaced. Before it was checking the output format as the input and output were guaranteed to be the same, but it should really have been checking the input format.

It will need to be tweaked though, as it will need to be always enabled to be able to handle format changes. It might be possible to change the end to `deint=interlaced` to ensure we dont waste cpu attempting to deinterlace a progressive source

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -345,7 +341,7 @@ class decklink_producer : public IDeckLinkInputCallback
         graph_->set_text(print());
         diagnostics::register_graph(graph_);
 
-        if (FAILED(input_->EnableVideoInput(mode_->GetDisplayMode(), bmdFormat8BitYUV, 0))) {
+        if (FAILED(input_->EnableVideoInput(mode_->GetDisplayMode(), bmdFormat8BitYUV, bmdVideoInputEnableFormatDetection))) {
```
Julusian: I wouldn't be surprised if this call fails when the card doesn't support format detection, so it would be nice to check that.

Julusian:
@pkeuter have you tested this with a source running at a different resolution?

A `FORMAT` override would be nice, but I am happy for that to be a later PR

pkeuter:
@ronag @Julusian, I have not tested all use cases extensively as necessary, because I am close to my holidays and needed this for a client. But if you guys can give me some time, I will test more situations and update this PR. Personally I do think this is an important feature, so it would we nice if it can still make it into the release of 2.2.

src/modules/decklink/producer/decklink_producer.cpp
premultiply: I dont think that deint=interlace is working correctly in this case as I think it relies on frame metadata which might not be availible in this case as SDI does not support anything like this.

But give it a try.

But I think the bwdif filter needs to be set to deint=all for interlaced input formats and skipped at all for progressive input formats.
But it should never produce bad quality on progressive input even if set to deint=all for any input format as the internal motion estimation skips the deinterlacing when there is nothing to do. Of course the motion estimation does cost some CPU time...

Julusian: @premultiply we are having to convert the decklink frame data into a frame that can be fed into ffmpeg, so it should be easily possible for us to set the metadata of those frames as we do so

src/modules/decklink/producer/decklink_producer.cpp
Julusian: I was curious and have some cards I suspected do not support format detection and can confirm that this fails with that card.

```
[2018-09-08 19:25:15.785] [info]    Received message from Console: play 1-10 decklink 1\r\n
[2018-09-08 19:25:15.831] [warning] Device does not support video-format: 1080i50
[2018-09-08 19:25:15.860] [warning] Decklink producer not supporting detection
[2018-09-08 19:25:15.871] [error]   Exception: c:\code\casparcg22\src\modules\decklink\producer\decklink_producer.cpp(351): Throw in function __cdecl caspar::decklink::decklink_producer::decklink_producer(const struct caspar::core::video_format_desc &,int,const class caspar::spl::shared_ptr<class caspar::core::frame_factory> &,const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &,const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &,bool)
[2018-09-08 19:25:15.871] [error]   Dynamic exception type: class boost::exception_detail::clone_impl<struct caspar::caspar_exception>
[2018-09-08 19:25:15.871] [error]   [struct boost::errinfo_api_function_ * __ptr64] = EnableVideoInput
[2018-09-08 19:25:15.871] [error]   [struct caspar::tag_msg_info * __ptr64] = Intensity Pro [1|1080i5000] Could not enable video input.
```

It's easy to check for suppport though, I used the following to add a temporary line to the log
```
int status = 0;
        if (FAILED(attributes_->GetFlag(BMDDeckLinkSupportsInputFormatDetection, &status)) || !status) {
            CASPAR_LOG(warning) << L"Decklink producer not supporting detection";
        }
```

pkeuter: Hey @Julusian, thanks for testing! I'll add that to the PR when checked the other input formats / conversions.

pkeuter:
@ronag @Julusian I've tested with fps and format conversion (PAL->1080i5000), 1080i5000->PAL, PAL->1080p5000, 1080p5000->PAL). They all work. I've also added back the deinterlacer, but based on the correct parameters this time.

Also added; the `FORMAT` parameter, for specifying a manual format override.

Please check if there are things I've missed, but otherwise this can be merged.

pkeuter:
@ronag @premultiply I've implemented the changes.

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -323,14 +341,27 @@ class decklink_producer : public IDeckLinkInputCallback
...
+        , input_format(format_desc_)
+        , vfilter_(vfilter)
+        , afilter_(afilter)
     {
+        // use user-provided format if available, or choose the channel's output format
+        if (!format.empty()) {
+            input_format = core::video_format_desc(format);
+        }
+
+        mode_         = get_display_mode(input_, input_format.format, bmdFormat8BitBGRA, bmdVideoOutputFlagDefault);
```
Julusian: This wasn't you, but can you change bmdFormat8BitBGRA to bmdFormat8BitYUV. 
We shouldn't be checking if the device supports a mode in one pixel format then start it using another

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -323,14 +341,27 @@ class decklink_producer : public IDeckLinkInputCallback
...
+        video_filter_ = Filter(vfilter_, AVMEDIA_TYPE_VIDEO, format_desc_, mode_);
+        audio_filter_ = Filter(afilter_, AVMEDIA_TYPE_AUDIO, format_desc_, mode_);
+
+        field_count_ = mode_->GetFieldDominance() != bmdProgressiveFrame && mode_->GetFieldDominance() != bmdProgressiveSegmentedFrame ? 2 : 1;
```
Julusian: I don't think we want to do this check any more. The only thing we use field_count_ for is to set the size of the frame_buffer, so the field could be removed.

As the format could change after initialisation we should prepare the buffer for that and do a `frame_buffer_.set_capacity(3);` below instead

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -382,10 +424,30 @@ class decklink_producer : public IDeckLinkInputCallback
...
+        auto mode = newDisplayMode->GetDisplayMode();
+        auto fmt     = get_caspar_video_format(mode);
+
+        auto new_fmt     = core::video_format_desc(fmt);
+
+        CASPAR_LOG(info) << print() << L" Input format changed from " << input_format.name << L" to " << new_fmt.name;
+
+        input_format = new_fmt;
+
+        input_->PauseStreams();
```
Julusian: All of these input calls should be checked for success like we are in the constructor.
If any of these fail we will currently be ending up with a broken producer without anything in the logs.

It will need to be done in a try catch block like this so that the exception gets back to us correctly.
```
try {

} catch (...) {
    exception_ = std::current_exception();
    return E_FAIL;
}
```

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -382,10 +424,30 @@ class decklink_producer : public IDeckLinkInputCallback
...
+
+        // reinitializing filters because not all filters can handle on-the-fly format changes
+        video_filter_ = Filter(vfilter_, AVMEDIA_TYPE_VIDEO, format_desc_, mode_);
+        audio_filter_ = Filter(afilter_, AVMEDIA_TYPE_AUDIO, format_desc_, mode_);
+
+        // reinitializing video input with the new display mode
+        input_->EnableVideoInput(
```
Julusian: Does audio still work after this with no call to `input_->EnableAudioInput`?

pkeuter: Yes it does, I've double checked and it works. No need for `EnableAudioInput`.

pkeuter:
@Julusian Thanks for the advise! I've gone through your reviews and resolved them. I've also made some other changes to avoid the filtergraph not understanding the conversion. There will be no error messages anymore when switching between formats.

pkeuter:
@Julusian No problem! I have corrected those points. And I also added the audio check in the `VideoInputFrameArrived` method. When running multiple streams on a insuficiently equipped machine, this will crash CasparCG eventually. 

According to the docs of the Decklink SDK:

> 
> New audio packet-only valid if audio capture has
> been enabled with IDeckLinkInput::EnableAudioInput
> The audio packet will be NULL under the following
> circumstances:
> - Audio input is not enabled.
> - If video processing is sufficiently delayed old
>  video may be received with no audio.
> 
> 

We always enable the audio input, so the only reason for there being no audio is the second one, which we do not need with a video buffer of 3 frames.

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -432,29 +432,39 @@ class decklink_producer : public IDeckLinkInputCallback
...
+            if (FAILED(input_->FlushStreams())) {
+                CASPAR_THROW_EXCEPTION(caspar_exception() << msg_info(print() + L" Unable to flush input stream.")
+                                                          << boost::errinfo_api_function("FlushStreams"));
+            }
+
+            if (FAILED(input_->StartStreams())) {
+                CASPAR_THROW_EXCEPTION(caspar_exception() << msg_info(print() + L" Unable to start input stream.")
+                                                          << boost::errinfo_api_function("StartStreams"));
+            }
             input_->StartStreams();
```
Julusian: You forgot to remove a line :)

PR #1110, opened 2018-11-14
2.0.x

mryvz:
local image files cannot be loaded webgl context. 

added ;
> browser_settings.universal_access_from_file_urls = cef_state_t::STATE_ENABLED;

related link below,
https://answers.unrealengine.com/questions/674383/tampering-allow-files-access-from-cef-browser-and.html


petterreinholdtsen:
Will this change allow javascript from remote web sites to access local files?  Would it be possible to make this setting a value to configure in the casparcg.config file?

Julusian:
Unfortunately we are not accepting any fixes for anything older than 2.2 anymore.

If this is still a problem in 2.2, then we will accept a PR against master to resolve this

Issue #1109, opened 2018-11-14
Bluefish devices capture interlaced formats incorrectly 


jamesdwise:
### Description
I am trying to improve the implementation of capture of interlaced formats when using a Bluefish device.  Currently capturing from an interlaced format into a channel of the same type has issues.

### Solution suggestion.
I need to better understand what the Caspar Engine expects / wants from the Bluefish producer.

Eg. currently the bluefish producer provides 25 1920x1080 frames to Caspar when capturing 1080i50 into a channel configured as 1080i50.  However I have discovered that Caspar much prefers, if ithe bluefish producer provides 50 field buffers, ie. 1920x540 frames, which alternately contain the appropriate fields.   But this does not look 100% correct on either the screen consumer or a SDI consumer.

---

### Additional information
Based on what i can see, via the screen and sdi consumers, it looks like  some of the fields are not being fully handled or handled incorrectly.
For example it looks like the sdi output ( via an sdi consumer) appears to only be showing data from 1 field. 

Can anyone provide any insight as to what the bluefish producer should be supplying to caspar

jamesdwise:
How should I create the 50p frame from the i50 input?

basic line doubling for each field?
interpolation of f1/f2?
return an entire p50 frame containing both the field 1 and field 2?

premultiply:
You can use ffmpeg for high quality deinterlacing and conversion like in decklink producer.

baltedewit:
> How should I create the 50p frame from the i50 input?

Interpolation of fields will give the best results. Using e.g. bwdif filter from ffmpeg with the send_field option. (this is how the decklink producer does it)

ronag:
@jamesdwise can we close this?

jamesdwise:
yeah this can be closed now ;)

PR #1115, opened 2018-11-21
improvements to the bluefish producer, and the way it handles interlaced sources

jamesdwise:
This PR  fixes the issues raised in #1109 

Now handling interlaced sources better
- giving caspar a full frame for each field.
- giving audio at correct rate ( ie. field rate)
- using HW de-interlacing instead of ffmpeg filters, which should hopefully give best quality on channels that are all the same format.


ronag:
LGTM, thanks

PR #1116, opened 2018-12-03
Disable html module when CEF is unavailable.

petterreinholdtsen:
Add FindCEF cmake rule to detect when CEF is available.


Julusian:
It would be better to require the user to explicitly state they do not want to include CEF as a build option. Otherwise it risks producing an 'incomplete' build if CEF wasn't found but should have been (eg typo in path name)

petterreinholdtsen:
While I would be happy to make such change, I do not know cmake well enough to know how to do it.  Any clues to spare?

petterreinholdtsen:
I rewrote the rules to make enabling the HTML module a cmake option.  It is enabled by default.  I hope it addresses your concerns. :)

petterreinholdtsen:
It would be great if this change was included in the next release.  Is there any hope to have it included?

PR #1122, opened 2018-12-05
Update of license info

jamesdwise:
Update of license file on each of the interop headers for  GPL compliance.


petterreinholdtsen:
This pull request do not only include license updates, but also code changes. How come?

Also, the copyright year change from 2014 in BlueVelvetCUtils.h and 2017 in BlueVelvetC.h to 2018 without any copyrightable changes to the files.  This seem strange, given that the copyright year should reflect when the copyright terms of the content starts.

Issue #1004, opened 2018-05-23
Feature request predifined layers


MauriceVeraart:
Feature request

in #871 @Julusian  is mentioning predefined layers to play / initialize
while the server starts up
although specificly made for the timecode option
it would also be a nice option in general use.

like for instance load a colorbar after startup 

```
<channel>
  <timecode>free [free|clock|layer]</timecode>
  <timecode_offset>???</timecode_offset> 
  <timecode_layer>10</timecode_layer>
  <predefined_layers>
    <layer id="0">BG_LIGHT LOOP</layer> // video file
    <layer id="10">DECKLINK 5 FORMAT 1080i5000</layer>
    <layer id="20">[HTML] http://localhost:3000/overlay.html</layer>
  </predefined_layers>
</channel>
```

i bet @Julusian  already made this ;)

Julusian:
The predefined layers I have done was added for timecode, but does not have to be used for timecode purposes and can have many layers defined as shown in that example.
It is something I have wanted for a while (and I know others have too), so I made sure to make it generic.
A 2.2/2.3 PR is planned, so this will be included in that.

dotarmin:
> @Julusian It is something I have wanted for a while (and I know others have too), so I made sure to make it generic. A 2.2/2.3 PR is planned, so this will be included in that.

Closing this and referring to #871

Issue #872, opened 2018-02-22
Feature under development: Scheduled commands


jesperstarkar:
In an ongoing project we're working on adding scheduled commands. @Julusian is the lead programmer. Related to #871 and #850.

Based on a channels timecode, all channel (and layer)-related commands can be buffered and scheduled to be executed by the server in the future, triggered by a timecode.

**Proposed API**

```
>>TIME [channel] 
<<2xx [some positive response]
<<[00:00:00:00] //some timecode
returns current timecode of the channel
```

```
>>SCHEDULE [timecode] [token] [command] 
<<2xx [some optimistic positive response]
Schedules a command for a given timecode. Multiple commands can be stacked into the same frame/timecode. Token needs to be unique. Scheduling a new command (at any timecode) with an existing token removes the old scheduled command with the same token.
```

```
>>UNSCHEDULE  [token]
<<2xx [some positive response] // or
<<xxx [some negative response] // if token is missing or in the past
Removes a scheduled command
```

```
>>UNSCHEDULE  ALL
<<2xx [some positive response]
Removes all scheduled commands
```

```
>>LIST SCHEDULED
<<2xx [some positive response]
<<[00:00:00:00] [token] //list, multiple lines
Return a list of all scheduled timecodes with their tokens, in chronological order. Only commands of the future, never current or previous ones.
```

```
>>LIST SCHEDULED [timecode]
<<2xx [some positive response]
<<[00:00:00:00] [token] //list, multiple lines
Return a list of scheduled timecodes with their tokens for a given timecode, in chronological order. Only commands of the future, never current or previous ones.
```

**Other ideas**
- command `EXPEDITE [token]` to trigger a scheduled command immediately.
- relative timing offset in timecode (+00:00:03:00)

The first prototype implementation is working well and it is insanely cool to be able to batch commands like this.

premultiply:
I would not recommend to do it by waiting for a specific or absolute timecode to occur. Because you can never trust timcodes to be monotonic and there are issues with wrap around, commands sent to late and played out 24h later etc.

I woud suggest to do it with relative timing.
Other commands like "DELAY" would benefit from it.

jesperstarkar:
You do have a valid point on the 24h paradox. However, I think we will aim for absolute timecodes, and leave the housekeeping of validating scheduled commands to the client. 

[...]

Issue #871, opened 2018-02-22
Feature under development: Timecode support


jesperstarkar:
In an ongoing project we're working on adding timecode support. @Julusian is the lead programmer.

**Proposed features**

jesperstarkar:
Realted to #821 and #772

premultiply:
Timecode has to be bound to the channel frame output which must be in sync with a primary producer (e.g. decklink) providing the channel clock. This channel clock may or may not genlocked (external reference). And this reference signal may or may not be locked to UTC time. So the only thing you can do is to sync the timecode output once on server start to system time and continue with counting it up on each frame. Else you would run into problemes with skipped or duplicated timecode values.
And keep in mind that each channel is not genlocked to each other. It may have an independant clock source so it needs its own timecode generator...

TomKaltz:
Yeah timecode should be based off of output frame counting shouldn't it?  It's my understanding that ffmpeg producer works via PTS and it's possible to drop or duplicate frames since it's just trying to air based on PTS and not tick successive frames to playout.  @ronag can you shed some light on this?

ronag:
FFMPEG bases playback on time stamps.

TomKaltz:
Yeah I guess it doesn't matter.  There is a chance timecode could become discontinuous due to a performance-related skipped frame so worrying about a skipped or dupped frame due to channel PTS mismatch is acceptable.

Julusian:
> So the only thing you can do is to sync the timecode output once on server start to system time and continue with counting it up on each frame

I disagree with that. If the channel timecode doesn't update to input timecode at regular intervals, then you risk your clock drifting, especially if the server is dropping frames. Also what would happen if your timecode source wasnt running when you started up your caspar server?
If the input timecode isnt monotonic then I would expect the output timecode to follow that as it is the reference timecode that we are following.

> And keep in mind that each channel is not genlocked to each other.

Yes, the intention is that each channel will be able to have its own timecode source.

premultiply:
Worse case: Input and output may not be in sync and channels may not be in sync and timecode may jump forward and backwards, freezes or is missing at. Happy timing!
Good luck and have fun with it. ;-)

maceyldn:
Can i ask, Are you planning on VITC support to the Decklink/Bluefish Producers,
This would be really helpful in installations running SCTE 104/35 for Ad Insertion for those using it as a main chanel playout.

Would be nice to have a "play clip @ TC" functionality aswell.

Andy

[...]

PR #1140, opened 2018-12-19
feat: Timecode & scheduler

Julusian:
#871 #872 #1004

This adds support for timecode input and output via decklinks, and for being able to schedule commands to be executed at various timecodes. It also adds a batch command syntax to amcp.

To be able to use a decklink as the source, it must exist on the stage. To allow this from startup, the xml config can now have producers defined.

Further docs: https://docs.google.com/document/d/1xkksxIdNHMO95RFF8wJ3vYRbXCHwKnjkMwVqwAyjv3A/edit?usp=sharing


ronag:
Can you put the AMCP updates into a separate PR?

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -509,6 +509,12 @@ class decklink_producer : public IDeckLinkInputCallback
                 }
             }
 
+            if (new_timecode != core::frame_timecode::empty())
+            {
+                // Account for latency introduced by ffmpeg filter
+                new_timecode = new_timecode - (format_desc_.field_count > 1 ? 3 : 1);
```
ronag: This is not going to be entirely accurate...

ronag: Hm, no that's a bad idea, but I still think you can measure the filter latency using pts

ronag: hmm... maybe this is good enough in practice

ronag: leave it

ronag:
@Julusian I have it on my list to go through this... can you please resolve the conflicts?

didikunz:
Hi, is there a build with that included? It looks to me, that it's still not merged into the main branch, right?

ronag:
@didikunz: Not merged. There are lot's of conflicts.

didikunz:
I would encourage @Julusian or who ever is responsible to resolve this, as this is a power full feature. Is it in the NHK builds?

MauriceVeraart:
Yes

swordh:
Hi Guys,
Love the work, keep it up :)  waiting for this feature to be merged, only reason that I use the NRK version and Windows. 
Are there any plans on getting this merged soon?

poldon:
Is there any plan to merge this still?  This is a very good feature, and one I think a lot of people will benefit from.

GeorgNation (changes requested):
I think it should be in release. The only thing missing is the ability to switch the time zone, timecode with a drop frame, specifying milliseconds and the ability to hold an event. This would make the tool ideal.

Julusian:
I think it is time to close this PR, discussion can continue here or in the issue #871 

The client who originally funded this work no longer has an interest in it, and some more work is needed to flesh out timecode support to make it more usable.
I am not entirely sure on the resulting quality of this change as it was never used in production (other than with the default config of following the system clock). For this to be actually useful to anyone, I expect that support will need to be added to the ffmpeg_producer, and possibly more producers and consumers.

I've had a quick attempt at rebasing this onto current builds, as there were pretty major merge conflicts due to various portions of this having been picked and merged already. If anyone wants to continue working on this you can start with https://github.com/CasparCG/server/commit/056c9469da08b5c958ccf5fc12713e62a94490fa

I have no interest in the scheduler portion of this PR, we found it to be hard to work with and to understand the state of the system when cancelling commands or amending the scheduled commands.

PR #1146, opened 2019-01-02
Small fixes

drakmor:
Fixed errors after static code analysys

### DIFF
--- a/src/modules/bluefish/producer/bluefish_producer.cpp
+++ b/src/modules/bluefish/producer/bluefish_producer.cpp
@@ -348,7 +348,7 @@ struct bluefish_producer : boost::noncopyable
             blue_->render_buffer_capture(BlueBuffer_Image_HANC(schedule_capture_frame_id_));
             dma_ready_captured_frame_id_ = capturing_frame_id_;
             capturing_frame_id_          = schedule_capture_frame_id_;
-            schedule_capture_frame_id_   = (++schedule_capture_frame_id_ % 4);
+            schedule_capture_frame_id_   = (schedule_capture_frame_id_ + 1) % 4;
         }
     }
 
@@ -372,7 +372,7 @@ struct bluefish_producer : boost::noncopyable
                                    image_size);
         double fps = rate;
         if (is_1001)
-            fps = (rate * 1000) / 1001;
+            fps = (static_cast<double>(rate) * 1000) / 1001;
 
         CASPAR_SCOPE_EXIT
         {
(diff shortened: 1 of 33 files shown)


src/modules/bluefish/producer/bluefish_producer.cpp
```diff
@@ -348,7 +348,7 @@ struct bluefish_producer : boost::noncopyable
             blue_->render_buffer_capture(BlueBuffer_Image_HANC(schedule_capture_frame_id_));
             dma_ready_captured_frame_id_ = capturing_frame_id_;
             capturing_frame_id_          = schedule_capture_frame_id_;
-            schedule_capture_frame_id_   = (++schedule_capture_frame_id_ % 4);
+            schedule_capture_frame_id_   = (schedule_capture_frame_id_ + 1) % 4;
```
Julusian: schedule_capture_frame_id_ is no longer being incremented. Was this intentional?

src/modules/decklink/decklink.cpp
```diff
@@ -60,14 +60,21 @@ std::vector<std::wstring> device_list()
...
             String m_name;
+#endif        
             bool   success = SUCCEEDED(decklink->GetModelName(&m_name));
             decklink->Release();
             std::wstring model_name = L"Unknown";
 
-            if (success)
-                model_name = u16(m_name);
-
+            if (success && m_name != nullptr)
```
Julusian: Can this not use the get_model_name function defined in util.h?

src/modules/decklink/util/util.h
```diff
@@ -252,9 +252,17 @@ static com_ptr<IDeckLink> get_device(size_t device_index)
 template <typename T>
 static std::wstring get_model_name(const T& device)
 {
+
```
Julusian: Why was this change needed?
It was much cleaner before without any ifdefs, and worked as far as Im aware.

Julusian:
I'm curious, what did you use to do this static analysis?

src/modules/bluefish/producer/bluefish_producer.cpp
drakmor: schedule_capture_frame_id_  was previously increased, and then overwritten by a new value. The modification of a variable is unsequenced relative to another operation on the same variable. This may lead to undefined behavior.

src/modules/decklink/decklink.cpp
drakmor: Yes we can

src/modules/decklink/util/util.h
drakmor: Because different string types in Windows/Linux/Apple
1. https://forum.blackmagicdesign.com/viewtopic.php?p=423675#p423675
2. https://www.viva64.com/en/w/v745/print
3. Like this - https://github.com/Streampunk/macadam/blob/master/src/macadam.cc

drakmor:
> I'm curious, what did you use to do this static analysis?

PVS Studio

src/modules/decklink/util/util.h
Julusian: Yes, but we don't need to do it via ifdefs in this way.
`String` is a custom typedef in https://github.com/CasparCG/server/blob/master/src/modules/decklink/decklink_api.h that on windows is `BSTR` and on linux is `const char*`

I am assuming there was some static analysis problem here due to BSTR being more than just a byte array and it being treated as such when passed into `u16(`, so how about adding a `std::wstring to_string(String);` for each platform instead in decklink_api.h which this code can use instead of u16?

drakmor: It's best idea.

drakmor:
Repo rebased

PR #1153, opened 2019-01-08
feat: AMCP Command batching

Julusian:
Example usage:

```
>>REQ abc BEGIN
>>PLAY 1-10 GREEN
>>REQ bbb PLAY 1-11 BLUE
>>PLAY
<<xxx PLAY ERROR // parsing errors immediately return an error, but the batch is continued.
>>COMMIT
<<202 PLAY OK
<<RES bbb 202 PLAY OK
<<RES abc 202 COMMIT OK

>>BEGIN
>>PLAY 1-10 GREEN
>>DISCARD
```

When a command is received as part of the batch, it is executed up until it needs to set/get something on the stage. It is then added to the batch for later execution. This ensures that as much work as possible is done before locking the stage(s) to execute the batch to minimise disruption to channel mixing


Julusian:
@ronag do you want this #1153 rebased on top of this or rebased to be separate?

Some of the changes here may not fully make sense without #1153, as some were to make it possible to get the scheduler and command repo accessible from other commands

ronag:
@Julusian I have it on my list to go through this... can you please resolve the conflicts?

PR #1163, opened 2019-01-16
Code Cleaning, Processing with Clang-Tidy.

drakmor:


drakmor:
Done

Julusian:
What command did you use to run clang-tidy?
I have just found that building with clang is now outputting a bunch of warnings that appear to be introduced by this change

drakmor:
I use code cleaning through ReShaper. There is processing through clang-tidy.

Julusian:
Thanks, looks like I need to figure out a shell script for it myself at some point then :)

PR #1167, opened 2019-01-20
Remove boost::noncopyable pattern

drakmor:


drakmor:
`void operator=(const A&) = delete; ` instead of  ` void operator=(const A&) = delete` for prevent `A=B=C;` construction

drakmor:
@ronag return classic `const A& operator=(const A&)` ? Or leave boost :: noncopyable?

ronag:
I don’t get the void return. What’s that about? I don’t see the point when you are deleting the operator.

drakmor:
Updated. Removed "void".

ronag:
LGTM

PR #1168, opened 2019-01-20
boost::lexical_cast -> std::st*

drakmor:


ronag:
Not exactly the same, but good enough

dimitry-ishenko:
IMHO with c++11/14 support dependency on boost can be greatly reduced

PR #1178, opened 2019-02-16
Added support for DataVideo TC-100/TC-200 Character Generators

drakmor:
Related issue #1032.

Test build - https://1drv.ms/u/s!AsCJ7q9Sqbclk4oEa9M6zoJrtAbPIg

Example config:
```
        <channel>
            <consumers>
                <screen>
		   <datavideo-colors>1</datavideo-colors>
              	   <device>2</device>
                   <aspect-ratio>16:9</aspect-ratio>
                   <stretch>none</stretch>
                   <windowed>true</windowed>
                   <vsync>true</vsync>
                   <borderless>true</borderless>
                   <interactive>false</interactive>
                   <always-on-top>true</always-on-top>
                   <x>0</x>
                   <y>0</y>
		</screen> 
            </consumers>
        </channel>
```
fill
key
coded

### DIFF
--- a/src/modules/screen/consumer/screen_consumer.cpp
+++ b/src/modules/screen/consumer/screen_consumer.cpp
@@ -82,6 +82,13 @@ struct configuration
         aspect_invalid,
     };
 
+    enum class colour_spaces
+    {
+        RGB               = 0,
+        datavideo_full    = 1,
+        datavideo_limited = 2
+    };
+
     std::wstring    name          = L"Screen consumer";
     int             screen_index  = 0;
     int             screen_x      = 0;
@@ -97,6 +104,7 @@ struct configuration
     bool            interactive   = true;
     bool            borderless    = false;
     bool            always_on_top = false;
+    colour_spaces   colour_space  = colour_spaces::RGB;
 };
 
 struct frame
@@ -256,6 +264,7 @@ struct screen_consumer
                 shader_ = get_shader();
                 shader_->use();
                 shader_->set("background", 0);
+                shader_->set("window_width", screen_width_);
 
                 for (int n = 0; n < 2; ++n) {
                     screen::frame frame;
@@ -266,8 +275,18 @@ struct screen_consumer
                         reinterpret_cast<char*>(GL2(glMapNamedBufferRange(frame.pbo, 0, format_desc_.size, flags)));
 
                     GL(glCreateTextures(GL_TEXTURE_2D, 1, &frame.tex));
-                    GL(glTextureParameteri(frame.tex, GL_TEXTURE_MIN_FILTER, GL_LINEAR));
-                    GL(glTextureParameteri(frame.tex, GL_TEXTURE_MAG_FILTER, GL_LINEAR));
+                    GL(glTextureParameteri(frame.tex,
+                                           GL_TEXTURE_MIN_FILTER,
+                                           (config_.colour_space == configuration::colour_spaces::datavideo_full ||
+                                            config_.colour_space == configuration::colour_spaces::datavideo_limited)
+                                               ? GL_NEAREST
+                                               : GL_LINEAR));
+                    GL(glTextureParameteri(frame.tex,
+                                           GL_TEXTURE_MAG_FILTER,
+                                           (config_.colour_space == configuration::colour_spaces::datavideo_full ||
+                                            config_.colour_space == configuration::colour_spaces::datavideo_limited)
+                                               ? GL_NEAREST
+                                               : GL_LINEAR));
                     GL(glTextureParameteri(frame.tex, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE));
                     GL(glTextureParameteri(frame.tex, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE));
                     GL(glTextureStorage2D(frame.tex, 1, GL_RGBA8, format_desc_.width, format_desc_.height));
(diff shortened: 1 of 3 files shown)


Julusian:
Thanks for this. I'll have a look through the code soon. 
One thought from the description is that I think it would be better in the config to have a colourspace field (default of rgb) instead of datavideo-colors.

drakmor:
Like this <colour-space>RGB|datavideo-limited|datavideo-full<colour-space>?

Julusian:
Yes, that would be perfect

drakmor:
Done

src/modules/screen/consumer/screen_shader.cpp
```diff
@@ -26,37 +26,81 @@ namespace caspar { namespace screen {
...
+  
+            void main()
+	    {
                 vec4 color = texture(background, TexCoord.st);
-
                 if (key_only)
                     color = vec4(color.aaa, 1);
-
+                
+                else if (colour_space == COLOUR_SPACE_DATAVIDEO_FULL) {  // Full range 0-255
```
Julusian: It would be nice in this code to use color.stpq instead of rgba once color is not rgb to make it clearer. And to use TexCoord.x for the same reason

src/modules/screen/consumer/screen_shader.cpp
```diff
@@ -26,37 +26,81 @@ namespace caspar { namespace screen {
...
+                    color = dtv_color(color);
+                    if (mod(floor(TexCoord.s * window_width),2.0f) == 0.0f) { // even pixel
```
Julusian: I've read that this type of if statements based on pixel values in glsl are really unoptimal due to how the gpu executes the work. The if statements based on shader properties are fine. It would be better for it to have a single flow that does more math instead. https://stackoverflow.com/questions/37827216/does-if-statements-slow-down-my-shader
Also, I am pretty sure that mod operations on floats aren't very accurate (eg that could return 0.0001), so something mathematical or with tolerance would be good for this too.

src/modules/screen/consumer/screen_shader.cpp
```diff
@@ -26,37 +26,81 @@ namespace caspar { namespace screen {
...
+            uniform int colour_space;
+            uniform int window_width;
 
-            void main() {
+            // rgb2yuv_709, rgb=0~255, y=16~235, uv=16~240
+            mat4 RGBA2YCbCrA = mat4(0.183f, -0.101f, 0.439f, 0.0f, 0.614f, -0.338f, -0.399f, 0.0f, 0.062f, 0.439f,-0.040f, 0.0f, 0.0f, 0.0f, 0.0f, 1.0f);
+   
+            vec4 dtv_color(vec4 rgba)
+            {
+              if(rgba.a > 0.0f)
```
Julusian: Same comment about if statements as below. Something like https://github.com/CasparCG/server/blob/1a543c8d953e47b58914de61adffb4c187598599/src/accelerator/ogl/image/image_shader.cpp#L92 would be better and would have no effect on the result

src/modules/screen/consumer/screen_shader.cpp
```diff
@@ -26,37 +26,81 @@ namespace caspar { namespace screen {
...
+                       color.g = clamp(((color.g + next_color.g) * 0.427734375f) + RANGE_HALF + 0.5f, RANGE_16, RANGE_235);
+                    } else { // odd pixel
+                       vec4 prev_color = dtv_color(texture(background, TexCoord.st - vec2(1.0 / window_width, 0.0f)));
+                       color.g = clamp(((color.b + prev_color.b) * 0.427734375f) + RANGE_HALF + 0.5f, RANGE_16, RANGE_235);
+                    }
+                    color.r = clamp((color.r * 0.85546875f) + RANGE_16 + RANGE_HALF, RANGE_16, RANGE_235);
+	            color.b = clamp((color.a * 0.85546875f) + RANGE_16 + RANGE_HALF, RANGE_16, RANGE_235);
+                
+                } else if (colour_space == COLOUR_SPACE_DATAVIDEO_LIMITED) {  // Limited range 16-235
+                    color = dtv_color(color);
```
Julusian: This looks like pretty much the same code as above, so could it be refactored into a common function? 
0.85 would be good to make into a constant like RANGE_HALF to make it clear where it comes from. And the 0.42 would be better as `0.5 * RANGE_LIMITED` to make that clearer too. The glsl compiler should optimise out the extra operation, so clearer code is better.

drakmor:
Screen shader optimized. ~20% perfmormance increased.

drakmor:
Rebased and reformatted.

src/modules/screen/consumer/screen_consumer.cpp
```diff
@@ -107,7 +115,7 @@ struct frame
     GLsync fence = nullptr;
 };
 
-struct screen_consumer
+struct screen_consumer : boost::noncopyable
```
Julusian: Looks like you are accidentally reverting another PR of yours.

drakmor:
Need some fixes. Please not approve.

drakmor:
All done.

Known bug - 3 vertical lines on 1-4 first columns on  light colors and crop right on 4 columns. This is TC-200 hardware error. (On CG-200/CG-350 same error)

didikunz:
Very cool, thank you!

vimlesh1975:
Nice

On Feb 21, 2019 5:38 PM, didikunz <notifications@github.com> wrote:

Very cool, thank you!

PeterAkakpo:
**Added support for DataVideo TC-100/TC-200 Character Generators** 
How does this work, what kind of support is been referred to here

baltedewit:
If I understand this correctly, you can set up a CasparCG screen consumer on an hdmi output, plug in a TC-100 or TC-200 to that hdmi output and it will generate a fill/key pair to be used in a vision mixer.

didikunz:
> If I understand this correctly, you can set up a CasparCG screen consumer on an hdmi output, plug in a TC-100 or TC-200 to that hdmi output and it will generate a fill/key pair to be used in a vision mixer.

That is correct. The trick here is, that video uses YUV color space with a lower rate for the two color signals. That is called 4:2:2 and HDMI uses 4:4:4 so the two colors deference signals can be encodes ito one channel (2+2=4) so one channel is uses to output the key signal- The DataVideo boxes tear this appart and output two SDI signals, one for fill and one for key.

PR #1181, opened 2019-02-19
Fix boost::filesystem:canonical symlink bug (infinity loop) in Windows

drakmor:
Fixing bug in Boost library - https://www.bountysource.com/issues/32827214-boost-filesystem-canonical-enters-infinite-loop-when-current-path-is-a-directory-symlink-on-windows


Julusian:
@drakmor do you think this should be merged to 2.2 also?

drakmor:
It would be nice

PR #1182, opened 2019-02-21
chore: Refactor glsl shaders into separate files

Julusian:
This will make it easier to work with the shader files, as they will not be scattered in a hard to follow fashion around multiple c++ files and editors will be able to give proper linting.

The bin2c program converts any file to a c header containing a const char array, so they can be embedded in the executable


baltedewit:
This one makes me _really_ happy :D

Issue #946, opened 2018-03-21
Memory leak on win 7 x64 with default config


saltomodules:
### Expected Behaviour

Memory usage remains more or less stable all other factors staying the same.

### Current Behaviour

Memory usage keeps going up with about 100 - 300 kb per second even when casparcg isn't really doing anything. 

This begins immediately after starting casparcg with the default configuration.

Behavior is present in latest build 88171e16  but is als observable in earlier builds the earliest  tested being 7ba5e475 

### Steps to Reproduce 
Start casparcg with default config and monitor memory usage

### Environment
* CasparCG Server version: 
2.2.0.4 88171e16 Beta 1  (2018-03-20)
is also observed in 2.2.0.3 7ba5e475 Beta 1 (2018-02-28)
* Operating system:
Win 7 x64
* Graphics driver:
Nvidia 391.01 (for GTX660)
* Decklink drivers:
none

baltedewit:
Judging from the task menu I can reproduce this on Windows 10.

This appears to be coming from the system-audio consumer, removing it stopped the excessive growth of memory. Although it appears there is still some minor growth with just the screen consumer. I don't notice any growth without consumers.

dotarmin:
I can confirm this.

ronag:
I can only reproduce it with system audio.

Jeff-Clark-MST:
Can confirm reproducing this on Window 10.

Removing `<system-audio />` from my config for each channel solves.

E.g. 
Causes Mem Leak:
```
  <channels>
    <channel>
      <video-mode>1080i5000</video-mode>
      <consumers>      
         <decklink>
          <device>1</device>
          <embedded-audio>true</embedded-audio>
	  <keyer>external</keyer>
        </decklink>
        <system-audio />  <!-- causes mem leak -->
      </consumers>
    </channel>
  </channels>
```

No Mem Leak:

```
  <channels>
    <channel>
      <video-mode>1080i5000</video-mode>
      <consumers>      
         <decklink>
          <device>1</device>
          <embedded-audio>true</embedded-audio>
          <keyer>external</keyer>
        </decklink>
      </consumers>
    </channel>
  </channels>
```

Punkley:
confirmed..

Xeing:
confirmed:

Server version:
2.2.0 66a9e3e2 Stable
Operating system:
Win 10 x64
Graphics driver:
Nvidia 417.35 (for GTX1050)
Decklink drivers:
none

PR #1183, opened 2019-02-23
Fixed memory leak in system audio consumer

drakmor:
Related issue #946


drakmor:
Something is wrong? The receiving buffer was allocated, for the original buffer everything is setted manually. It seems like a second allocation is not needed at all. https://github.com/FFmpeg/FFmpeg/blob/976dae8b32f48d17cccfd6b19d2beb01770dfa7c/libavutil/frame.c#L268

drakmor:
@ronag first allocation https://github.com/CasparCG/server/blob/2d57f6283063e2102dc2864f88efa677bf8801fd/src/modules/oal/consumer/oal_consumer.cpp#L179 second allocation https://github.com/CasparCG/server/blob/2d57f6283063e2102dc2864f88efa677bf8801fd/src/modules/oal/consumer/oal_consumer.cpp#L210

PR #1184, opened 2019-02-23
Updated FFMPEG to 4.1.1 (Windows)

drakmor:
Updated to stable FFMPEG release (21.02.2019)


dotarmin:
I'm not sure if this such as good idea to merge this without updating the Linux version as well, input?

baltedewit:
> I'm not sure if this such as good idea to merge this without updating the Linux version as well, input?

agreed, although there are no code changes in this PR so all that should take is to update the docker files with a different ffmpeg from github.

Julusian:
> I'm not sure if this such as good idea to merge this without updating the Linux version as well, input?

Yep, I can do that when merging.

I was thinking it would be nice to merge this after #1048 to make it easier to check it all works and then it becomes a 2 line change to update the linux version

dotarmin:
> I was thinking it would be nice to merge this after #1048 to make it easier to check it all works and then it becomes a 2 line change to update the linux version

Sounds like you have a plan! Good work @Julusian

PR #1185, opened 2019-02-25
Ingoring carret return when shader convertion to C header.

drakmor:
Added bin2c arguments validation and ingoring carret return when shader convertion to C header.
Fixed windows build after #1182


Julusian:
@drakmor what GPU are you using here?
I am finding that it works fine without this on both Intel on Linux and a Quadro on Windows

drakmor:
NVidia Gtx 1080 ti

вт, 5 мар. 2019 г., 21:35 Julian Waller <notifications@github.com>:

Julusian:
Turns out it crashes in Debug but is fine in Release. How strange..

Issue #1145, opened 2019-01-01
2.2.0 stable Sound issues 1080p5994 


BonjwaMaurice:
### Expected behaviour
Same smooth playback as in version 2.1.0

### Current behaviour
Sound problems occur with 1080p5994 (and 1080p2997, 720p5994, 720p2997). It sounds like scratching or something similar. The problem persists with different videos. With other video modes like 1080p6000 or 1080p3000 there are no problems.

----

### Environment
Server version: v2.2 stable

brucedk:
I have the same problem here, just made a post in caspar forum, and make some tests that I share there

https://casparcgforum.org/t/problems-with-audio-on-ntsc/1487

frogcast2:
Here is an example of my case:
https://youtu.be/WmbvabTNj8U

same behaviour with 2997 or NTSC for all resolutions and all video files.

brucedk:
I made some tests with #1192, but for me the problems still remains... I will test in other machines that I have caspar to a more accurate result and post here in a couple days

Issue #838, opened 2018-02-15
Implement new nb_samples API for NTSC support


ronag:
Some of the mixer changes broken NTSC and requires the new nb_samples API to be implemented by the ffmpeg and decklink producers.

TomKaltz:
I can't even get the server to start correctly with an NTSC channel.  Is that related to this issue?

ronag:
Hm, maybe I will have to check. Unfortunately this change was required in order to get channel and layer producer working properly.

5opr4ni:
Hi

If you have tested with the version currently on Goole Drive, it  works with nothing. I have NTSC running with the version before that, ( 2.2.0.3.....2529 Beta 1 x64). It plays my 1080i5000 files ~nicely, but NTSC stutter some.

5opr4ni:
seems weird.

ronag:
@TomKaltz @5opr4ni sorry, I accidentally uploaded incompatible ffmpeg dlls when updating scanner.exe. Should be fixed now.

ronag:
@5opr4ni NTSC should run now. Still has the 1 sample off by frame problem but so far I haven't been able to notice it.

TomKaltz:
Just as a comparison, the distortion in system audio consumer is a LOT worse than the distortion in the files recorded by ffmpeg consumer at 1080i5994.

TomKaltz:
So it looks like support for NTSC framerates will not be included in 2.2 unless someone in the community can help implement audio sample handling in the ffmpeg and decklink producers.  Is there someone that might be able to tackle this issue?  @ronag any guidance on to what were looking for here?

TomKaltz:
Is there any update for this?

TomKaltz:
@Julusian what is the plan here?

Julusian:
I think this issue is specifically referring to the `nb_samples` parameter to `receive` on producers. So at some point this parameter should either be start being used, or it should be removed.

So I would classify this as tech debt, not something that needs to be done with any urgency, as #1615 solves the immediate problem this wanted to solve.  

I do think there is value in the approach of leaving it to producers to provide the correct number of samples with each frame, but I think it will be much harder to get working and cover all edge cases, so I think #1615 is a good enough solution but isnt the best solution.  
I suspect that fixing all the quirks for the route_producer is going to involve implementing this approach for just that producer.

Copying what I wrote elsewhere, for future reference:
[...]

PR #1192, opened 2019-03-14
NTSC Audio Fix

d2productions:
Modified audio mixer to prevent asynchronous audio cadences from causing crackling, and modified video_channel to actually use the cadence instead of the first element of the cadence array.
With regard to issues #838 #1145

### DIFF
--- a/src/core/mixer/audio/audio_mixer.cpp
+++ b/src/core/mixer/audio/audio_mixer.cpp
@@ -36,6 +36,7 @@
 #include <stack>
 #include <vector>
 
+
 namespace caspar { namespace core {
 
 using namespace boost::container;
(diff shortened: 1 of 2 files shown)


src/core/mixer/audio/audio_mixer.cpp
```diff
@@ -100,9 +101,14 @@ struct audio_mixer::impl
...
         for (auto& item : items) {
             auto ptr  = item.samples.data();
-            auto size = std::min(item.samples.size(), result.size());
+            auto size = result.size();
             for (auto n = 0; n < size; ++n) {
-                mixed[n] = static_cast<double>(ptr[n]) * item.transform.volume + mixed[n];
+		if (n<item.samples.size()){
+                    mixed[n] = static_cast<double>(ptr[n]) * item.transform.volume + mixed[n];
+		}else{
+		    auto offset = (item.samples.size()) - (channels-(n%channels));
```
ronag: What does this do?

d2productions: The core issue at hand is that the producers and mixer/consumers don't synchronize their use of the audio cadence. This code duplicates the last sample from each audio channel if the mixer is on a longer cadence value than the producer in question. In the opposite case, where the mixer wants a short cadence and the producer created a longer one, it simply drops the extra sample (or two, depending on the cadence). This method is undetectable in the real world, as duplicating the last value prevents a click or pop (as opposed to a zero value), and it never allows the audio to fall out of sync beyond a couple samples before self-correcting when the inverse case occurs. The max audio sync deviation is +/- ~1/800th of a frame.

PR #1199, opened 2019-04-03
Linux support for the newtek module

Palakis:
Currently, the newtek module is only enabled and built on Windows. This PR makes the newtek module available in the Linux version so that NDI can be used there. The iVGA consumer is disabled on Linux.

To do:
- [X] Test on Linux
- [x] Search for libndi in the OS' standard library paths
- [x] Test on Windows (to avoid breaking iVGA)


src/modules/newtek/util/ndi.cpp
```diff
@@ -81,12 +81,7 @@ NDIlib_v3* load_library()
 
 #else
     // Try to load the library
-    void* hNDILib = dlopen(dll_path.c_str(), RTLD_LOCAL | RTLD_LAZY);
-
-    if (!hNDILib && runtime_dir) {
```
Julusian: It would be good to leave this if block as a fallback.
It is a 'standard' search path for the ndi libs, so we should try it if the system paths dont find anything

Palakis: Agreed, gonna fix this.

gbraad:
Very interested in this ...

PR #1201, opened 2019-04-06
Add UYVY support in NDI Producer

ianshade:
This adds UYVY pixel format support for incoming NDI sources, so the default formats would become:

- UYVY for sources without alpha - it's native for NDI so most of the time no conversion needs to be done by the library (except deinterlacing if needed).  I estimate up to 50% less CPU usage per such source, comparing to previously used RGBX.
- RGBA for sources with alpha

If I get everything right, in order to make UYVY work, frame needs to be rendered in its original resolution to a framebuffer, while being converted to BGRA. Rendering a packed 4:2:2 frame while scaling it, without prior conversion, would produce incorrect results as Cb and Cr components are on the same component in the texture, so filtering would cause them to be mixed. I added a separete shader to perform this conversion, but as I'm thinking now, things it does could be done by the existing image shader, while still rendering to a framebuffer first...
I would appreciate reviews and ideas how to make it better.

I am also working on adding all the other possible pixel formats (YV12, NV12, I420 and UYVA), but the NDI library doesn't support deinterlacing on them, so I'll have to figure out my own.

EDIT:
I made minor changes, yet still the conversion should probably be moved somewhere... I'm thinking about a way to pass textures from/to producers and consumers, as conversion to UYVY/UYVA has to be added somewhere to optimize the NDI Consumer, and it also would be handy in another module I'm working on.


Julusian:
> @ianshade I made minor changes, yet still the conversion should probably be moved somewhere.

Could it be done in the mixing shader? There are already some conversions from planar YCbCr formats. https://github.com/CasparCG/server/blob/master/src/accelerator/ogl/image/shader.frag#L488

ianshade:
Ok, I moved it to the existing shader

walterav1984:
@dotarmin 
Might this be an option for 2.3 LTS branch?

dotarmin:
> Might this be an option for 2.3 LTS branch?

@walterav1984, I would like the community to vote for this one in terms of LTS or if it can wait until next version. To be able to create the first LTS with support from the community members for a certain amount of time and to focus on stability, performance and reliability we need to streamline what goes into v2.3.0 LTS.

Community, thumb up on the PR description for include it and a Rocket for waiting until next version.

@Julusian, @5opr4ni, @ronag, what's your input?

src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -234,9 +234,66 @@ struct image_kernel::impl
...
         }
 
-        // Setup shader
+        // Setup conversion
 
         shader_->use();
+        std::shared_ptr<texture> conversion_texture;
+
+        if (params.pix_desc.format == core::pixel_format::uyvy) {
+            int width  = params.pix_desc.planes.at(0).width;
```
Julusian, line 243: Does this need to be a separate call into the shader?
The normal shader call should be able to handle doing the colour conversion as part of the mix. Which is how the planar yuv that ffmpeg produces is handled currently

ianshade: You can convert anything planar, semi-planar like NV12, or packed 4:4:4(:4) while mixing, but can you do it when components are interleaved like in UYVY? I couldn't find a way to make linear interpolation work correctly with it, hence the decision to first draw the frame in 1:1 scale.

dotarmin:
@ianshade, I have added this pr to v2.3.0 LTS milestone. 

Best regards,
Armin

src/accelerator/ogl/image/image_kernel.cpp
Julusian: Good point, I forgot about that

src/accelerator/ogl/image/shader.frag
```diff
@@ -509,6 +510,23 @@ vec4 get_rgba_color()
         return vec4(get_sample(plane[0], TexCoord.st / TexCoord.q).bgr, 1.0);
     case 9:		//rgb,
         return vec4(get_sample(plane[0], TexCoord.st / TexCoord.q).rgb, 1.0);
+    case 10:	//uyvy,
+        {
+            float y = get_sample(plane[0], TexCoord.st).g;
+            float cb;
+            float cr;
+            if(mod(TexCoord.s*texture_width, 2.0) < 1.0)
```
Julusian, line 518: Is there any way this could be done without an if statement?
From what I remember if statements like this are real performance killers

src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -234,9 +234,66 @@ struct image_kernel::impl
...
+            int height = params.pix_desc.planes.at(0).height;
+
+            conversion_texture = ogl_->create_texture(width, height, 4);
+            conversion_texture->attach();
+
+            shader_->set("plane[0]", texture_id::plane0);
+            shader_->set("plane[1]", texture_id::plane1);
```
ducthiem90, line 250: Don't need plane[1] for uyvy422 format

src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -234,9 +234,66 @@ struct image_kernel::impl
...
+            shader_->set("is_hd", params.pix_desc.planes.at(0).height > 700 ? 1 : 0);
+            shader_->set("pixel_format", params.pix_desc.format);
+            shader_->set("texture_width", width);
```
ducthiem90, line 253: shader_->set("opacity", 1.0); // need set opacity is 1 for convert color

src/accelerator/ogl/image/shader.frag
ducthiem90: @Julusian  Maybe:
`float posx = round(mod(TexCoord.s * texture_width, 2.0));
cb = textureOffset(plane[0], TexCoord.st, ivec2(0 - posx, 0)).r;
cr = textureOffset(plane[0], TexCoord.st, ivec2(1 - posx, 0)).r;`

ducthiem90:
I have something can be processing at here:
https://github.com/ducthiem90/server-2.2/blob/2.3.x/src/accelerator/ogl/image/shader.frag#L515

but the code is right only while video size = channel video format size.
@Julusian could you please get exactly a sample of texture by offset?

ducthiem90:
#1288

src/accelerator/ogl/image/shader.frag
ducthiem90: #1288

ducthiem90:
> > @ianshade I made minor changes, yet still the conversion should probably be moved somewhere.
> 
> Could it be done in the mixing shader? There are already some conversions from planar YCbCr formats. https://github.com/CasparCG/server/blob/master/src/accelerator/ogl/image/shader.frag#L488

@Julusian  I have dealt at #1288 , don't need convertion and clean code

Julusian:
#1288 has been merged as an alternate solution to this

Issue #1194, opened 2019-03-21
Audio is pitched from internet radio stream


TondaKrist:
### Expected behaviour
Audio should not be pitched

### Current behaviour
Audio is pitched

----

### Steps to reproduce
1. PLAY 1-1 http://ice.abradio.cz/blanikfm128.mp3
2. listen to pitched audio on any consumer (declink or system audio)
3. compare to http://ice.abradio.cz/blanikfm128.mp3

### Environment
* Server version: 2.2 and 2.3
* Operating system: Windows 10

premultiply:
#1170

dotarmin:
@premultiply, can this issue be discussed in #1170?

premultiply:
Yes, I think so.
There is no Samplerate conversion or it is not explained that CasparCG (currently) has only support for 48kHz only.

ryci-us:
@premultiply , yesterday i found the same - CasparCG doesn't do sample rate conversion, so only 48kHz plays ok.

dotarmin:
Refering to #1170 for further discussions

Issue #1170, opened 2019-01-22
Version 2.2 audio out of sync


Xeing:
I have found synchronization failures when trying to play several files with versions 2.2 and 2.3 that however in version 2.1 work correctly.

This file is an example:
https://www.dropbox.com/s/uf6d4ld9m1zsbq5/avst.mp4?dl=0

### Environment
Server version:
2.2.0 66a9e3e Stable
Operating system:
Win 10 x64
Graphics driver:
Nvidia 417.35 (for GTX1050)
Decklink drivers:
none

Julusian:
Is this a problem with all files/sources, or only a couple of files you have?

Xeing:
I have tried several videos and it seems that:

1. In all videos the audio and video do not start at the same time and there is a small lag.

2. If the samplerate of the file is 44100, it plays at 48000, so after a few seconds the sound is completely out of phase.

m0vse:
I can confirm that I am seeing the same effect, playing a long (90 minute) MP4 file (25fps with 44.1Khz audio) the audio was massively out of sync almost from the start and got progressively worse. This same video played perfectly in 2,07 and 2.1. I used Handbrake to re-encode the audio to 48K and the sync problems are gone, so this is a suitable workaround for now but something has definitely changed.

m0vse:
This was using a Decklink 4K Extreme with the latest BMD drivers.

WuGGu:
Hi everyone, same problem also with the file of @Xeing. async playback of mp4 files.

CasparCG 2.2.0 Stable
Windows 10
BMD Decklink Duo 
Decklink Drivers: 11.1 newest

ryci-us:
looks like a bug is in https://github.com/CasparCG/server/blob/bfe314ea5d09762962690a0e39038d777717e18c/src/modules/ffmpeg/producer/av_producer.cpp#L226-L230

From ffmpeg-filters manual 7.37 asetrate - Set the sample rate without altering the PCM data. This will result in a change of speed and pitch.

ryci-us:
I have tested with 2.3.0 https://github.com/CasparCG/server/commit/5fadfc858a797a25a1adf502dc8eb23be45372b4 and can confirm that everything is fixed.
This issue can be closed.

premultiply:
https://github.com/CasparCG/server/pull/1215#discussion_r291501178

abmkrishnan:
Is there any workaround without modifying the source code to play correctly?

[...]

PR #1215, opened 2019-06-07
FFmpeg producer allways resample to 48kHz

ryci-us:
Fixes audio out of sync when source is not in 48kHz sample rate.
Fixes: https://github.com/CasparCG/server/issues/1194 https://github.com/CasparCG/server/issues/1170

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -223,7 +223,7 @@ struct Filter
                 filter_spec = "anull";
             }
 
-            filter_spec += (boost::format(",aresample=async=1000:first_pts=%d:min_comp=0.01,asetrate=r=%d,"
+            filter_spec += (boost::format(",aresample=async=1000:first_pts=%d:min_comp=0.01,aresample=%d,"
                                           "asetnsamples=n=1024:p=0") %
                             av_rescale_q(start_time, TIME_BASE_Q, {1, format_desc.audio_sample_rate}) %
                             format_desc.audio_sample_rate)


src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -223,7 +223,7 @@ struct Filter
                 filter_spec = "anull";
             }
 
-            filter_spec += (boost::format(",aresample=async=1000:first_pts=%d:min_comp=0.01,asetrate=r=%d,"
+            filter_spec += (boost::format(",aresample=async=1000:first_pts=%d:min_comp=0.01,aresample=%d,"
```
premultiply: Looks to me as this PR uses resample filter twice now.
Can this be combined into only one filter step?

Maybe
`aresample=osr=48000:async=1000:first_pts=%d:min_comp=0.01`
would be a better solution?

ryci-us: i think no, first filter Stretch/squeeze the data to make it match the timestamps, second one changes sample rate to 48kHz.

ryci-us: as i remember async is not a parameter, it's another filter which computes new sample rate to match sync, and its results are used to adjust aresample output sample rate.

ronag: You can combine

premultiply: async is a parameter of the aresample-filter

ryci-us: yes, you are right. Everyhing works. I will make new PR to fix it.

PR #1222, opened 2019-07-08
Keep increasing frame time until we hit input_duration_ when we play out still frames;

### LINKED ISSUE #1114: LOADBG AUTO not playing
On 2.1 and earlier it is possible to schedule content to play automatically after the foreground item finishes playing by using `LOADBG CONTENTITEM AUTO`

Currently on 2.2 (git master Ubuntu 18.04) content does not start playing automatically.

Output from INFO command below, I notice at least one time counter that seems to have wrapped on layer 10 (which is the misbehaving layer).

```
201 INFO OK
<channel>
   <framerate>50</framerate>
   <framerate>1</framerate>
   <mixer>
      <audio>
         <volume>0</volume>
         <volume>0</volume>
         <volume>0</volume>
         <volume>0</volume>
         <volume>0</volume>
         <volume>0</volume>
         <volume>0</volume>
         <volume>0</volume>
      </audio>
   </mixer>
   <stage>
      <layer>
         <10>
            <background>
               <file>
                  <name>REDACTED/2900.mp4</name>
                  <path>/data/media//REDACTED/2900.mp4</path>
                  <stime>0</stime>
                  <stime>0</stime>
                  <streams>
                     <0>
                        <fps>0</fps>
                        <fps>0</fps>
                     </0>
                     <1>
                        <fps>25</fps>
                        <fps>1</fps>
                     </1>
                  </streams>
                  <time>0</time>
                  <time>0</time>
               </file>
               <loop>false</loop>
               <producer>transition</producer>
               <transition>
                  <frame>0</frame>
                  <frame>0</frame>
                  <type>cut</type>
               </transition>
            </background>
            <foreground>
               <file>
                  <name>REDACTED/1265.mp4</name>
                  <path>/data/media//REDACTED/1265.mp4</path>
                  <stime>3.04</stime>
                  <stime>3.0800000000000001</stime>
                  <streams>
                     <0>
                        <fps>25</fps>
                        <fps>1</fps>
                     </0>
                     <1>
                        <fps>0</fps>
                        <fps>0</fps>
                     </1>
                  </streams>
                  <time>3.04</time>
                  <time>-9223372036851.6992</time>
               </file>
               <loop>false</loop>
               <paused>false</paused>
               <producer>ffmpeg</producer>
            </foreground>
         </10>
         <15>
            <background>
               <producer>empty</producer>
            </background>
            <foreground>
               <file>
                  <path>file:///data/media/REDACTED/redacted.html?artist=MARTIN+GARRIX+FT+BONN&amp;title=High+On+Life</path>
               </file>
               <paused>false</paused>
               <producer>html</producer>
            </foreground>
         </15>
         <20>
            <background>
               <producer>empty</producer>
            </background>
            <foreground>
               <file>
                  <path>/data/media/REDACTED/fullscreen_wh.png</path>
               </file>
               <paused>false</paused>
               <producer>image</producer>
            </foreground>
         </20>
      </layer>
   </stage>
</channel>
```

gizahNL:
fixes #1114  not working with video track shorter then audio track(s) without changing behavior regarding duration selection

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -774,6 +774,11 @@ struct AVProducer::Impl
         if (buffer_.empty() || (frame_flush_ && buffer_.size() < 4)) {
             if (buffer_eof_) {
                 frame_eof_ = true;
+                if (frame_time_ < input_duration_ && frame_duration_ != AV_NOPTS_VALUE) {
+                    frame_time_ += frame_duration_;
+                } else if (frame_time_ < input_duration_) {
+                    frame_time_ = input_duration_;
+                }
                 return core::draw_frame::still(frame_);
             }
             graph_->set_tag(diagnostics::tag_severity::WARNING, "underflow");
@@ -822,10 +827,6 @@ struct AVProducer::Impl
         }
 
         auto frame_time = frame_time_;
-
-        if (frame_eof_ && frame_duration_ != AV_NOPTS_VALUE) {
-            frame_time += frame_duration_;
-        }
 
         return av_rescale_q(frame_time, TIME_BASE_Q, format_tb_);
     }
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -524,7 +524,6 @@ struct AVProducer::Impl
 
     int64_t          frame_count_    = 0;
     bool             frame_flush_    = true;
-    bool             frame_eof_      = false;
     int64_t          frame_time_     = AV_NOPTS_VALUE;
     int64_t          frame_duration_ = AV_NOPTS_VALUE;
     core::draw_frame frame_;
@@ -758,7 +757,6 @@ struct AVProducer::Impl
                 frame_time_     = buffer_[0].pts;
                 frame_duration_ = buffer_[0].duration;
                 frame_flush_    = false;
-                frame_eof_      = false;
             }
         }
 
@@ -772,11 +770,10 @@ struct AVProducer::Impl
         boost::lock_guard<boost::mutex> lock(buffer_mutex_);
 
         if (buffer_.empty() || (frame_flush_ && buffer_.size() < 4)) {
-            if (buffer_eof_) {
-                frame_eof_ = true;
-                if (frame_time_ < input_duration_ && frame_duration_ != AV_NOPTS_VALUE) {
+            if (buffer_eof_ && !frame_flush_) {
+                if (frame_time_ < duration_ && frame_duration_ != AV_NOPTS_VALUE) {
                     frame_time_ += frame_duration_;
-                } else if (frame_time_ < input_duration_) {
+                } else if (frame_time_ < duration_) {
                     frame_time_ = input_duration_;
                 }
                 return core::draw_frame::still(frame_);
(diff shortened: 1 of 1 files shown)


src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -774,6 +774,11 @@ struct AVProducer::Impl
         if (buffer_.empty() || (frame_flush_ && buffer_.size() < 4)) {
```
Julusian: frame_flush_ is true after a seek. How does your change affect accuracy of frame_time_ around a seek?

src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -823,10 +828,6 @@ struct AVProducer::Impl
 
         auto frame_time = frame_time_;
```
Julusian: Minor tidying to inline this variable

src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -774,6 +774,11 @@ struct AVProducer::Impl
         if (buffer_.empty() || (frame_flush_ && buffer_.size() < 4)) {
             if (buffer_eof_) {
                 frame_eof_ = true;
```
Julusian: frame_eof_ appears to never be read now, so it should be removed

src/modules/ffmpeg/producer/av_producer.cpp
gizahNL: I haven't tested the real world effect in that case; Perhaps it's better to expand if condition to exclude in this case?

However I do believe it would be better for my patch to compare to duration_ as duration might be different (less) and we might overflow time a bit more in that case. 

Unless overflowing is wanted behavior (i.e.: we keep increasing frame_time_ with duration as long as we keep playout stills, which is what old (2.1) caspar did.

What do you think?

TondaKrist:
Hi Julusian, will be this "feature" merged in 2.3? Still not triggering the next clip when LOADBG AUTO is set  in some clips, where audio track is a few frames longer...

grahamspr86:
@Julusian I'll be interested in seeing this merged into 2.3 the bug is annoying having to set the length to make sure it will autoplay :)

grahamspr86:
Tried this solution over the weekend seems to solve the issue of loadbg not working when the video and audio tracks are not the same length.

Lange:
I can also confirm this solves the issue, and will be running it in production all of next week.

Julusian:
Can someone provide some clips which show this behaviour being fixed please

Lange:
@jdeng89 do you have any of our files on-hand which we know caused this issue?

jdeng89:
@Julusian I've attached a re-encoded version of the PAL Sample File which breaks in master, but does not break with this change.

[PAL.zip](https://github.com/CasparCG/server/files/4285099/PAL.zip)

grahamspr86:
> Can someone provide some clips which show this behaviour being fixed please

I can dig someone out as well if needed which I had in December which didn't work without this modification.

TondaKrist:
Hi, this occurs in 1080 p/i too. Please see file attached. Seems that video track will never reach its end according to OSC an info. [download here](https://bit.ly/2TkXsPN)

Server 2.1 Beta 2 playing next loaded media OK

Julusian:
Thanks for the clips, they both helped in reproducing the bug and testing this fix.

This is merged in 7483a0a92e274268b8d403c095d74c20f9ddcfd0, with a few additional bits I found from testing whenever `SEEK` or `DURATION` parameters were used in the clip before (not caused by this or exclusive to `AUTO`)
a1367bf137615576bd8e2d2a05ff3525bd87ed6a

TondaKrist:
Awesome, thanks. I will test and report when build will be finished and ready for download in master.

PR #1227, opened 2019-08-07
Fix audio on NTSC system

ducthiem90:
I found problem with audio on ntsc system. And propose a solution to fix it


Julusian:
This needs a thorough review and test, but I just found a painful bug in this area in 2.1NRK (and remembered this PR existed touching the same area).

But a question; as you are always setting the tag to `nullptr`, how does this work to identify the streams?

The basic problem with the tag concept, is that routes do not modify this tag which can cause audio to get really messed up if there are more than one route of one source into a channel.

jesperstarkar:
@ducthiem90 We are looking at merging this now. Can you please do a write-up of the core issue and the fix, to better explain what we should look for and verify?

ducthiem90:
> @ducthiem90 We are looking at merging this now. Can you please do a write-up of the core issue and the fix, to better explain what we should look for and verify?

Sorry. It is not ready to merge. I will check it and find another solution

PR #1228, opened 2019-08-08
Fix decklink synchronization clock for interlace field mode

ducthiem90:
I found a problem in casparcg version 2.2 when use decklink consumer for synchronization clock in interlace field mode

And I propose a solution to fix it

### DIFF
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -461,8 +461,12 @@ struct decklink_consumer : public IDeckLinkVideoOutputCallback
             std::shared_ptr<void>     image_data(scalable_aligned_malloc(format_desc_.size, 64), scalable_aligned_free);
             std::vector<std::int32_t> audio_data;
 
+            auto                           time = std::chrono::high_resolution_clock::now();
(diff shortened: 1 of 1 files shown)


martastain:
Any news regarding this? I am eager to test the fix.

sirfnomi:
I'm also waiting for this fix

ducthiem90:
> Any news regarding this? I am eager to test the fix.

Because Casparcg version 2.2 render interlaced field mode as progressive mode with double rate. Ex: 1080i50 rendered with 50fps, but decklink device locked 25fps. So, 2 field are take consecutively but don't have sleep time, it's wrong. I added a sleep time between 2 field (1/2 fps. Ex: 1080i50 will sleep 20ms)

Julusian:
Thanks for looking at this, this is something I noticed a while ago in 2.2, but haven't figured out the right solution yet.
I'm not sure about this solution, as when Caspar falls behind the decklink (eg from high system load), the decklink forces caspar to run as fast as possible to catch up, and this will end up limiting the speed of that happening. But I haven't had any better ideas, so maybe this will be a good first step.

Julusian:
https://github.com/CasparCG/server/pull/1255 has been merged as an improved version of this

PR #1236, opened 2019-10-09
Fix build in MSVC 2019

xanecs:
I was unable to build without this modification in Visual Studio 2019.
Allow a `const` left hand side for comparing TitleHolders.


Julusian:
Thanks. I have manually merged this into master

PR #1244, opened 2019-11-16
feat(html): add functionality to window.remove() method

krzyc:
For a long time window.remove() call did nothing. This modification allows it to close CEF, empty last frame and change file/path state to empty string.

Probably this is not enough, as it should remove producer at all, but this behavior will need more changes. For now it allows proper monitoring of HTML producer by OSC/AMCP.


dotarmin:
@Julusian @ronag, can you please review this PR when you have time so we can get the fix into the project? :)

krzyc:
@Julusian I have moved state to html_client. Last_frame bit works for me - after removing html_producer is producing empty frame.

krzyc:
I have also changed state behavior - now it empties state instead of setting empty file/path. It will work better when we add more state entries.

src/modules/html/producer/html_producer.cpp
```diff
@@ -524,7 +536,7 @@ class html_producer : public core::frame_producer
 
     std::wstring print() const override { return L"html[" + url_ + L"]"; }
 
-    core::monitor::state state() const override { return state_; }
+    core::monitor::state state() const override { return client_->state(); }
```
Julusian, line 539: You should check client_ for null here, like all the other methods do.
I dont know how it could ever be null, but consistency

src/modules/html/producer/html_producer.cpp
```diff
@@ -361,7 +371,11 @@ class html_client
         auto name = message->GetName().ToString();
 
         if (name == REMOVE_MESSAGE_NAME) {
-            // TODO
+            // TODO fully remove producer
+            this->close();
+            last_frame_ = core::draw_frame::empty();
```
Julusian, line 376: Thinking about this use of `last_frame_` I dont think it is reliable enough. If there are any frames left in `frames_` then your override here will quickly be lost.

Either `frames_` should also be emptied (and risk losing the last frame or two of an animation).
or this could be changed to allow any buffered frames to play before replacing with empty
```
{
    std::lock_guard<std::mutex> lock(frames_mutex_);
    frames_.push(draw_frame::empty());
}
```


Also you are missing a use of `last_frame_mutex_` here.

krzyc: Sounds good.

PR #1255, opened 2020-03-06
Add field delay to decklink consumer in interlace mode to smooth impact on channel

scriptorian:
The mixer runs with progressive frames even when the channel is set to be interlaced (eg 50 fps for a 50i channel). The current code pulls two frames from the channel at output 'frame' rate (eg 25 fps) which has an impact on the processing occurring upstream.
This change updates Decklink consumer to add a delay before pulling the second frame from the mixer. It uses the same approach as #1228 but with some changes which include addressing Julian's concerns expressed in that issue:
1. The second field delay time is reduced proportionally if the tick-time is extended to allow the decklink consumer to catch up
2. There is an extra line in the graph to show the field 2 tick time
3. The video and audio copies are now done per field to smooth the impact on the CPU

### DIFF
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -290,6 +290,23 @@ struct decklink_consumer : public IDeckLinkVideoOutputCallback
 
     std::atomic<bool> abort_request_{false};
 
+    bool doFrame(std::shared_ptr<void>& image_data, std::vector<std::int32_t>& audio_data, bool topField)
+    {
+        core::const_frame frame(pop());
+        if (abort_request_)
+            return false;
+
+        int firstLine = topField ? 0 : 1;
+        for (auto y = firstLine; y < format_desc_.height; y += field_count_) {
+            std::memcpy(reinterpret_cast<char*>(image_data.get()) + (long long)y * format_desc_.width * 4,
+                        frame.image_data(0).data() + (long long)y * format_desc_.width * 4,
+                        (size_t)format_desc_.width * 4);
+        }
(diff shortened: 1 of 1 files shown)


walterav1984:
Thanks for keeping interest on interlace developments.
1. Besides probably HD has this also been tested on SD?
2. Are NTSC SD/HD drop/non drop frame rates also improved/tested?

src/modules/decklink/consumer/decklink_consumer.cpp
```diff
@@ -419,7 +440,13 @@ struct decklink_consumer : public IDeckLinkVideoOutputCallback
         }
 #endif
         try {
-            auto tick_time = tick_timer_.elapsed() * format_desc_.fps / field_count_ * 0.5;
+            auto elapsed = tick_timer_.elapsed();
+            int  fieldTimeMs = static_cast<int>(1000 / format_desc_.fps);
+            std::chrono::high_resolution_clock::time_point f2TimePoint =
```
Julusian: Could you add a comment to explain why this is such a complex timepoint to calculate, and a quick justification (to not limit the speed when the trying to refill the decklink frame buffer)

scriptorian:
> Thanks for keeping interest on interlace developments.
> 
> 1. Besides probably HD has this also been tested on SD?
> 2. Are NTSC SD/HD drop/non drop frame rates also improved/tested?

I have tested this change on SD, 625 and 525, no problems seen. I don't expect any problems with 59.94/60.00 frame rates - I've done a quick test and all seems to be working as expected.

ducthiem90:
Have a pull request to do same here
#1228

PR #1260, opened 2020-03-13
Tighten handling of streams with packets beyond duration

### LINKED ISSUE #874: STOP ON LAST FRAME
I have two video files, one file (ANDREWS_012.mp4) play and returns to the first frame when done playing, the other (FREEFON_003.mp4) stays on the last frame when done playing. is there way to always stay on the last frame when play is done? the above mentioned files are in the attached file.
[fillers.zip](https://github.com/CasparCG/server/files/1750745/fillers.zip)

scriptorian:
Fixes #874.
The provided files are unusual in that they have a video stream that repeats. These repeated video packets are beyond the reported duration of the file, meaning that to play correctly the producer must be limited by the duration rather than an eof flag from the source.

This change tightens up the time checking to make sure that the producer output stops on the last available frame before the end time calculated from the reported duration. The producer stops two frames earlier than with the current code.

This change will also affect LOADBG and similar when setting a LENGTH parameter.

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -653,12 +653,14 @@ struct AVProducer::Impl
                 auto start    = start_.load();
                 auto duration = duration_.load();
 
-                start     = start != AV_NOPTS_VALUE ? start : 0;
-                auto end  = duration != AV_NOPTS_VALUE ? start + duration : INT64_MAX;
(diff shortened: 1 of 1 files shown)


Julusian:
I have done a quick test to compare the last frame in #874 and other scenarios and am happy with the result

Julusian:
I have reverted this now, as it appeared to cause more problems than it solved

PR #1262, opened 2020-03-17
Fixes #1239. Fix audio filter spec for resample when sample rates are different

### LINKED ISSUE #1239: Audio out of sync for 44100Hz in v2.2.0 and above with SEEK command
### Current behaviour
When playing a video file (44100Hz) in v2.2.0 the audio is going out of sync only when I used the SEEK command otherwise it is playing normally.  But if I convert that video file to 48000Hz then it is playing normally in v2.2.0 even with the SEEK command. But in v2.1.0 it is playing normally without converting to 48000Hz. Is there any workaround to play the file without converting to 48000Hz?

----

### Steps to reproduce
1. PLAY 1-1 VIDEO_FILE SEEK 1000 LENGTH 5000

### Environment
* Server version: v.2.2.0 and above

---

scriptorian:
The audio resample filter takes a parameter for the first pts. This was being calculated with the wrong timebase leading to the outputs being out of sync.

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -223,9 +223,18 @@ struct Filter
                 filter_spec = "anull";
             }
 
+            // Find first audio stream to get a time_base for the first_pts calculation
+            AVRational tb = {1, format_desc.audio_sample_rate};
+            for (auto n = 0U; n < input->nb_streams; ++n) {
+                const auto st = input->streams[n];
+                if (st->codecpar->codec_type == AVMEDIA_TYPE_AUDIO && st->codecpar->channels > 0) {
+                    tb = st->time_base;
+                    break;
+                }
+            }
             filter_spec += (boost::format(",aresample=async=1000:first_pts=%d:min_comp=0.01:osr=%d,"
                                           "asetnsamples=n=1024:p=0") %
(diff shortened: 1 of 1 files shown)


ronag:
first_pts is in output sample rate, not input sample rate. I think the previous code was correct. This PR breaks some files.

ronag:
See:
```cpp
    if (s->firstpts_in_samples != AV_NOPTS_VALUE) {
        if (!s->async && s->min_compensation >= FLT_MAX/2)
            s->async = 1;
        s->firstpts =
        s->outpts   = s->firstpts_in_samples * s->out_sample_rate;
    } else
        s->firstpts = AV_NOPTS_VALUE;

```

ronag:
This is reverted for now. @scriptorian can you provide a file to reproduce the issue you are trying to fix here?

scriptorian:
Hi Robert,
Thanks for reviewing this. The file I used was one I had in my media folder after looking at #874 - ANDREWS_012.mp4 - it is downloadable from there. It's a weird file with other problems but the audio blips when the graphic adds the ticks near the end showed the problem clearly - they became muddy and misaligned when using the existing code and returned to normal with the code I changed.
I can't remember the detail now but I will have looked closely at the pts values and something was definitely off with the existing code.

ronag:
> I can't remember the detail now but I will have looked closely at the pts values and something was definitely off with the existing code.

If you have time to take another look I would be willing to help looking into this.

Issue #1248, opened 2020-01-05
Route Producer introduces delay


ianshade:
### Expected behaviour
There should be 0 frames of delay when routing layers between channels and within a channel, and there should be 0 frames (or maximally 1 frame in specific cases) of delay when routing channels to other channels.

### Current behaviour
The Route Producer currently introduces delay even when routing layers on the same channel. Routing a layer to a layer on the same channel adds 1 frame of delay. Routing a layer to another channel and routing that channel back adds 3 frames of delay. 

### Solution suggestion
Routing a layer seems obvious: a frame comes from a producer on layer n; it is uploaded to the GPU and is available to the mixer when drawing layer n, so it is also available to be drawn on layer m where we routed layer n.
When routing a layer to another channel it also should work fine assuming that drawing of corresponding frames of all channels happens after producers' frames for all layers on all channels have been uploaded.
When routing a channel to another channel, corresponding frames of those channels should be drawn in order that allows drawing one channel on another. If we routed channel m to channel n then channel m should be drawn first and that texture should be used when drawing channel n. 
It becomes more complex with circular dependencies and that might require some buffering hence the maximal 1 frame of delay, but I don't see many cases when one would like to route channel n to channel m and channel m back to channel n at the same time.

----

### Steps to reproduce

1. Setup two channels with the same framerate
2. `MIXER 1 GRID 2` (for convenient display)
3. `PLAY 1-1 SOMEVIDEO`
4. `PLAY 1-2 route://1-1` (routing layer 1 to the same channel)
5. `PLAY 2-1 route://1-1` (routing layer 1 to channel 2)
6. `PLAY 1-3 route://2` (routing channel 2 back to channel 1)

### Environment
* Commit: 4a8f03f
* Server version: 2.3
* Operating system: Windows 10

---

### Screenshots
https://imgur.com/a/Xznms7p

PeterC89:
Duplicate of #815 (ish)

Julusian:
This is not a simple thing to resolve in the current architecture.
Each channel runs in parallel on its own clock, so there is no guarantee of the source channel producing the frames before the route tries to consume it

premultiply:
Maybe a command to instantaneous MOVE the content of layer or channel to somewhere else would help to avoid the buffering between async channels for some cases.
At the end this would work like a preset and program bus on a vision mixer.

Julusian:
[...]

Issue #1226, opened 2019-07-24
Audio Glitching Server 2.2/2.3 - late frames route


grahamspr86:
I'm seeing issues with audio when the CasparCG channel is set to 1080i50 and playing back videos that are 1080p25. Not all the time or on the same videos the audio will not play back correctly sounds kinda like it's skipping.

Setting the CasparCG channel to 1080p25 seems to resolve the issue.

Videos are played on channel 2 then routed to channel 1 which is played out on a DeckLink HD Extreme 3D

Anyone else experienced this?

ducthiem90:
> I'm seeing issues with audio when the CasparCG channel is set to 1080i50 and playing back videos that are 1080p25. Not all the time or on the same videos the audio will not play back correctly sounds kinda like it's skipping.
> 
> Setting the CasparCG channel to 1080p25 seems to resolve the issue.
> 
> Videos are played on channel 2 then routed to channel 1 which is played out on a DeckLink HD Extreme 3D
> 
> Anyone else experienced this?

Do you test with system audio consumer?

grahamspr86:
Afraid I didn't think of giving that a try.

The issue wasn't complete gone when using 1080p25 it would still happen just very rarely.

When I get a bit of time I'll give it a try using the system audio consumer.

grahamspr86:
This seems to be getting caused by route `PLAY 1-10 route://2-10`

Is causing dropped and late frames sitting doing nothing, if I play the video on channel 1 it plays fine.

I've tried versions between d442a8f84f68838f91c82f5f57c9634e84f7a23a and the release of 2.2 and they all have the same issue, switch back to 2.0 and the issue is gone.

![Diag](https://user-images.githubusercontent.com/37407811/62962201-50feef80-bdf6-11e9-8b11-b3ad5f85626e.PNG)

ducthiem90:
> PLAY 1-10 route://2-10

You can try: PLAY 1-10 route://2-10 BUFFER 2

grahamspr86:
> > PLAY 1-10 route://2-10
> 
> You can try: PLAY 1-10 route://2-10 BUFFER 2

That does seem to resolve it!

saltomodules:
> > PLAY 1-10 route://2-10
> 
> You can try: PLAY 1-10 route://2-10 BUFFER 2

Is the `BUFFER` option also present in 2.1? I don't see any references to it in the code on that branch. 

[...]

PR #1267, opened 2020-03-31
feat: zero delay routes

scriptorian:
Adds logic to the stage processing to re-order the pulls from producers so that sources get pulled before the routes they support.
This provides zero delay routes within a channel. When a route is from a different channel the delay will be 1 frame.

### DIFF
--- a/src/core/producer/stage.cpp
+++ b/src/core/producer/stage.cpp
@@ -86,13 +86,14 @@ struct stage::impl : public std::enable_shared_from_this<impl>
     {
     }
 
-    std::map<int, layer_frame> operator()(const video_format_desc&                   format_desc,
-                                          int                                        nb_samples,
-                                          std::vector<int>&                          fetch_background,
-                                          std::function<void(int, core::draw_frame)> routesCb)
+    std::vector<draw_frame> operator()(const video_format_desc&                     format_desc,
+                                       int                                          nb_samples,
+                                       std::vector<int>&                            fetch_background,
+                                       std::function<void(int, const layer_frame&)> routesCb)
     {
         return executor_.invoke([=] {
             std::map<int, layer_frame> frames;
+            std::vector<draw_frame>    stage_frames;
 
             try {
                 for (auto& t : tweens_)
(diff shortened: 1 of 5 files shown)


src/core/producer/stage.h
```diff
@@ -58,8 +58,10 @@ class stage final
 
     explicit stage(int channel_index, spl::shared_ptr<caspar::diagnostics::graph> graph);
 
-    std::map<int, layer_frame>
-    operator()(const video_format_desc& format_desc, int nb_samples, std::vector<int>& fetch_background);
+    std::map<int, layer_frame> operator()(const video_format_desc& format_desc,
```
Julusian, line 61: the return type of this should be changed to use draw_frame instead of layer_frame.
It was changed to layer_frame for routes to use, so with that inside the stage this complexity can be removed

src/core/video_channel.cpp
```diff
@@ -78,6 +78,19 @@ struct video_channel::impl final
     std::atomic<bool> abort_request_{false};
     std::thread       thread_;
 
+    std::function<void(int, core::draw_frame)> routesCb = [&](int layer, core::draw_frame frame) {
+        std::lock_guard<std::mutex> lock(routes_mutex_);
+        for (auto& r : routes_) {
```
Julusian, line 83: Im not keen on how this will iterate through all of routes_ once for every layer

src/core/producer/stage.cpp
```diff
@@ -69,19 +98,60 @@ struct stage::impl : public std::enable_shared_from_this<impl>
...
+                // build a map of layers that are sourced from route producers
+                std::map<int, std::pair<int, int>> routed_layers;
                 for (auto& p : layers_) {
-                    auto& layer = p.second;
-                    auto& tween = tweens_[p.first];
+                    auto producer = std::move(p.second.foreground());
+                    if (0 == producer->name().compare(L"route")) {
+                        std::wstring         routeSrc(producer->print());
+                        boost::wsmatch       what;
+                        static boost::wregex expr(L"route\\[(?<CHANNEL>\\d+)(\\/(?<LAYER>\\d+))?\\]",
```
Julusian, line 108: Could we do something with extending/changing the types instead of relying on regex?
Perhaps like the dynamic_pointer_cast done a few lines below. Or will that have a significant performance impact from all the failed casts?

src/core/producer/stage.h
scriptorian: Agreed, done. layer_frame is now only used for the route callback.

src/core/video_channel.cpp
scriptorian: As discussed, the change to iterating routes for each layer is required to achieve the zero-delay target. I have optimised the loop and with the likely number of routes per channel this should not be a problem.

src/core/producer/stage.cpp
scriptorian: Agreed, done.

dotarmin:
> If you think this is mergeable I won't object but I would prefer it landed post 2.3.

@Julusian, what's your input?

Julusian:
This should resolve a problem being faced at SVT #1226 and would like #1248, so while this may want some work I am going to merge it for 2.3 and we should revisit it with a better solution later

PR #1288, opened 2020-04-23
Support UYVY pixel format for DeckLink and NDI producer

ducthiem90:
Another way to do #1201


ianshade:
Doesn't this upload the same data to the GPU twice?

ducthiem90:
> Doesn't this upload the same data to the GPU twice?

Yes, but I think it does not problem

Julusian:
> > Doesn't this upload the same data to the GPU twice?
>
> Yes, but I think it does not problem

I agree it would be better to avoid doing this, , but I dont think that is a blocking issue to this being merged

Julusian:
Manually merged in 7b910c5cda7d9041105343695e391ab85344f3c9

PR #1302, opened 2020-05-25
Add Support for Kronos K8 card type, and bug fixes

jamesdwise:
This Update adds:
- support for the Kronos K8 card type which would previously identify as "unknown"
- improved handling of the hardware watchdog mechanism in 4k and UHD modes
- improved routing for users of a 3-out Neutron card type.


dotarmin:
@jamesdwise, thanks for you contributions. The Technical Steering Committee decided to include this in the upcoming release at the end of Q2 this year.

PR #1306, opened 2020-05-29
Add logic to detect circular route setup and skip the recursion

### LINKED ISSUE #1295: Server crash on zero route delay
I tested new route producer (zero route delay). It's good. But there was a user error that caused the server to crash:
PLAY 1-2 ROUTE://1-1
PLAY 1-1 ROUTE://1-2

I think we should handle this bug to avoid the problem by mistake

scriptorian:
Fix #1295 
 - Prevents the server from crashing
 - Makes sure new calls to these layers work as expected

Creating a circular route leads to frozen video and looped audio (which will likely buzz).
This fix treats this state as a possible intermediate-state that can occur, that the user must be able to proceed from, without adding logic to the protocol level.

### DIFF
--- a/src/core/producer/stage.cpp
+++ b/src/core/producer/stage.cpp
@@ -50,12 +50,16 @@ struct stage::impl : public std::enable_shared_from_this<impl>
     monitor::state                      state_;
     std::map<int, layer>                layers_;
     std::map<int, tweened_transform>    tweens_;
+    std::set<int>                       routeSources;
 
     executor executor_{L"stage " + std::to_wstring(channel_index_)};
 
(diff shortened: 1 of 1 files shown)


Julusian:
This diag shows an interesting effect of the produce-time climbing steadily with a circular route, until it then crashes.
image

I think the cause is an ever growing depth of draw_frame. As each frame will be wrapping the last in a new draw_frame wrapping. And I suspect the crash could be either when the call stack depth is reached.

But this resolves the use case of this state being a brief intermediate-state, so I shall merge this as it is an improvement

PR #1315, opened 2020-06-16
Startup producers in config file

Julusian:
A small portion taken from #1140, as it is useful on its own and quite well isolated.

This allows for defining producers to be created at startup in the config file. This can be useful as a holding state, or for channels which never change.

```
<channel>
  ...
  <producers>
    <producer id="0">AMB LOOP</producer>
    <producer id="10">DECKLINK 5 FORMAT 1080i5000</producer>
    <producer id="20">[HTML] http://localhost:3000/overlay.html</producer>
  </producers>
</channel>
```

Additionally, some refactoring was done to make the amcp types a bit cleaner and easier to work with 

Note: Not very tested yet, I will un-draft this once it has been tested more
Test build: https://builds.julusian.dev/casparcg/casparcg-server-2.3.x-plus-startup-producers.zip

### DIFF
--- a/src/protocol/CMakeLists.txt
+++ b/src/protocol/CMakeLists.txt
@@ -25,6 +25,7 @@ set(SOURCES
 		util/lock_container.cpp
 		util/strategy_adapters.cpp
 		util/http_request.cpp
+		util/tokenize.cpp
 
 		StdAfx.cpp
 )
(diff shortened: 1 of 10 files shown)


jesperstarkar:
Running in production at TV 2 Denmark, working well with a decklink producer at least.

sirfnomi:
Hi, can we also set mixer for specific layer in startup ?

PR #1316, opened 2020-06-16
Fix #1314 Don't receive from layer when a circular route is detected

### LINKED ISSUE #1314: Circular route has creeping frame-time before crashing
### Expected behaviour
A circular route should not cause a crash, but should either freeze or show black.

### Current behaviour
This diag shows an interesting effect of the produce-time climbing steadily with a circular route, until it then crashes.
image

I think the cause is an ever growing depth of draw_frame. As each frame will be wrapping the last in a new draw_frame wrapping. And I suspect the crash could be when the call stack depth is reached.

### Environment
* Commit: [e.g. ab1234c]
* Server version: 2.3-dev

scriptorian:


Julusian:
@scriptorian This looks like it will blank out every layer being played on the channel. Could we try and blank only the layers involved in the circular route?

scriptorian:
Good spot, thanks. Updated.

PR #1317, opened 2020-06-16
Fix #1281 Change decoding number of threads default value

### LINKED ISSUE #1281: Remove thread limit when using ffmpeg
I have a 4K video with hevc codec. I played it with caspar and obs studio. I ascertain that caspar don't use cpu best for decode, while obs studio is good. So caspar received late frame when obs studio play it smooth. I debuged and here is result:
- Caspar decode video time: 0.03s - 0.1s / frame
- OBS Studio decode video time: 0.02s - <0.04s / frame.
- Both caspar and obs studio use ffmpeg version 4.2.2.
That is video for test: https://drive.google.com/open?id=1n0-DuFJo7qa-x1Pmn6y0QjFxGRXIunnG

scriptorian:
FFmpeg decoding was set to use 4 threads by default. This change checks the codec capabilities and sets the default number of threads to auto if supported or the detected number of cores if codec threading is supported. Configuration parameter _ffmpeg/producer/threads_ will override this value.

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -104,8 +104,12 @@ struct Decoder
 
         FF(av_opt_set_int(ctx.get(), "refcounted_frames", 1, 0));
 
-        // TODO (fix): Remove limit.
-        FF(av_opt_set_int(ctx.get(), "threads", env::properties().get(L"configuration.ffmpeg.producer.threads", 4), 0));
+        int defaultNumThreads =
+            codec->capabilities & AV_CODEC_CAP_AUTO_THREADS
+                ? 0 : codec->capabilities & (AV_CODEC_CAP_FRAME_THREADS | AV_CODEC_CAP_SLICE_THREADS)
+                        ? std::thread::hardware_concurrency()
+                        : 1;
+        FF(av_opt_set_int(ctx.get(), "threads", env::properties().get(L"configuration.ffmpeg.producer.threads", defaultNumThreads), 0));
         // FF(av_opt_set_int(ctx.get(), "enable_er", 1, 0));
 
         ctx->pkt_timebase = stream->time_base;
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -106,8 +106,11 @@ struct Decoder
 
         int defaultNumThreads =
             codec->capabilities & AV_CODEC_CAP_AUTO_THREADS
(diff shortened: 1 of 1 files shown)


src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -104,8 +104,12 @@ struct Decoder
...
         FF(av_opt_set_int(ctx.get(), "refcounted_frames", 1, 0));
 
-        // TODO (fix): Remove limit.
-        FF(av_opt_set_int(ctx.get(), "threads", env::properties().get(L"configuration.ffmpeg.producer.threads", 4), 0));
+        int defaultNumThreads =
+            codec->capabilities & AV_CODEC_CAP_AUTO_THREADS
+                ? 0 : codec->capabilities & (AV_CODEC_CAP_FRAME_THREADS | AV_CODEC_CAP_SLICE_THREADS)
+                        ? std::thread::hardware_concurrency()
+                        : 1;
+        FF(av_opt_set_int(ctx.get(), "threads", env::properties().get(L"configuration.ffmpeg.producer.threads", defaultNumThreads), 0));
```
ronag, line 112: I think it's better to just simply set it to "auto".

scriptorian: Thanks Robert.
I was just writing up some notes about this in the issue: https://github.com/CasparCG/server/issues/1281#issuecomment-644859100

ronag: Sounds good!

src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -104,8 +104,12 @@ struct Decoder
 
         FF(av_opt_set_int(ctx.get(), "refcounted_frames", 1, 0));
 
-        // TODO (fix): Remove limit.
-        FF(av_opt_set_int(ctx.get(), "threads", env::properties().get(L"configuration.ffmpeg.producer.threads", 4), 0));
+        int defaultNumThreads =
+            codec->capabilities & AV_CODEC_CAP_AUTO_THREADS
+                ? 0 : codec->capabilities & (AV_CODEC_CAP_FRAME_THREADS | AV_CODEC_CAP_SLICE_THREADS)
```
ronag, line 109: With slice threads it makes sense to limit it to e.g 8 threads... since it doesn't scale very well.

scriptorian: So you are proposing min<int>(8, concurrency/2) when slice threads, concurrency/2 otherwise?

scriptorian: Done!

Issue #1165, opened 2019-01-19
Inconsistent channel tick rate for interlaced decklink


Julusian:
### Expected behaviour
When the channel has no producers and a decklink consumer, the diag graph should have a consume-time line at roughly 50%

### Current behaviour
Fine when the channel is progressive, but when interlaced the consume-time line is at 0 with an occasional spike. (In debug builds, there are a lot more spikes)
If the decklink consumer is removed, then the line returns to normal.

Other consumer types are also showing massively varying tick-times, and their playback (especially noticable on screen consumer) is stuttery.
The decklink output looks fine, I assume because of the buffering done on the card.

Playback of files with ffmpeg looks to be fine, but the html producer is really unhappy with this if the template uses requestAnimationFrame. Templates doing dom changes using setTimeout run smoother (but not perfect).

I believe this is because when a frame is scheduled to the decklink, the consumer takes 2 frames from the buffer, and the channel then works as fast as possible to refill the buffer (as the buffer is what controls the rate of the channel). 

### Environment
* Commit: 713723609ccc0f2956170c2bd99cf11c053bde26
* Server version: 2.3-dev & 2.2.0
* Operating system: Windows 10

---

### Screenshots
Normal:
![image](https://user-images.githubusercontent.com/1327476/51433232-2a048480-1c3e-11e9-9e70-a16ad89744f3.png)
Debug: 
![image](https://user-images.githubusercontent.com/1327476/51432829-308efe00-1c36-11e9-86c7-46dc228979bf.png)
Release: 
![image](https://user-images.githubusercontent.com/1327476/51433204-8c10ba00-1c3d-11e9-9278-6cca60f0edcf.png)
Release timings: 
![image](https://user-images.githubusercontent.com/1327476/51433212-9e8af380-1c3d-11e9-9fcc-956dcfedd311.png)

Julusian:
I think this is the cause of #1150 and maybe #1103.

@ronag I'm not sure how best to resolve this. Perhaps using prec_timer or similar in decklink_consumer::send() would be reliable enough?

premultiply:
> I believe this is because when a frame is scheduled to the decklink, the consumer takes 2 frames from the buffer, and the channel then works as fast as possible to refill the buffer (as the buffer is what controls the rate of the channel).

Yes, sounds reasonable. Might be caused by the interlacer consuming two progressive frames at the same time from the internal pipeline to create an interlaced frames with two fields.

brucedk:
[...]

PR #1322, opened 2020-06-28
fix(decklink_consumer): frame scheduling

ronag:


ronag:
@dotarmin Can you please try and put back the 2000 card in a machine and try to reproduce the issue with current master, and then try it with this PR and see if there is any difference.

dotarmin:
Yes I can do that tomorrow @ronag 👊

Julusian:
From a glance this looks to revert https://github.com/CasparCG/server/pull/1255 so this will reintroduce timing issues for interlaced channels instead

ronag:
> From a glance this looks to revert #1255 

Yep. I'm sorry I missed that PR.

> reintroduce timing issues for interlaced channels instead

What exactly was the issues?

Julusian:
> What exactly was the issues? "which has an impact on the processing occurring upstream." is a bit vague.

The original issue: https://github.com/CasparCG/server/issues/1165

Additionally I think the route producer was affected too, as it was trying to maintain a tiny frame buffer, which meant requesting 2 frames in short succession would often result in the buffer running out.

ronag:
I don't like the current solution but I don't have a better suggestion at the moment.

Issue #1352, opened 2020-11-13
CasparCG Server - Flash Content File Loading Problem 


jeclark-adobe:
Background:
 
Adobe is retiring Flash Player at the end of 2020. To facilitate the ability for enterprise users (including broadcasters) to continue to use Flash Player for critical legacy applications, we introduced a set of "Enterprise Enablement" features in recent Flash Player versions, which allow administrators to specify what content Flash Player is allowed to load. 
 
Once Flash Player is retired and security updates are no longer available, Flash Player will only load content explicitly allowed by system administrators.  The goal is to minimize the risk that average end-users will encounter malware targeting unpatched instances of Flash Player, while allowing system administrators to continue to use Flash for legacy application.  This approach creates a secure-by-default state, and the requirement to explicitly allow required content minimizes the attack surface presented to client endpoints by limiting or elimintating exposure to untrusted content, depending on the administrator’s choices.

You can read more about those features and related config file directives on page 28 of the Flash Player System Administrators guide, here:
https://www.adobe.com/content/dam/acom/en/devnet/flashplayer/articles/flash_player_admin_guide/pdf/latest/flash_player_32_0_admin_guide.pdf

Problem: 

We received a number of inquiries from the CapsparCG community indicating that our AllowListUrlPattern feature does not work for Flash content loaded by CasparCG Server.

On closer inspection, we observed that CasparCG Server is not sending file URIs that conform to RFC 8089.
https://tools.ietf.org/html/rfc8089

Historically, many of the early security issues in Flash Player were related to ambiguity around URI resolution and normalization issues. We believe that attempting to normalize these paths to conform to the URI standard from Flash Player may lead to unexpected side-effects (including security issues).  Because we are days away from the last planned Flash Player release to ever ship from Adobe, the risk of injecting a bug that could persist indefinitely in the web ecosystem is too high to consider.  It's much safer and more reasonable to require that file URIs be normalized at the client application.  In the supported browser plug-in use-case, the browser does this normalization automatically.

In the case of CaparCG Server, we see a number of non-conformant URIs in our cursory testing which, when passed to Flash Player, result in a failure to match any of the possible AllowListUrlPattern options.

Examples:
C:\Users\labuser\Desktop\CasparCG Server 2.0.7\CasparCG Server\Server\templates\\cg20.fth.pal

[...]

PR #1353, opened 2020-11-17
Cause flash files to be loaded by url

silid:
This will fix https://github.com/CasparCG/server/issues/1352 

Part of the flash eol workflow allows the loading of white listed urls.
System paths cannot be white listed so this PR causes the system path to be converted to a URL.

This should ideally be retrospectively fitted to 2.3, 2.0 and other versions in active use.

### DIFF
--- a/src/modules/flash/producer/flash_producer.cpp
+++ b/src/modules/flash/producer/flash_producer.cpp
@@ -580,12 +580,8 @@ spl::shared_ptr<core::frame_producer> create_producer(const core::frame_producer
     if (!boost::filesystem::exists(filename))
         CASPAR_THROW_EXCEPTION(file_not_found() << msg_info(L"Could not open flash movie " + filename));
 
-    CASPAR_LOG(info) << L"Loading template host: " + url_from_path(filename);
-    return create_destroy_proxy(spl::make_shared<flash_producer>(dependencies.frame_factory,
-                                                                 dependencies.format_desc,
-                                                                 url_from_path(filename),
-                                                                 template_host.width,
-                                                                 template_host.height));
+    const url = url_from_path(filename);
+    return create_destroy_proxy(spl::make_shared<flash_producer>(dependencies.frame_factory, dependencies.format_desc, url, template_host.width, template_host.height));
 }
 
 spl::shared_ptr<core::frame_producer> create_swf_producer(const core::frame_producer_dependencies& dependencies,
@@ -598,12 +594,9 @@ spl::shared_ptr<core::frame_producer> create_swf_producer(const core::frame_prod
 
     swf_t::header_t header(filename);
 
-    CASPAR_LOG(info) << L"Loading flash media: " + url_from_path(filename);
-    auto producer = spl::make_shared<flash_producer>(dependencies.frame_factory,
-                                                     dependencies.format_desc,
-                                                     url_from_path(filename),
-                                                     header.frame_width,
-                                                     header.frame_height);
+    const url = url_from_path(filename);
+    auto producer = spl::make_shared<flash_producer>(
+            dependencies.frame_factory, dependencies.format_desc, url, header.frame_width, header.frame_height);
 
     producer->call({L"start_rendering"}).get();
 
--- a/src/modules/flash/producer/flash_producer.cpp
+++ b/src/modules/flash/producer/flash_producer.cpp
@@ -21,6 +21,10 @@
 
 #include "../stdafx.h"
 
+#include "WinInet.h"
+#include "shlwapi.h"
+#include "winerror.h"
+
 #if defined(_MSC_VER)
 #pragma warning(disable : 4146)
 #pragma warning(disable : 4244)
@@ -151,6 +155,22 @@ std::mutex& get_global_init_destruct_mutex()
     static std::mutex m;
 
     return m;
+}
+
+std::wstring url_from_path(std::wstring in)
+{
+    DWORD        out_length = INTERNET_MAX_URL_LENGTH * 2 + 4;
+    PWSTR        out_buf = (PWSTR)malloc(out_length);
+    HRESULT      ret     = UrlCreateFromPathW(in.c_str(), out_buf, &out_length, NULL);
+    std::wstring out;
+    if (SUCCEEDED(ret)) {
+        out = out_buf;
+        free(out_buf);
+        return out;
+    } else {
+        free(out_buf);
+        return in;
+    }
 }
 
 class flash_renderer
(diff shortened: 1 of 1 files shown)


didikunz:
Please can this be merged? We need that asap, as people needs to be able to update their systems before the holiday season. And already started to panic.

ronag:
This will break Linux builds. Can you make it platform independent?

Julusian:
Linux shouldn't be affected as flash isn't supported or built there https://github.com/CasparCG/server/blob/master/src/modules/CMakeLists.txt#L13

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -153,6 +157,22 @@ std::mutex& get_global_init_destruct_mutex()
...
 
+std::wstring url_from_path(std::wstring in)
+{
+    DWORD        out_length = INTERNET_MAX_URL_LENGTH * 2 + 4;
+    PWSTR        out_buf = (PWSTR)malloc(out_length);
+    HRESULT      ret     = UrlCreateFromPathW(in.c_str(), out_buf, &out_length, NULL);
+    std::wstring out;
+    if (SUCCEEDED(ret)) {
+        out = out_buf;
+        free(out_buf);
```
ronag, line 168: This looks like use after free.

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -555,27 +575,35 @@ spl::shared_ptr<core::frame_producer> create_producer(const core::frame_producer
...
+                                                                 url_from_path(filename),
+                                                                 template_host.width,
+                                                                 template_host.height));
 }
 
 spl::shared_ptr<core::frame_producer> create_swf_producer(const core::frame_producer_dependencies& dependencies,
                                                           const std::vector<std::wstring>&         params)
 {
-    auto filename = env::media_folder() + L"\\" + params.at(0) + L".swf";
+    auto filename = env::media_folder() + params.at(0) + L".swf";
```
ronag, line 594: This will break some configurations that don't end with slash

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -555,27 +575,35 @@ spl::shared_ptr<core::frame_producer> create_producer(const core::frame_producer
...
 
     if (!boost::filesystem::exists(filename))
         return core::frame_producer::empty();
 
     swf_t::header_t header(filename);
 
-    auto producer = spl::make_shared<flash_producer>(
-        dependencies.frame_factory, dependencies.format_desc, filename, header.frame_width, header.frame_height);
```
ronag, line 578: `const url = filename`

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -555,27 +575,35 @@ spl::shared_ptr<core::frame_producer> create_producer(const core::frame_producer
...
     auto template_host = get_template_host(dependencies.format_desc);
 
-    auto filename = env::template_folder() + L"\\" + template_host.filename;
+    auto filename = env::template_folder() + template_host.filename;
 
     if (!boost::filesystem::exists(filename))
         CASPAR_THROW_EXCEPTION(file_not_found() << msg_info(L"Could not open flash movie " + filename));
 
-    return create_destroy_proxy(spl::make_shared<flash_producer>(
-        dependencies.frame_factory, dependencies.format_desc, filename, template_host.width, template_host.height));
```
ronag, line 564: `const url = url_from_path(filename)`

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -555,27 +575,35 @@ spl::shared_ptr<core::frame_producer> create_producer(const core::frame_producer
 {
     auto template_host = get_template_host(dependencies.format_desc);
 
-    auto filename = env::template_folder() + L"\\" + template_host.filename;
+    auto filename = env::template_folder() + template_host.filename;
```
ronag, line 578: This will break some configurations that don't end with slash

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -153,6 +157,22 @@ std::mutex& get_global_init_destruct_mutex()
...
+    HRESULT      ret     = UrlCreateFromPathW(in.c_str(), out_buf, &out_length, NULL);
+    std::wstring out;
+    if (SUCCEEDED(ret)) {
+        out = out_buf;
+        free(out_buf);
+        return out;
+    } else {
+        free(out_buf);
+        return in;
+    }
```
ronag, line 173: Would prefer a boost based implementation.

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -153,6 +157,22 @@ std::mutex& get_global_init_destruct_mutex()
     return m;
 }
 
+std::wstring url_from_path(std::wstring in)
+{
+    DWORD        out_length = INTERNET_MAX_URL_LENGTH * 2 + 4;
+    PWSTR        out_buf = (PWSTR)malloc(out_length);
```
ronag, line 163:
```cpp
SCOPE_EXIT {
  free(out_buf);
};

```

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -153,6 +157,22 @@ std::mutex& get_global_init_destruct_mutex()
     return m;
 }
 
+std::wstring url_from_path(std::wstring in)
+{
+    DWORD        out_length = INTERNET_MAX_URL_LENGTH * 2 + 4;
```
ronag, line 162: Why +4?

src/modules/flash/producer/flash_producer.cpp
silid: In practice I've found this is not the case and media_folder is always returned with a slash regardless of what was in the config and so the path was always called with a double slash.

This is because trailing slash is already enforced here:
https://github.com/CasparCG/server/blob/94db7f222d852947eb61ae10b3ac6c9cbbd785be/src/common/env.cpp#L107-L110

This is used already in the image producer:
https://github.com/CasparCG/server/blob/4a8f03f3fd4eedf437e5eac3ea1ff20c817c7c22/src/modules/image/producer/image_producer.cpp#L199

src/modules/flash/producer/flash_producer.cpp
silid: Trailing slash is already enforced here and this was resulting in a double slash:
https://github.com/CasparCG/server/blob/94db7f222d852947eb61ae10b3ac6c9cbbd785be/src/common/env.cpp#L107-L110

src/modules/flash/producer/flash_producer.cpp
silid: Now using const

src/modules/flash/producer/flash_producer.cpp
silid: now using const.

src/modules/flash/producer/flash_producer.cpp
silid: This is to allow for null termination of wstring.

src/modules/flash/producer/flash_producer.cpp
```diff
@@ -153,6 +158,20 @@ std::mutex& get_global_init_destruct_mutex()
...
 
+std::wstring url_from_path(std::wstring in)
+{
+    DWORD        out_length = INTERNET_MAX_URL_LENGTH * 2;
+    PWSTR        out_buf = (PWSTR)malloc(out_length + 4);
+    CASPAR_SCOPE_EXIT { free(out_buf); };
+    HRESULT      ret     = UrlCreateFromPathW(in.c_str(), out_buf, &out_length, NULL);
+    std::wstring out;
+    if (SUCCEEDED(ret)) {
+        return out_buf;
```
ronag, line 169: `std::wstring(out_buf, out_length)`

src/modules/flash/producer/flash_producer.cpp
ronag: I think it should be `const auto`?

src/modules/flash/producer/flash_producer.cpp
ronag: I think it should be const auto?

didikunz:
As a c++ and Git noob I still see this as being open. Is there any problem left to solve? Or will it still show open after merged in?

didikunz:
HELLO, ANYBODY HOME? Is there a reason, why we can not merge this in before the deadline? It's now 17 days since this pull request is open. People need to be able to update their systems BEFORE Christmas and some of them need to have their IT department do that for them and we all know, how long that takes. So I would love to see this done today, please.

ronag (approved):
LGTM

ronag:
@Julusian wdyt?

ronag:
> HELLO, ANYBODY HOME? Is there a reason, why we can not merge this in before the deadline? It's now 17 days since this pull request is open. People need to be able to update their systems BEFORE Christmas and some of them need to have their IT department do that for them and we all know, how long that takes. So I would love to see this done today, please.

We are low on active maintainers at the moment.

didikunz:
Sorry to annoy you, but it still says "Open" is it merged now or not?

ronag:
> Sorry to annoy you, but it still says "Open" is it merged now or not?

No, it needs an approval from @Julusian

didikunz:
Aha, I see

didikunz:
@Julusian: Do we get that done today?

PR #1355, opened 2020-12-07
v2.3.1

Julusian:
@ronag are you happy for your ffmpeg changes to be in a 2.3.1 release?
It looks like the `SEEKABLE` property changes default behaviour and my query could do with concluding on, so maybe that change should be held back for now

### DIFF
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -602,10 +602,7 @@ struct decklink_consumer : public IDeckLinkVideoOutputCallback
             }
         }
 
-        if (!frame) {
-            return !abort_request_;
-        }
-
+        if (frame)
         {
             std::unique_lock<std::mutex> lock(buffer_mutex_);
             buffer_cond_.wait(lock, [&] { return buffer_.size() < buffer_capacity_ || abort_request_; });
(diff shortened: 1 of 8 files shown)


ronag:
@Julusian I'll try to look into seekable this week.

Julusian:
@ronag Thanks. I have a deadline of the 16th to have this released, but sooner is better so I would like to get it done this week. It only needs to have the flash fix, but it would be nice to have your other fixes too.

jeclark-adobe:
From the Flash Player side of things, we've been pursuing a generalized workaround for desktop applications that embed the system copy of the ActiveX Flash Player and pass non-conforming file paths.  We shipped that workaround yesterday, in the form of the EnableInsecureAllowListPathMatching flag.  This also happens to be the last generally available release of Flash Player that Adobe will ship.

Patching CasparCG is absolutely the better choice, but if you're an administrator with constraints that making taking the fix prohibitive, this might be useful as a stop-gap.  There are security implications, which is why it's disabled by default and begins with the EnableInsecure* pattern.  It should be a last resort, and some condiseration should be given to whether or not clients where this is required to run are sufficiently isolated from confidential file stores through permissions and firewall rules, etc.

You can read more about the EnableInsecureAllowListPathMatching flag and the Enterprise Enablement feature in general in the Flash Player System Administrator's Guide, here: 

https://www.adobe.com/content/dam/acom/en/devnet/flashplayer/articles/flash_player_admin_guide/pdf/latest/flash_player_32_0_admin_guide.pdf

Julusian:
@ronag Im going to pick around those seek changes for now so that I can get this done. Once you have looked at it then it can be included in the next patch release

PR #1362, opened 2021-01-25
Update Intel TBB library to v2021.1.1

scriptorian:


ronag:
Wrong branch? Should land on master first?

Julusian:
I asked for this to be against this branch to avoid pulling the other changes from master into 2.3.2
But I should have also merged it to master at the same time

nerijus:
As I understand only Windows build uses TBB v2021.1.1, Linux (docker) build uses system tbb library, right? If so, is it possible to make docker build use the new TBB version as well?

PR #1371, opened 2021-02-21
Allow system-audio to specify output device

pbelbin:
Logs the current available OpenAL devices.

Allows .config file to have within the screen (which uses OpenAL for audio) consumer element:
```
<system-audio>
   <device-name>[device name from OpenAL device list]</device-name>
</system-audio>
```

When matching device is found, use that instead of system default.


src/modules/oal/consumer/oal_consumer.cpp
```diff
@@ -62,15 +62,41 @@ extern "C" {
...
 {
-    ALCdevice*  device_  = nullptr;
-    ALCcontext* context_ = nullptr;
+    ALCdevice*          device_  = nullptr;
+    ALCcontext*         context_ = nullptr;
+    const configuration config;
 
   public:
-    device()
+    device(const configuration& config)
```
dimitry-ishenko, line 77: Since you are storing a copy of `config` inside `device`, you can pass by value and take advantage of the move semantics.

pbelbin:
Who can review and commit this change?

I saw a comment emailed to me re comment from @dimitry-ishenko, but I no longer see the comment here, so I suppose it was retracted/deleted before I can respond?

Please advise!

ronag (approved):

dimitry-ishenko:
@pbelbin I was going to suggest to use move semantics when passing around `configuration` and replace C-style casts with `static_cast`, but then I've realized I didn't have the time to examine the code and decided it wasn't important :smile:

pbelbin:
I discovered something interesting while testing another usage scenario.

Initially, I was using CasparCG to use 2 video outputs (alpha and fill) as a single channel, with a single audio output (to the 'fill' video output).  This works nicely.

But, I then started trying use of 4 video outputs (2 channels, each with alpha and fill), and each with it's own audio output (or so I thought).

With the 2 channel scenario, I see in the logging that I have added, that the enumeration of available audio devices is happening twice, and each time, the configured output audio device names are shown and found (which suggests they are both being used, too), **but**, audio for both sets of playout channels *only ever appears in one of the output devices*.  It's like it's ignoring that the 2nd channel has it's own audio device.

Because I happen to have CasparCG feeding all video outputs to a single video switch, it's not a huge problem.  But, it's a bit disconcerting, and, would cause issues if the 2 output channels were being fed to 2 totally different switchers!

Can someone provide some additional information here?  What is *supposed* to happen as far as audio handling of multiple video card outputs being used as separate channels?

Without knowing more about how CasparCG is designed in this respect, it's hard to say whether this is 'normal' and 'expected', or whether it's a 'bug'.

pbelbin:
It *appears* that this is some kind of known behavior with plain video card outputs, [based on this comment](https://casparcgforum.org/t/how-to-not-use-system-default-audio-device/3939/8), but I have to say, I think it's pretty strange for this to be 'normal behavior'.

Perhaps this will be improved in the future, but it appears, this is *not a bug*.

pbelbin:
hey there @Julusian, do you have an idea as to when this will be reviewed?

pbelbin:
Hey @ronag , it's been nearly a month, and @Julusian has not provided any feedback or response at all.  What normally happens in this project when that sort of thing happens? Can/should someone else be tagged to review it?

src/modules/oal/consumer/oal_consumer.cpp
```diff
@@ -96,18 +122,59 @@ class device
...
     ALCdevice* get() { return device_; }
+
+  private:
+    ALCchar* iterateDevices(const char* list)
+    {
+        ALCchar* result = nullptr;
+        ALCchar* ptr;
+
+        // generate ascii string for comparison purposes vs what openAL provides
+        std::string short_device_name(config.audio_device_name.begin(), config.audio_device_name.end());
```
Julusian, line 133: This should use the `u8()` from https://github.com/CasparCG/server/blob/master/src/common/utf.cpp to ensure the value is converted appropriately

src/modules/oal/consumer/oal_consumer.cpp
```diff
@@ -96,18 +122,59 @@ class device
...
 };
 
-void init_device()
+void init_device(configuration config)
 {
     static std::unique_ptr<device> instance;
     static std::once_flag          f;
 
-    std::call_once(f, [] { instance = std::make_unique<device>(); });
+    std::call_once(f, [&] { instance = std::make_unique<device>(config); });
```
Julusian, line 172: I don't think this should be using call_once anymore. A singleton means it will open the first device and then reuse that for every usage. 
Have you tried running 2 channels with different outputs?

Julusian (changes requested):

src/modules/oal/consumer/oal_consumer.cpp
pbelbin: Ok, I have made some changes based on your feedback, appears to be working.

src/modules/oal/consumer/oal_consumer.cpp
pbelbin: If you look at my comments around [Feb 21](https://github.com/CasparCG/server/pull/1371#issuecomment-782958310), you should see that I *have* tried using multiple output devices.

I gather though, that you're saying that the line you're calling out here is likely the cause of the first output device being the only device that's used.

So this should just be unwrappered, and let the call to make_unique<device>() be done each time?

Hmmm.  Will there be any consequences of this change that need to be accounted for elsewhere, such as: memory leaks?

pbelbin: I wonder if it's possible that this might have something also to do with video and audio getting out of sync in some situations?

Julusian: I have given this a test myself with 2 channels defined and as I suspected I am seeing the available devices being listed only once. Which aligns with what this line is showing being called just once

From reading around the existing code more this makes sense in how it is written, and is going to require a dive into understanding openal and how to handle threading with it. Simply unwrapping that call_once will not work for a few reasons, the main one is that openal has a context which has a device assigned. The current context is the one that is used, and by default is shared across all threads. The `ALC_EXT_thread_local_context` extension will need to be used to handle having a context per consumer with all openal operations being done inside the consumer `executor_`. This is instead of how we currently create the context the first time init_device is called

I have no idea if creating multiple contexts for the same output device will cause issues, that is something that should be tested once the rest is implemented.

pbelbin: Does that mean you want to back off from changing this for the time being?

pbelbin: I'd be happy to get this merged in with the singleton call for the time being, as, at least, this will allow the single audio output device to be selectable.  What you're describing seems to be more along the lines of allowing support for multiple audio output devices, which, as you say, seems to be a bigger effort.  Not saying that bigger effort shouldn't be done, but, I wonder if it should be done, instead, within a separate request.  What do you think?

Julusian: Yeah, I am ok with being able to just change the default for now. But then I think this needs a couple of other (simple) changes to make it fit that better.
Instead of being able to define the device-name inside the consumer configuration, there device name should be defined at the root level, in a block similar to these other producers/consumers https://github.com/CasparCG/server/blob/5e6e83f5d6645a147d1f25eab23dd7aaf605ca52/src/shell/casparcg.config#L49
And I dont think it should be possible to pass the device name via AMCP.

pbelbin: Ok, I checked in a change to not allow AMCP to specify the audio device, but, I think it's creating extra work to have the audio device being indicated as a top-level thing.  With the limitation that there's only allowed to be 1 audio output per instance of CasparCG when using 'system-audio', given we already have 'system-audio' being a child element of the consumer, *and* that we'd like to make it multi-audio device moving forward, it makes more sense to me to leave it where it is currently.  The documentation should be updated to make it clear that for a system-audio output device, there can only be 1 such device (per running instance of CasparCG)), and when the time comes that multi-audio devices are supported, then this restriction will simply be relaxed.  No having to move it.  I didn't make 'system-audio' be where it is.  But, keeping it where it is is very convenient.

pbelbin: Hmmmm.  Regarding the <system-audio> being located as a direct child of the top level element.....

Would you please explain the reasoning behind wanting to have a <system-audio> section as a top-level child section?

As with most things like this, there are *many* ways it *could* be done.

To me, the <system-audio> is a <consumer>, so it makes sense to be there, but, for multiple devices, it makes sense there'd be also an id (or perhaps just the existing device-name> used to allow multiple <system-audio> sections, and let the <screen> identify which <system-audio> it wants to use for audio output.

eg:

```xml
<consumers>
   <system-audio>
      <device-name>audio device 1</device-name>
   </system-audio>
   <system-audio>
      <device-name>audio device 2</device-name>
   </system-audio>
.
.
   <screen>
      <device>1</device>
      <system-audio-device>audio device 1</system-audio-device>
   </screen>
   <screen>
      <device>2</device>
      <system-audio-device>audio device 1</system-audio-device>
   </screen>
```

The above example shows 2 different video outputs, both of which are sending audio out via the `audio device 1` audio output.

not sure how the management of the summing of both audio streams would be managed in this scenario, but, at least the configuration would allow it?

pbelbin: I suppose, it might also be *possible* to, at start-up, scan the screen consumers for references to audio devices, and 'automagically' add a definition where a matching system-audio is not found?

Julusian: I mean that at the root level should be a config like:
```
<system-audio>
	<default-device>something</default-device>
</system-audio>
```

Once it is possible to use multiple devices simultaneously, then it can also be defined under the channel config:
```
<channel>
	<consumers>
		<system-audio>
		   <device-name>[device name from OpenAL device list]</device-name>
		</system-audio>
	</consumers>
</channel>
```
And if device-name is omitted there, it should fall back to the default device specified at the root level.


My concern is that with the single device being defined inside the channels it implies that it is possible to use multiple at once, and you will have to be careful when editing the config to not accidentally change the selected system audio device by changing the order

pbelbin: > Once it is possible to use multiple devices simultaneously, then it can also be defined under the channel config:

Ah, but, that's an issue, isn't it?  The way you're describing it, I don't see a connection being made that can be used to know that when video is played out via a screen consumer, *which* audio device should be used.

If a system has multiple audio consumers and video consumers, which audio and video devices are being paired up, so that when video is sent to a particular consumer, the accompanying audio gets sent to a particular audio device?

There needs to be *something* that says: when video is played out, to send audio to a specific audio consumer too.  In the case of the decklink consumer, it seems to have an 'embedded-audio' element.

In the case of displayport or hdmi outputs, I suppose you might say it's similar, that the device *also* has the ability to carry audio.

But, in the case of someone using a totally different audio device from video (eg: external usb audio interface), then they're totally separate devices, and there needs to be a way to say that when video is played out on screen consumer 1, that audio goes out via this specific other device.

pbelbin: Most of this discussion is about how to make multiple audio devices workable.  The current issue is just about being able to make the existing configuration be able to identify which single device should be used.

With *this* issue, I just want to allow the existing global system audio element to be able to indicate *which* audio device to consistently use.  I really don't see the point in adding *another* element, since it's clear that there is only a single system audio consumer supported today, and I think the element being where it is indicates this fairly well.

Julusian: Im going to skip the parts about multiple devices, so we can focus on getting this merged.

Except it is perfectly valid and works fine to have multiple system audio consumers. They all use the same device, but their audio gets combined and output.
This is a fragment of my config for this:
```
  <channels>
    <channel>
      <consumers>
        <decklink>
          <device>1</device>
                <embedded-audio>true</embedded-audio>
                <latency>normal</latency>
                <buffer-depth>5</buffer-depth>
        </decklink>
        <system-audio>
          <device-name>DVS Transmit 3-4 (Dante Virtual Soundcard)</device-name>
        </system-audio>
      </consumers>
    </channel>
    <channel>
      <consumers>
        <decklink>
          <device>2</device>
                <embedded-audio>true</embedded-audio>
                <latency>normal</latency>
                <buffer-depth>5</buffer-depth>
        </decklink>
        <system-audio></system-audio>
      </consumers>
    </channel>
  </channels>
```

Why should the device-name be set in the first `<system-audio>` tag, and not outside of the channels tree?

Additionally, if we do go with setting the device-name inside the channels tree, then that will make supporting multiple devices be a breaking change. As the behaviour of any channel with an empty `<system-audio></system-audio>` may change. It will no longer follow the one before it, but will use the system default device.

This config is more durable, explanatory and future proof:
```
  <system-audio>
	<default-device>DVS Transmit 3-4 (Dante Virtual Soundcard)</default-device>
  </system-audio>
  <channels>
    <channel>
      <consumers>
        <decklink>
          <device>1</device>
                <embedded-audio>true</embedded-audio>
                <latency>normal</latency>
                <buffer-depth>5</buffer-depth>
        </decklink>
        <system-audio></system-audio>
      </consumers>
    </channel>
    <channel>
      <consumers>
        <decklink>
          <device>2</device>
                <embedded-audio>true</embedded-audio>
                <latency>normal</latency>
                <buffer-depth>5</buffer-depth>
        </decklink>
        <system-audio></system-audio>
      </consumers>
    </channel>
  </channels>
```

pbelbin: I would like to avoid having to make changes that work great for just this change, but when multiple audio device support happens, causes further breaking changes.

So, given the XML sample you've provided above (the 'durable' one), what changes would you then make when multiple audio device support is added?

Julusian: When we support using multiple audio devices, then we will add the property inside the channel consumers block:
```
    <channel>
      <consumers>
        <system-audio>
          <device-name>DVS Transmit 3-4 (Dante Virtual Soundcard)</device-name>
        </system-audio>
      </consumers>
    </channel>
```

If a consumer doesn't specify `device-name`, then the one specified by the `default-device` property will be used instead.

Sidonai-1:
Very excited to try the next build after the merge.

In the end, are we allowed to specify different system-audio devices in the same config? That opens up Caspar to feed Dante networks.

I'd appreciate an example of how to set the devices, I guess I'll find one in the config file itself.

Thanks a lot!

Julusian:
> In the end, are we allowed to specify different system-audio devices in the same config? 

Not yet, this still only allows for one system audio device to be used.

PR #1372, opened 2021-02-21
feat: port 2.1 image scroll producer

ronag:


ronag:
@5opr4ni this is ready for you to be verified

5opr4ni:
> @5opr4ni this is ready for you to be verified

I’ll have a look. Was in transmission when I saw it, and almost missed putting on some graphics. 🤣 (Mello)

dotarmin:
@ronag This is not building unfortunately on the build server. It complains about unresolved external symbols and the build fails with two errors of the same type (see below).

`error LNK2019: unresolved external symbol` 

Other branches works as expected.

ronag:
Probably missing a reference to the cpp file in some make file.

dotarmin:
> Probably missing a reference to the cpp file in some make file.

Roger that. Let me know when you have looked at it so I can run the build.

5opr4ni:
All expected commands works as expected regarding to function but not time.
Duration and end_time are probably based on another time. (A DURATION of 20 took 40 seconds.)
These commands where tested:
PLAY 1-10 mellocred speed 8

PLAY 1-10 mellocred speed 4 BLUR 4

PLAY 1-10 mellocred speed 8 BLUR 10

PLAY 1-10 hscrolltest speed 8 BLUR 10

PLAY 1-10 mellocred DURATION 20 BLUR 4

PLAY 1-10 mellocred END_TIME "2021-03-10 14:45:17"

5opr4ni (changes requested):
All expected commands works as expected regarding to function but not time.
Duration and end_time are probably based on another time. (A DURATION of 20 took 40 seconds.)

src/modules/image/producer/image_scroll_producer.cpp
```diff
@@ -0,0 +1,471 @@
...
+
+        if (vertical)
+            return height_ + format_desc_.height;
+        else
+            return width_ + format_desc_.width;
+    }
+
+    double speed_from_duration(double duration_seconds) const
+    {
+        return get_total_num_pixels() / (duration_seconds * format_desc_.fps * static_cast<double>(format_desc_.field_count));
```
ronag, line 281:
```suggestion
        return get_total_num_pixels() / (duration_seconds * format_desc_.fps);
```

5opr4ni (approved):

5opr4ni:
Tested hard during rehearsals and broadcast of Melodifestivalen. Worked flawlessly! 🥳

PR #1394, opened 2021-10-21
fix: incorrect variable name in FindFFmpeg

oscarwcl:
It seems like the prefix passed into `pkg_check_modules` is a mistake:
later calls expect `PC_LIB${_component}_*` to be set by
`pkg_check_modules`, not `${_component}_*`

Before this change, pkg_check_modules would set the
`${_component}_INCLUDE_DIRS` variable. The later call to `find_path`
also expects to set this variable, but doesn't if it has already been
set. On some systems, (e.g. Arch Linux), `pkg_check_modules` sets this
variable to an empty string as the header files are in the default
include directory. Since `find_path` won't overwrite the variable with
the correct value, it will appear like the include directories weren't
found, even if they do exist.

### DIFF
--- a/src/CMakeModules/FindFFmpeg.cmake
+++ b/src/CMakeModules/FindFFmpeg.cmake
@@ -63,7 +63,7 @@ macro(find_component _component _pkgconfig _library _header)
      # in the FIND_PATH() and FIND_LIBRARY() calls
      find_package(PkgConfig)
      if (PKG_CONFIG_FOUND)
-       pkg_check_modules(${_component} ${_pkgconfig})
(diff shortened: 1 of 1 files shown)


Julusian:
I'm having to revert this, as it breaks building on linux on ubuntu when using our built in docker version of ffmpeg, It might affect system versions of it too, I have not tested.
Having that `PC_LIB` prefix causes `FFMPEG_LIBRARIES` to not include any of the dependencies it needs to link against (eg freetype xcb). I expect it is probably a case of needing other bits in hat file to be updated to also use the PC_LIB prefix lines, but I do not know cmake well enough to do that myself

PR #1399, opened 2021-11-06
608/708 Closed Captions out Decklink 4k Extreme

tvadi:
Hi Bob, I have search all over and you seem to be the only one that seems to know anything about getting CasparCG to output embedded captions out a Decklink SDI.  Does your version work for that?  Of have any guidance?  Thanks so much!


Julusian:
To contact the author of the fork that would be best done on their fork and not here, as they will not have been notified of this

PR #1402, opened 2021-11-22
ffmpeg producer: Make decoder wait for capacity.

### LINKED ISSUE #1385: Spectacular memory leak in build from 865502c on Ubuntu 18.04
Made a build in docker from current master (at 865502c). 

As soon as you start a (h264) clip in that build memory starts running op very fast. 16gb is exhausted in 30 secs or so.

Logs don't say much that seems relevant, but there are lots of latency warnings when memory starts getting full 

Probably has something to do with recent commits related to threading

gizahNL:
Fixes #1385


toresbe:
This patch has been tested on Docker in Debian and fixes the issue, which for us led to downtime.

PR #1404, opened 2021-11-22
Abstract setting thread priorities & implement for linux.

gizahNL:
Currently only done for THREAD_PRIORITY_TIME_CRITICAL which is the
highest (realtime priority class) available on Windows. SCHED_RR
should work for that.


gizahNL:
Setting linux to a RT priority of 99 is needed here otherwise we lose about 1 frame per minute.

gizahNL:
> Setting linux to a RT priority of 99 is needed here otherwise we lose about 1 frame per minute on my testbench playing 4 HD channels out via NDI

99 caused some issues on duration testing, so settling on 49

PR #1410, opened 2021-11-24
opengl performance improvements

gizahNL:
Low hanging fruit: change glWaitSync behavior
Optimize fragment shader to discard when invisible (alpha < 0.01)
Change shader (less branching due to non-uniform flow control)
Using a separate shader program as a "fast path" & using openGL blend func (executed on GPU ROP, less texture reads) shaves off another ~5%

Total improvement from ~80-85% to ~50-55% utilization when running 4 HD 50i channels each with 2 layers, close to 2.1 GPU utilization.

### DIFF
--- a/src/accelerator/ogl/util/device.cpp
+++ b/src/accelerator/ogl/util/device.cpp
@@ -273,15 +273,13 @@ struct device::impl : public std::enable_shared_from_this<impl>
 
             auto fence = glFenceSync(GL_SYNC_GPU_COMMANDS_COMPLETE, 0);
 
-            //The whole loop could be eliminated by choosing a timeout close to infinity
-            //since the loop is unbounded that would also work.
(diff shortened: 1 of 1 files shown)


Julusian:
Trying this with 4x 1080i50 channels (each playing 2 AMB) on ubuntu 22.04 with a GTX1060, I am seeing gpu usage go from 40-45% to 38-42%, which is not a significant improvement. What gpu and os are you using?

On windows it gets stuck in an error loop when playing any media with `caspar::gl::ogl_invalid_framebuffer_operation_ext`

> Change shader (less branching due to non-uniform flow control)

It has been quite a while (~10 years) since I have had to think about optimising cuda code, but from what I remember branching is only an issue when threads in the same cluster make take different routes. So for us, different branches being used for each frame being composited should have no major impact?  

What is the cost of frequently switching shaders? some layers on a channel could be on the fast and some on the slow shader


As it currently stands, I am not convinced that this will give a noticeable performance benefit to most users, so I am not convinced it is worth the extra complexity

PR #1413, opened 2021-12-08
Meson poc

gizahNL:
Proof of concept of building via Meson (linux only atm)

Biggest advantage is native support for precompiled headers, that can be enabled & disabled at will, as well as with debug options.

Also a speed boost (though part of it can be explained due to cmake building with -O3):
sh -c 'cmake -GNinja ../src && ninja'  723,42s user 44,86s system 794% cpu 1:36,74 total
sh -c 'meson ../src && ninja'  353,33s user 40,67s system 848% cpu 46,412 total

cmake is still intact (tested on Linux)


dimitry-ishenko:
@gizahNL have you tried enabling [target_precompiled_headers](https://cmake.org/cmake/help/git-stage/command/target_precompile_headers.html) with cmake?

gizahNL:
> @gizahNL have you tried enabling [target_precompiled_headers](https://cmake.org/cmake/help/git-stage/command/target_precompile_headers.html) with cmake?

I have not :sweat_smile: I always end up in a fight with Cmake when I touch it... Also the systems I target don't support that cmake feature :/

Julusian:
I know nothing about meson but I personally am open to it. Others in the project probably need to give this kind of change approval too though

My main question is can windows use the same meson files like we are doing today with cmake? Avoiding breaking the other platform is annoying enough at times already, so I dont want to make it worse.

If you pick this up again, I am curious how cmake&make compare to cmake&ninja.

src/shell/meson.build
```diff
@@ -51,3 +56,15 @@ server = executable(
     install_rpath: '$ORIGIN/../lib'
 )
 
+staging_args = ['./copy_staging.sh', '@BUILD_ROOT@', '@SOURCE_ROOT@', '@INPUT@']
+if have_cef
+    staging_args += cef_path
+endif
+
+staging = custom_target(
```
eli-schwartz, line 64: Maybe it makes more sense to have this as a run_target?

src/shell/meson.build
```diff
@@ -51,3 +56,15 @@ server = executable(
...
+    'staging',
+    input: server,
+    output: 'fake',
+    command: staging_args,
+    build_always: true,
```
eli-schwartz, line 69: This is deprecated since Meson 0.47.0 (if you include `meson_version: '>= ...'` in your project() arguments it will ensure users have a new enough Meson, and will issue warnings if you use features newer than what you claim to support).

It's replaced by `build_always_stale: true` and `build_by_default: true`, which may not both be desired.

Julusian:
I don't think this is a direction we want to go in for now. There is some work underway to improve the existing cmake configs, which I think resolves the comment about precompiled headers.

I have tried to compare the meson vs cmake with the `-O3` disabled for cmake to do a fair comparison, but I am getting many weird compiler errors.  

So I don't see a compelling reason to switch to meson right now

PR #1415, opened 2022-02-07
Added automatic duplex setting of Decklink device (Decklink SDK update)

pkeuter:
I often notice that the Decklink duplex settings have been reset after doing a Windows update. When this happens, it yields some inconsistent behaviour.

To handle this, I've added a configuration setting which provides a way to consistently set the duplex-behavior of the specified channel e.g.:

```
<decklink>
  <device>1</device>
  <duplex>full</duplex>
</decklink>
```
Full is a connected output (`SDIx&SDIy`). Other options are `half` which is just `SDIx`. Setting nothing keeps the current behavior (leave the setting as is).

Because this functionality was not available in the current SDK provided with CasparCG which is quite old, I have also upgraded this.

There is some functionality change in the `DoesSupportVideoMode`-function which I've modified but as far as I can see this has no implications other than a slightly changed log-line.


Julusian:
What is the oldest sdk version that supports this property?
I think that users need to be using drivers the same version or newer than the version of the sdk we use, so it would be nice to not use the absolute latest if possible

pkeuter:
I need to look that up. You are right by the way, you need a driver version equal to or higher than the SDK used. 

My time is limited so I'll get back on that version (and likely modified PR) later.

pkeuter:
superseded by #1416

PR #1416, opened 2022-02-14
Added automatic duplex setting of Decklink device (Decklink SDK update 10.6.6) 

pkeuter:
This supersedes https://github.com/CasparCG/server/pull/1415. It is the same change, but with the Decklink SDK updated to 10.6.6 instead of 12.2.2


pkeuter:
@Julusian this should be more to your liking

PR #1419, opened 2022-02-22
fix: compatibility with ffmpeg 5.0

oscarwcl:
See individual commits for what each change is.

The final commit does effectively revert 7ac31ee94f810a41c931e2000f11c874b7da7930 for ffmpeg >= 5.0. I'm not sure exactly what this is doing or why it's necessary, but it doesn't seem to be possible anymore in ffmpeg 5.0. I don't know if there's another way of doing something equivalent. I've added a condition to keep the old behaviour on ffmpeg 4.x, but this doesn't seem like a very good long-term solution, so any suggestions here are appreciated.

None of these changes should affect behaviour on ffmpeg 4.x.

### DIFF
--- a/src/modules/ffmpeg/producer/av_input.cpp
+++ b/src/modules/ffmpeg/producer/av_input.cpp
@@ -126,7 +126,11 @@ void Input::internal_reset()
 
     static const std::set<std::wstring> PROTOCOLS_TREATED_AS_FORMATS = {L"dshow", L"v4l2", L"iec61883"};
 
+#if LIBAVFORMAT_VERSION_MAJOR >= 59
+    const AVInputFormat* input_format = nullptr;
+#else
     AVInputFormat* input_format = nullptr;
+#endif
     auto           url_parts    = caspar::protocol_split(u16(filename_));
     if (url_parts.first == L"http" || url_parts.first == L"https") {
         FF(av_dict_set(&options, "multiple_requests", "1", 0)); // NOTE https://trac.ffmpeg.org/ticket/7034#comment:3
--- a/src/modules/ffmpeg/util/av_util.cpp
+++ b/src/modules/ffmpeg/util/av_util.cpp
@@ -11,6 +11,7 @@ extern "C" {
 #include <libavfilter/avfilter.h>
 #include <libavutil/channel_layout.h>
 #include <libavutil/frame.h>
+#include <libavutil/imgutils.h>
 #include <libavutil/pixfmt.h>
 }
 #if defined(_MSC_VER)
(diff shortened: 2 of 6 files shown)


src/modules/ffmpeg/util/av_util.cpp
```diff
@@ -122,47 +124,49 @@ core::pixel_format get_pixel_format(AVPixelFormat pix_fmt)
 core::pixel_format_desc pixel_format_desc(AVPixelFormat pix_fmt, int width, int height, std::vector<int>& data_map)
 {
     // Get linesizes
-    AVPicture dummy_pict;
-    avpicture_fill(&dummy_pict, nullptr, pix_fmt, width, height);
+    int linesizes[4];
+    av_image_fill_linesizes(linesizes, pix_fmt, width);
```
ducthiem90, line 128: You can use function av_image_fill_arrays in place of avpicture_fill
uint8_t* data[4];
int linesizes[4];
av_image_fill_arrays(data, linesizes, nullptr, pix_fmt, width, height, AV_FRAME_ALIGN);

oscarwcl: Yes, although the data pointers are only used in one of the conditions below, so it seems cleaner to only create them when they're needed.

petterreinholdtsen:
This is also issue https://bugs.debian.org/1015780 .

src/modules/ffmpeg/producer/av_input.cpp
```diff
@@ -126,7 +126,7 @@ void Input::internal_reset()
 
     static const std::set<std::wstring> PROTOCOLS_TREATED_AS_FORMATS = {L"dshow", L"v4l2", L"iec61883"};
 
-    AVInputFormat* input_format = nullptr;
+    const AVInputFormat* input_format = nullptr;
```
Julusian, line 129: Error	C2664	'int avformat_open_input(AVFormatContext **,const char *,AVInputFormat *,AVDictionary **)': cannot convert argument 3 from 'const AVInputFormat *' to 'AVInputFormat *'	ffmpeg	C:\Code\casparcg\official-master\src\modules\ffmpeg\producer\av_input.cpp	158

src/modules/ffmpeg/util/av_util.cpp
```diff
@@ -124,47 +126,51 @@ core::pixel_format get_pixel_format(AVPixelFormat pix_fmt)
...
         }
         case core::pixel_format::ycbcr:
         case core::pixel_format::ycbcra: {
             // Find chroma height
-            auto size2 = static_cast<int>(dummy_pict.data[2] - dummy_pict.data[1]);
-            auto h2    = size2 / dummy_pict.linesize[1];
+            size_t sizes[4];
+            ptrdiff_t linesizes1[4];
+            for (int i = 0; i < 4; i++) linesizes1[i] = linesizes[i];
+            av_image_fill_plane_sizes(sizes, pix_fmt, height, linesizes1);
```
Julusian, line 158: Error	C3861	'av_image_fill_plane_sizes': identifier not found	ffmpeg	C:\Code\casparcg\official-master\src\modules\ffmpeg\util\av_util.cpp	158

Julusian (changes requested):
I get a couple of build errors when compiling against ffmpeg 4.2.2. If later 4.2.x is better, then we can probably update to that.

I don't want to break compatability with ubuntu 20.04, which is using 4.2.7 https://packages.ubuntu.com/focal/libavcodec58

petterreinholdtsen:
The Debian package include an adjusted patch casting the value, which
seem to get the build going with older versions of ffmpeg, se
<URL: https://salsa.debian.org/multimedia-team/casparcg-server/-/commit/6ec271399ed26f84be1bf1483701e8032a15e47a >.
-- 
Happy hacking
Petter Reinholdtsen

oscarwcl:
I've now tested with both ffmpeg 5 and with the `build-in-docker` script and it seems to work. I've used a couple of preprocessor version checks to deal with incompatibilities between ffmpeg 4.x and ffmpeg 5 rather than casting as in the patch above to make it clearer that these changes can be removed once support for older ffmpeg versions is dropped.

Julusian:
Confirmed to still be building on windows against 4.2, and a simple playback test looked ok.

> I've used a couple of preprocessor version checks to deal with incompatibilities between ffmpeg 4.x and ffmpeg 5 rather than casting as in the patch above to make it clearer that these changes can be removed once support for older ffmpeg versions is dropped.

I like this approach. It definitely helps with the question of why something was done, and when it can be removed.

STL1811:
Hi, 
Do you plan to use FFMPEG 5.0 in windows build also ?

Julusian:
I have no immediate plans to do that.
With the zeranoe ffmpeg builds no longer existing, updating ffmpeg becomes a task of evaluating the replacement windows builds to make sure they are a suitable drop in replacement (similar/good set of compiled codecs).

Actually updating the windows builds is a pretty straightforward task, that simply involves me remembering how to publish to nuget

PR #1427, opened 2022-04-25
feat: upgrade cef to 95 (using custom builds from obs)

Julusian:
Windows build pulled from https://cdn-fastly.obsproject.com/downloads/cef_binary_4638_windows_x64.zip (built from https://github.com/obsproject/cef/commit/8092a57fcf4295015a1b2affe863b8ab6dc88de3)
And packaged with https://github.com/julusian/cef-binary


gizahNL:
Looking at the OBS code it looks like the new accelerated paint function would also work on non windows platforms? That would be very interesting!

Julusian:
https://github.com/obsproject/obs-studio/pull/5648
> The macOS & Windows builds include shared texture support

Unfortunately not linux yet by the sound of it. I haven't tried even building this on linux yet, but I need to make sure it builds so shall check at the same time

PR #1433, opened 2022-07-12
feat: subregion copy

Julusian:
Output a subregion of a channel to a decklink. 
No scaling/rotation support, for that use a dedicated channel and route.

Intended to be paired with a channel running at a custom resolution, with the output split across multiple decklinks.

TODO: Test with interlaced, to make sure the line handling is correct

### DIFF
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -893,7 +893,7 @@ create_preconfigured_consumer(const boost::property_tree::wptree&               
     config.embedded_audio    = ptree.get(L"embedded-audio", config.embedded_audio);
     config.base_buffer_depth = ptree.get(L"buffer-depth", config.base_buffer_depth);
 
-    auto format_desc_str = ptree.get(L"video-format", L"");
+    auto format_desc_str = ptree.get(L"video-mode", L"");
     if (format_desc_str.size() > 0) {
         auto format_desc     = core::video_format_desc(format_desc_str);
         if (format_desc.format == core::video_format::invalid)
@@ -901,12 +901,15 @@ create_preconfigured_consumer(const boost::property_tree::wptree&               
         config.format = format_desc.format;
     }
 
-    config.src_x    = ptree.get(L"src-x", config.src_x);
-    config.src_y    = ptree.get(L"src-y", config.src_y);
-    config.dest_x   = ptree.get(L"dest-x", config.dest_x);
-    config.dest_y   = ptree.get(L"dest-y", config.dest_y);
-    config.region_w = ptree.get(L"region-w", config.region_w);
-    config.region_h = ptree.get(L"region-h", config.region_h);
+    auto subregion_tree = ptree.get_child_optional(L"subregion");
+    if (subregion_tree) {
+        config.src_x    = subregion_tree->get(L"src-x", config.src_x);
+        config.src_y    = subregion_tree->get(L"src-y", config.src_y);
+        config.dest_x   = subregion_tree->get(L"dest-x", config.dest_x);
+        config.dest_y   = subregion_tree->get(L"dest-y", config.dest_y);
+        config.region_w = subregion_tree->get(L"width", config.region_w);
+        config.region_h = subregion_tree->get(L"height", config.region_h);
+    }
 
     return spl::make_shared<decklink_consumer_proxy>(config);
 }
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -94,9 +94,12 @@ struct configuration
     int       base_buffer_depth = 3;
 
     core::video_format format = core::video_format::invalid;
-    int                region_x = 0;
-    int                region_y = 0;
-    // TODO - copy_width, copy_height, dst_x, dst_y ?
+    int                src_x = 0;
+    int                src_y    = 0;
+    int                dest_x    = 0;
+    int                dest_y    = 0;
+    int                region_w = 0;
+    int                region_h = 0;
 
     int buffer_depth() const
     {
@@ -345,7 +348,9 @@ struct decklink_consumer : public IDeckLinkVideoOutputCallback
 
         int firstLine = topField ? 0 : 1;
 
-        if (channel_format_desc_.format == decklink_format_desc_.format) {
+        if (channel_format_desc_.format == decklink_format_desc_.format && config_.src_x == 0 && config_.src_y == 0 &&
+            config_.region_w == 0 && config_.region_h == 0 && config_.dest_x == 0 && config_.dest_y == 0) {
+            // Fast path
             size_t byte_count_line = (size_t)decklink_format_desc_.width * 4;
             for (int y = firstLine; y < decklink_format_desc_.height; y += field_count_) {
                 std::memcpy(reinterpret_cast<char*>(image_data.get()) + (long long)y * byte_count_line,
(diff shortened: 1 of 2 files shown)


MauriceVeraart:
Interesting what is your plan in controlling this.
Will this be configured in de config / how ?
is it already testable with say a 2160P channel ?

baltedewit:
> Will this be configured in de config / how ?

Config file changes are in the PR: https://github.com/CasparCG/server/blob/235eecf15ca8554710fe018d3a2f4d7d39460db4/src/shell/casparcg.config#L75

MauriceVeraart:
i tried it with the build #16ee0e3d available on builds.casparcg.com
and it does not work on my side just 1080P out with the normal channel content

or i am doing it wrong <= most likely ;)
or it's the buildserver. 
the changes in casparcg.config Balte referenced are not there either

see log

`[2022-07-13 13:27:58.196] [info]    ############################################################################
[2022-07-13 13:27:58.196] [info]    CasparCG Server is distributed by the Swedish Broadcasting Corporation (SVT)
[2022-07-13 13:27:58.196] [info]    under the GNU General Public License GPLv3 or higher.
[2022-07-13 13:27:58.196] [info]    Please see LICENSE.TXT for details.
[2022-07-13 13:27:58.196] [info]    http://www.casparcg.com/
[2022-07-13 13:27:58.196] [info]    ############################################################################
[2022-07-13 13:27:58.196] [info]    Starting CasparCG Video and Graphics Playout Server 2.3.3 16ee0e3d Dev
[2022-07-13 13:27:58.390] [info]    Initializing OpenGL Device.
[2022-07-13 13:27:58.391] [info]    Initialized OpenGL 4.5.0 NVIDIA 516.59 NVIDIA Corporation
[2022-07-13 13:27:58.391] [info]    Initialized image module.
[2022-07-13 13:27:58.393] [info]    Initialized ffmpeg module.
[2022-07-13 13:27:58.393] [info]    Initialized oal module.
[2022-07-13 13:27:58.393] [info]    Initialized decklink module.
[2022-07-13 13:27:58.393] [info]    Initialized screen module.
[2022-07-13 13:27:58.393] [info]    Initialized newtek module.
[2022-07-13 13:27:58.420] [info]    Initialized html module.
[2022-07-13 13:27:58.420] [info]    Flash support is disabled
[2022-07-13 13:27:58.420] [info]    Initialized flash module.
[2022-07-13 13:27:58.421] [info]    Initialized bluefish module.
[2022-07-13 13:27:58.422] [info]    "D:/casparcg-server-16ee0e3df5c2a0bebe80c9019e19255ac52c6418-windows\casparcg.config":
[2022-07-13 13:27:58.422] [info]    -----------------------------------------
[2022-07-13 13:27:58.422] [info]    <?xml version="1.0" encoding="utf-8"?>
[2022-07-13 13:27:58.422] [info]    <configuration>
[2022-07-13 13:27:58.422] [info]       <paths>
[2022-07-13 13:27:58.422] [info]          <media-path>media/</media-path>
[2022-07-13 13:27:58.422] [info]          <log-path>log/</log-path>
[2022-07-13 13:27:58.422] [info]          <data-path>data/</data-path>
[2022-07-13 13:27:58.422] [info]          <template-path>template/</template-path>
[2022-07-13 13:27:58.422] [info]       </paths>
[2022-07-13 13:27:58.422] [info]       <lock-clear-phrase>secret</lock-clear-phrase>
[2022-07-13 13:27:58.422] [info]       <channels>
[2022-07-13 13:27:58.422] [info]          <channel>
[2022-07-13 13:27:58.422] [info]             <video-mode>1080p5000</video-mode>
[2022-07-13 13:27:58.422] [info]             <consumers>
[2022-07-13 13:27:58.422] [info]                <decklink>
[2022-07-13 13:27:58.422] [info]                   <device>1</device>
[2022-07-13 13:27:58.422] [info]                   <video-mode>720p5000</video-mode>
[2022-07-13 13:27:58.422] [info]                   <subregion>
[2022-07-13 13:27:58.422] [info]                      <src-x>10</src-x>
[2022-07-13 13:27:58.422] [info]                      <src-y>10</src-y>
[2022-07-13 13:27:58.422] [info]                      <dest-x>20</dest-x>
[2022-07-13 13:27:58.422] [info]                      <dest-y>20</dest-y>
[2022-07-13 13:27:58.422] [info]                      <width>1280</width>
[2022-07-13 13:27:58.422] [info]                      <height>720</height>
[2022-07-13 13:27:58.422] [info]                   </subregion>
[2022-07-13 13:27:58.422] [info]                </decklink>
[2022-07-13 13:27:58.422] [info]             </consumers>
[2022-07-13 13:27:58.422] [info]             <producers>
[2022-07-13 13:27:58.422] [info]                <producer id="0">testchart</producer>
[2022-07-13 13:27:58.422] [info]             </producers>
[2022-07-13 13:27:58.422] [info]          </channel>
[2022-07-13 13:27:58.422] [info]       </channels>
[2022-07-13 13:27:58.422] [info]       <controllers>
[2022-07-13 13:27:58.422] [info]          <tcp>
[2022-07-13 13:27:58.422] [info]             <port>5250</port>
[2022-07-13 13:27:58.422] [info]             <protocol>AMCP</protocol>
[2022-07-13 13:27:58.422] [info]          </tcp>
[2022-07-13 13:27:58.422] [info]       </controllers>
[2022-07-13 13:27:58.422] [info]       <amcp>
[2022-07-13 13:27:58.422] [info]          <media-server>
[2022-07-13 13:27:58.422] [info]             <host>localhost</host>
[2022-07-13 13:27:58.422] [info]             <port>8000</port>
[2022-07-13 13:27:58.422] [info]          </media-server>
[2022-07-13 13:27:58.422] [info]       </amcp>
[2022-07-13 13:27:58.422] [info]    </configuration>
[2022-07-13 13:27:58.422] [info]    -----------------------------------------
[2022-07-13 13:27:58.425] [info]    Initialized OpenGL Accelerated GPU Image Mixer for channel 1
[2022-07-13 13:27:58.426] [info]    video_channel[1|1080p5000] Successfully Initialized.
[2022-07-13 13:27:58.440] [info]    DeckLink 8K Pro [1-1|1080p5000] Disabled low-latency mode.
[2022-07-13 13:27:58.440] [error]   DeckLink 8K Pro [1-1|1080p5000] Failed to enable external keyer.
[2022-07-13 13:27:58.445] [info]    DeckLink 8K Pro [1-1|1080p5000] Initialized.
[2022-07-13 13:27:58.445] [info]    Initialized channels.
[2022-07-13 13:27:58.445] [info]    Initialized command repository.
[2022-07-13 13:27:58.473] [info]    image_producer[media/testchart.tif] Initialized
[2022-07-13 13:27:58.484] [info]    DeckLink 8K Pro [1-1|1080p5000] Reference signal: not detected.
[2022-07-13 13:27:58.520] [info]    Initialized startup producers.
[2022-07-13 13:27:58.521] [info]    Initialized controllers.
[2022-07-13 13:27:58.521] [info]    Initialized osc.
[2022-07-13 13:28:06.099] [info]    Received message from Console: exit\r\n
[2022-07-13 13:28:06.101] [info]    [asio] Shutting down global io_service.
[2022-07-13 13:28:06.101] [info]    [asio] Global io_service uninitialized.
[2022-07-13 13:28:06.107] [info]    video_channel[1|1080p5000] Uninitializing.
[2022-07-13 13:28:06.145] [error]   DeckLink 8K Pro [1-1|1080p5000] Failed to schedule fill video.
[2022-07-13 13:28:06.145] [info]    DeckLink 8K Pro [1-1|1080p5000] Scheduled playback has stopped.
[2022-07-13 13:28:06.197] [info]    DeckLink 8K Pro [1-1|1080p5000] Uninitialized.
[2022-07-13 13:28:06.197] [info]    Successfully shutdown CasparCG Server.
`

Julusian:
> So if I understand correctly, Decklink outputs are still bound to broadcast supported formats but when making a custom video mode (let's say 2500×1500px) now you can split this channel into several 1920×1080 SDI outputs? (let's assume we don't have a 4k card I guess).

Yes exactly this.
We are still limited by the formats that the decklinks/sdi support. 
That split is exactly what we are looking at doing. It was completely possible before by using a channel to composite, and a channel to scale/crop for each decklink, but this is very wasteful.

It handles the decklink being smaller or bigger than the channel, so you could even do things like using a 1080p sdi to a video wall processor, but as you only need 100 pixels of height, you could run the channel in 1920x100 to save the gpu work, and the need to transform everything

Julusian:
> i tried it with the build #16ee0e3d available on builds.casparcg.com

That commit is from master, not this pr. No idea what the buildserver is doing, but it does not appear to be building prs

Julusian:
You can try this build: https://builds.julusian.dev/casparcg/casparcg-server-v2.3.x-custom-resolutions.zip

This PR includes custom channel resolutions now too. I was intending to keep them separate, but they ended up being more interlinked than I was hoping

MauriceVeraart:
Ow that's just really nice to have the custom-res like this too.
Will test this out in more depth later.
Thanks for this

Julusian:
What gpu do you have?
I am hoping that threshold is the max texture size it supports, which should make it easy to detect and refuse to run instead

MauriceVeraart:
quadro RTX 6000

Julusian:
Hmm.. 
That card has a `GL_MAX_TEXTURE_SIZE` of `32768`, so those dimensions should fit fine. And `GL_MAX_TEXTURE_BUFFER_SIZE` is `134217728` which does have some headroom, but not masses. I dont see any other reported number that looks to be in the region of being a limit here https://opengl.gpuinfo.org/displayreport.php?id=3826

Perhaps its a power of 2 thing? 8192x4096x4 is 134217728, so even though textures dont have to be a power of 2, perhaps it is doing so inside the driver. Does it work with increasing the height up to 4096? or if you drop the height to 2048 can it be made twice as wide?

I am wondering if the leaking is some bad handling in caspar, of us making some assumptions of operations that can never fail, resulting in leaking of buffers or something. Not sure if its worth looking into properly.

MauriceVeraart:
i can go to 8192x8192 (with bad performance in the end but not the same fault.)
strange thing is that 12288 (multiple of 2048) does work again i guess it must be dividable by 2048 or 
maybe devidable by some other power of 2

Sidonai-1:
> You can try this build: https://builds.julusian.dev/casparcg/casparcg-server-v2.3.x-custom-resolutions.zip
> 
I'm going to test this build a bit, is it a 2.3.3 stable with just this new feature or does it include some of the latest fixes that have been happening? (up-down transitions comes to mind, for example)

Thanks

PeterAkakpo:
currently i am able to set video mode for Decklink consumer in the config file. how do i set video mode on the command line 
ADD 1-100 DECKLINK 1 embedded_audio low_latency 720p5000_video_mode
ADD 1-100 DECKLINK 1 embedded_audio low_latency video_mode 720p5000
none of the above commands sets the video mode.

Julusian:
Nothing has been added to AMCP for this, as that was not necessary for the use case I had.

PeterAkakpo:
thanks for the response. but i think once it is configurable in the config file, it should be available for ACMP too. this will enable setting different video mode on the fly.  my usse case is to add and remove decklink outputs on the fly with a click of a button

Spo-oky:
I was trying to use this feature with a 4k video and divided by 4 decklink's (Decklink duo 2, 4 outputs) to perform the 4k. All with low latency, but one or 2 decklink's get out of sync for small frames. This will be nice to do this 4k 3G.
If I check task manager node of CPU and GPU are in 100%. Is there a way that we can try to sync them manualy?

Sidonai-1:
> I was trying to use this feature with a 4k video and divided by 4 decklink's (Decklink duo 2, 4 outputs) to perform the 4k. All with low latency, but one or 2 decklink's get out of sync for small frames. This will be nice to do this 4k 3G.
> If I check task manager node of CPU and GPU are in 100%. Is there a way that we can try to sync them manualy?

I'm in a similar situation.

 In my case I'm using a 8KPro to output four 4K signals. I set the buffer to 1, as I've been told that higher buffers mean opening the door to higher delays. CPU and GPU usage is around 50%, so no dropped or late frames. The diag window looks smooth. I tried "low latency" on the decklinks but couldn't see much of a difference so I stopped using it.

However, there is between 1 and 3 frames of delay between certain signals. It changes which one is it every time you restart Caspar. Sometimes I'm lucky and there is only 1 frame of delay in 1 signal. I couldn't pinpoint the cause of this, it seems a bit random.

I guess setting the buffer to 0 is not an option, right? 🤣


Thanks!!

Julusian:
This is interesting to hear. In my testing I never managed to get it to go out of sync, but you are not the first case where it has had issues. It sounds like this is a bit broken currently then.

@Sidonai-1 The buffer depth is how many frames we aim to have buffered in the decklink driver. The lower it is the lower the output latency, but also the more likely that the decklink will have to repeat a frame as we didnt provide more in time.

Sidonai-1:
> In my testing I never managed to get it to go out of sync
I tested a panoramic video mode at 30fps, and most of the time it looked just fine. It was only when the content was moving faster that it showed tearing effects. If you sew up two signals side by side it is specially noticeable with clear solids on darker background moving quickly upwards or downwards.

PR #1437, opened 2022-08-31
ARM64 compatibility

filiphanes:


petterreinholdtsen:
A set of build logs for arm64 can be found at https://buildd.debian.org/status/logs.php?pkg=casparcg-server&arch=arm64 .

dimitry-ishenko:
I'd say replace them with built-in types. It's only two extra chars `_t`. And if you feel adventurous, I've written my own OSC library, which is less abandoned: 😄 

https://github.com/dimitry-ishenko-cpp/liboscpp

dimitry-ishenko:
> I'd say replace them with built-in types. It's only two extra chars _t.

Actually, I take it back. Per the standard, you are really supposed to write `std::uint32_t` instead of just `uint32_t`:

https://en.cppreference.com/w/cpp/header/cstdint#Synopsis

so using aliases saves 7 chars.

Issue #1451, opened 2022-12-06
Seeking some formats cause CasparCG server to crash


Bohoaush:
### Expected behaviour
Seek video and start playing

### Current behaviour
Crash of CasparCG server - "Aborted (core dumped)" full log bellow

----

### Steps to reproduce
1. Download file seekTOcrash.mpg (bellow) to your media folder
2. Start CasparCG server
3. Send command `PLAY 1-1 seekTOcrash SEEK 20`

### Environment
* Commit: [eacaf19](https://github.com/CasparCG/server/commit/eacaf194d2f22372784e068f77d16edb8d4ebf95) , [f8e48f5](https://github.com/CasparCG/server/commit/f8e48f5190aabd54b94b5fa0415df33a9cb71441) (I haven't tested this with other builds)
* Server version: v2.3.3
* Operating system: Ubuntu 20.04.5

---

### Additional information & resources
* [seekTOcrash.mpg](https://soubory.tvnoe.cz/s/seekTOcrash) - test file that crashes CasparCG server on SEEK
* Log:
```
############################################################################
[2022-12-06 09:27:50.181] [info]    Starting CasparCG Video and Graphics Playout Server 2.3.3 N/A Dev
[2022-12-06 09:27:50.403] [info]    Initializing OpenGL Device.
[2022-12-06 09:27:50.407] [info]    Initialized OpenGL 4.5.0 NVIDIA 525.60.11 NVIDIA Corporation
[2022-12-06 09:27:50.407] [info]    Initialized image module.
[2022-12-06 09:27:50.407] [info]    Initialized ffmpeg module.
[2022-12-06 09:27:50.407] [info]    Initialized oal module.
[2022-12-06 09:27:50.407] [info]    Initialized decklink module.
[2022-12-06 09:27:50.407] [info]    Initialized screen module.
[2022-12-06 09:27:50.407] [info]    Initialized newtek module.
[2022-12-06 09:27:50.560] [info]    Initialized html module.
[2022-12-06 09:27:50.560] [info]    "/home/plus2/caspar/server/casparcg_server/casparcg.config":
[2022-12-06 09:27:50.560] [info]    -----------------------------------------
[2022-12-06 09:27:50.560] [info]    <?xml version="1.0" encoding="utf-8"?>
[2022-12-06 09:27:50.560] [info]    <configuration>
[2022-12-06 09:27:50.560] [info]       <paths>
[2022-12-06 09:27:50.560] [info]          <media-path>media/</media-path>
[...]

PR #1452, opened 2022-12-07
fix(ffmpeg_producer): crash on seek with some formats

Bohoaush:
With some formats, CasparCG would crash when attempting to seek. #1451

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -153,11 +153,7 @@ class Decoder
                             packet = std::move(input.front());
                             input.pop();
                         }
-                        try {
-                            FF(avcodec_send_packet(ctx.get(), packet.get()));
-                        } catch (ffmpeg::ffmpeg_error_t) {
-                            // Do nothing...
-                        }
+                        avcodec_send_packet(ctx.get(), packet.get());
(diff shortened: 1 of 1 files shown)


Julusian:
I don't know enough about ffmpeg apis to be comfortable merging this without knowing why this is erroring, and the potential impact of ignoring these errors for other formats. I don't think it is a good idea to ignore every possible error that ffmpeg may return.

Also, with this change, you are catching an error which is thrown by the `FF(` wrapper checking the return type. If it is merged, it should be simplified to not be throwing, and immediately catching and discarding the error

Bohoaush:
Thank you very much for prompt reaction and constructive comment.
Now I have "unwrapped" the lib call from FF instead of catching the error it throws. 
I don't understand ffmpeg libraries either, but I've spent hours digging through ffmpeg trac tickets, and it looks like ffmpeg throws this error because of receiving an incomplete packet. I can try to figure out a way to validate the packet before sending it, but it is not easy for me (I'm clearly not a programmer).
The ffmpeg and ffplay apps themselves appear to be ignoring this error as well.

Bohoaush:
The problem this was aiming to fix has been fixed by [c676f8f](https://github.com/CasparCG/server/commit/c676f8f7701e45741f137dd3fdcc1f0132fa31fb)

Issue #1460, opened 2023-01-24
CASPARCG CRASHES WHEN PLAYING SOME FILE THAT  FFPLAY IS ABLE TO PLAY


PeterAkakpo:
### Expected behaviour
Casparcg should be able to play all file that FFplay is capable of playing

### Current behaviour
Casparcg crashes while playing some files.

----

### Steps to reproduce
1.  play the attached file PLAY 1-0 KEYS-008
2.
3. ...

### Environment
* Commit: LATEST
* Server version: 2.3.3
* Operating system: WINDOWS 11

---

### Screenshots
If applicable, add screenshots as complementary information.
[KEYS-008.zip](https://github.com/CasparCG/server/files/10493873/KEYS-008.zip)

jpc0:
There is an error thrown in FFPLAY as well, it then get's stuck in an infinite loop.

```
[mov,mp4,m4a,3gp,3g2,mj2 @ 000001c31ffac4c0] Packet corrupt (stream = 0, dts = 299000).
[NULL @ 000001c31ffbde00] Invalid NAL unit size (397226 > 43604).
[NULL @ 000001c31ffbde00] missing picture in access unit with size 43652
```

I get the same error in CasparCG

```
[2023-03-18 17:38:08.232] [error]   [ffmpeg] [h264 @ 000001E3FF410200] Invalid NAL unit size (397226 > 43604).
[2023-03-18 17:38:08.232] [error]
[2023-03-18 17:38:08.232] [error]   [ffmpeg] [h264 @ 000001E3FF410200] Error splitting the input into NAL units.
```

Which then causes the thread to throw, casparCG however does not crash.

The error thrown by ffmpeg `av_read_frame` in `av_input.cpp` is `AVERROR_INVALIDDATA`.

Not sure this is a bug since the input file is corrupted.

Julusian:
Is this any better now?
A fix went in the other week that resolved a crash in this area which was a pretty generic fix

Issue #1444, opened 2022-11-22
Partial file crash


baoizbak:
### Expected behaviour
Caspar to stop the current consumer or load the background clip if it's there. Older version of caspar just stop playing the clip (Froze). Something like this we can detect and tell caspar to load a new clip. There are unforseen circumstance that we failed to check for partial files. 

### Current behaviour
When playing a partial file, caspar will crash when it reached to the end.

----

### Steps to reproduce
1. Play any partial file (incomplete download)
2.
3. ...

### Environment
* Commit: 670bec33 Dev
* Server version: Server 2.3.3
* Operating system: [e.g. Windows 10]

---

### Screenshots
[2022-09-12 13:19:32.461] [info]    Received message from Console: play 1-10 partial\r\n
[2022-09-12 13:19:32.744] [warning] ffmpeg[partial.mp4|0.0000/706.4724] Latency: 4
[2022-09-12 13:19:40.131] [info]    Received message from Console: add 1 screen\r\n
[2022-09-12 13:19:40.134] [info]    Screen consumer [1|NTSC] Initialized.
[2022-09-12 13:21:10.667] [info]    Received message from Console: loadbg 1-10 partial\r\n
[2022-09-12 13:21:11.183] [error]   [ffmpeg] [mov,mp4,m4a,3gp,3g2,mj2 @ 000001D0C27C7A00] stream 1, offset 0xab3855f: partial file
[2022-09-12 13:21:11.183] [error]   
[2022-09-12 13:21:11.184] [error]   Exception: C:\Program Files (x86)\Jenkins\workspace\casparcg-server-dep\master\src\modules\ffmpeg\producer\av_input.cpp(64): Throw in function auto __cdecl caspar::ffmpeg::Input::{ctor}::<lambda_fb20fb3672206abbfd8e24b0901bbc5b>::operator ()(void) const
[2022-09-12 13:21:11.184] [error]   Dynamic exception type: class boost::exception_detail::clone_impl<struct caspar::ffmpeg::ffmpeg_error_t>
[2022-09-12 13:21:11.184] [error]   [struct boost::errinfo_api_function_ * __ptr64] = av_read_frame
[2022-09-12 13:21:11.184] [error]   [struct boost::errinfo_errno_ * __ptr64] = 1094995529, "Unknown error"
[2022-09-12 13:21:11.184] [error]   [struct caspar::ffmpeg::tag_ffmpeg_errn_info * __ptr64] = -1094995529
[...]

PR #1469, opened 2023-03-19
Fix for crash on seek

jpc0:
Partial fix for https://github.com/CasparCG/server/issues/1451 and https://github.com/CasparCG/server/issues/1460

It is ugly but it does fix the crashing for the time being, however the audio thread keeps running as well as the main ffmpeg_producer thread.

I'm not familiar with boost::thread or boost::exception so maybe someone know how kill the other two threads when this thread throws. There may be an architectural issue here that needs to be resolved.


jpc0:
Second commit seems to allow the other two threads to cleanly end. Still not sure if this is the correct solution. I also suspect this case will never happen
```
catch (ffmpeg::ffmpeg_error_t& ex) {
                if (auto errn = boost::get_error_info<ffmpeg_errn_info>(ex)) {
                    if (*errn == AVERROR_EXIT) {
                        return;
                    }
                }
                CASPAR_LOG_CURRENT_EXCEPTION();
```

jpc0:
Thinking about this code a bit more, if feels like it is pointless to keep in, also move the eof=true into the catch.

Let me know if this should be squashed into 1 commit.

````
catch (ffmpeg::ffmpeg_error_t& ex) {
            if (auto errn = boost::get_error_info<ffmpeg_errn_info>(ex)) {
                if (*errn == AVERROR_EXIT) {
                    return;
                }
            }
            CASPAR_LOG_CURRENT_EXCEPTION();
```

jpc0:
Looking at https://github.com/CasparCG/server/issues/1460 again...

This would prevent the crash but would also end the video playback. Adding a new try catch in the loop would prevent that, the question is whether audio sync would then go out.

V3.3 doesn't crash but won't play this file.

jpc0:
https://github.com/CasparCG/server/issues/1444 is also solved by this.

Would be great for @ronag to give some feedback here. https://github.com/CasparCG/server/issues/1444 https://github.com/CasparCG/server/issues/1451 https://github.com/CasparCG/server/issues/1460 are all related to the changes in commit [b657428](https://github.com/CasparCG/server/commit/b657428db652a2ef43601fc3923934498506a916).

Before that commit there was a single thread where any exception would kill the pipeline and be caught at the .run call [here](https://github.com/CasparCG/server/blob/5dcd50c53c4df4ecc1fe1ca3a53c5eee3e2f3afe/src/modules/ffmpeg/producer/av_producer.cpp#L587). commit [b657428](https://github.com/CasparCG/server/commit/b657428db652a2ef43601fc3923934498506a916) Now spawns 2 additional worker threads, one for video and another for audio and I'm not entirely sure how to stop one of the workers if the other throws other that setting the eof flag.

For the seek issue we would want play to continue theoretically however for partial files we would want the threads to stop processing more data.

With the try catch in the loop the video thread will stop but audio continues until a natural EOF is reached if ever, this could just end up stuck in a loop infinitely. But if we catch outside the loop seeking doesn't work.

If we want to just error on a file that is broken and not attempt to recover then breaking outside the loop is fine.

jpc0:
Should we be catching `catch (ffmpeg::ffmpeg_error_t& ex) {` here instead of all since we really only are looking for the ffmpeg errors?

Bohoaush:
Hi @jpc0 thank you for looking into this. 
Few days ago I tried commit 8a4befd (which now isn't in commit history) and it worked well for me. 
Yesterday i built from commit 712b006 (currently last one) and it stays on last "good" frame, but continuing according to info command. This is definitely better than crashing, but I think continuing playback is better behavior. If incomplete files are a concern, I don't think they are more common than damaged files, but I might be wrong. 
I have hundreds of old non-damaged files that would crash current master of CasparCG server if seeking and at least one file that crashes it during normal playback, because it has damaged macro block.

jpc0:
> Hi @jpc0 thank you for looking into this. 
> 
> Few days ago I tried commit 8a4befd (which now isn't in commit history) and it worked well for me. 
> 
> Yesterday i built from commit 712b006 (currently last one) and it stays on last "good" frame, but continuing according to info command. This is definitely better than crashing, but I think continuing playback is better behavior. If incomplete files are a concern, I don't think they are more common than damaged files, but I might be wrong. 
> 
> I have hundreds of old non-damaged files that would crash current master of CasparCG server if seeking and at least one file that crashes it during normal playback, because it has damaged macro block.

[](https://github.com/CasparCG/server/commit/4f37f1be63704d24c69a3f87842f7be7c33086d0)

Should be what you are looking for.

I rewrote history when my fix for VS2022 got pulled.

I left the two in history here though because I need maintainers to know which direction to go in.

The issue with continueing on error is that the loop hangs indefinitely. So will likely need to pull some deeper errors out of FFMPEG and catch two seperate errors.

Issue #1470, opened 2023-03-20
Build fails in VS 17 2022


jpc0:
### Expected behaviour
Build works

### Current behaviour
```
\modules\bluefish\consumer\bluefish_consumer.cpp(892,16): error C2220: the following warning is treated as an error
\src\modules\bluefish\consumer\bluefish_consumer.cpp(892,16): warning C4189: 'embedded_audio': local variable is initialized but not referenced
\src\modules\bluefish\consumer\bluefish_consumer.cpp(891,16): warning C4189: 'device_stream': local variable is initialized but not referenced
\src\modules\bluefish\consumer\bluefish_consumer.cpp(890,16): warning C4189: 'device_index': local variable is initialized but not referenced
\src\modules\bluefish\consumer\bluefish_consumer.cpp(894,16): warning C4189: 'keyer_audio_option': local variable is initialized but not referenced
\src\modules\bluefish\consumer\bluefish_consumer.cpp(893,16): warning C4189: 'keyer_option': local variable is initialized but not referenced
```
----

### Environment
* Commit: be5f396
* Operating system: Windows 11

---

PR #1471, opened 2023-03-20
fix: Build failure with VS 2022

jpc0:
Fix for https://github.com/CasparCG/server/issues/1470

Added praga ignore for warnings with conditional for MSVC


Julusian:
could you comment out the offending lines instead? Seeing as they aren't used they may as well be disabled instead of suppressing the warning

PR #1474, opened 2023-04-13
Update decklink_producer.cpp

maxmuzy:
When decklink is without reference it produces an exaggerated amount of log messages generating around 700MB of log per day. I don't see any harm in having it set to "trace", since when activated it makes any possibility of analyzing the log due to excessive messages unfeasible.


Julusian:
I have been meaning to look at this for a while, but haven't been able to test it.

I thought it was odd that this was happening, as I know of places where decklink inputs are used without any complains on the logging. I think this only happened for some video formats such as 59.94 and not 50, so it isnt seen by everyone.

And the reason for the logging is a bug in how it was generated. I have fixed that, so it should now no longer flood the logs for you, even without changing the level of the line. 2056600fa50f045399dd8429eba9ac61e2d378d5

Thanks for the PR, it was the nudge needed to check what the problem was :)

PR #1479, opened 2023-05-24
Set casparcg project as the default

jpc0:
In VS the default project currently is ALL_BUILD which in fact will not build. This will set the current project to casparcg which is the project we would want built when hitting build solution.

Not sure if this can go withing the Boostrap Windows script, the cmake documentation seems to indicate it must go in the same directory as a project call. [https://cmake.org/cmake/help/latest/prop_dir/VS_STARTUP_PROJECT.html](https://cmake.org/cmake/help/latest/prop_dir/VS_STARTUP_PROJECT.html)


Julusian:
could you try putting it into the bootstrap_windows file?
I suspect it will work, as it is doing an 'include' not an 'add_subdirectory' call to load the next file

jpc0:
You were correct, it does work. Have made the change

Issue #1480, opened 2023-05-24
Bump boost version to 1.67


jpc0:
### Description
As mention in [https://github.com/CasparCG/server/issues/1478#issuecomment-1560974249](https://github.com/CasparCG/server/issues/1478#issuecomment-1560974249) the only blocker at present to bumping the language minimum version to C++17 if we so want to is boost version 1.66 using `std::auto_ptr` in the locale library. This was changed in boost 1.67

### Solution suggestion
Bump the minimum boost version required to 1.67. The only breaking change that may possibly affect us is:

[https://www.boost.org/users/history/version_1_67_0.html](https://www.boost.org/users/history/version_1_67_0.html)
> [Atomic](https://www.boost.org/libs/atomic/):
>
>    Breaking change: Changed the result of the (op)_and_test operations added in Boost 1.66 to the opposite - the functions now return true if the operation result is non-zero. This is consistent with other test methods in Boost.Atomic and the C++ standard library. Users can define BOOST_ATOMIC_DETAIL_HIGHLIGHT_OP_AND_TEST when compiling their code to emit warnings on every use of the changed functions. This way users can locate the code that needs to be updated. ([#11](https://github.com/boostorg/atomic/issues/11)) 

however building with `BOOST_ATOMIC_DETAIL_HIGHLIGHT_OP_AND_TEST` defined does not seem to throw any warnings so I think we are good.

---

### Additional information
**IMPORANT** Question: Where is [https://github.com/CasparCG/server/blob/6483b20e1368b7f8727ebc06331734bb3e791cfd/tools/linux/image-versions#L7](https://github.com/CasparCG/server/blob/6483b20e1368b7f8727ebc06331734bb3e791cfd/tools/linux/image-versions#L7) generated, this is the only place I am not sure about where to update the value. Will just bumping it to `export IMAGE_BOOST=${IMAGE_PREFIX}boost:1.67-8` do the trick?

PR #1481, opened 2023-05-24
Bumped boost to version 1.67

jpc0:
See [https://github.com/CasparCG/server/issues/1480](https://github.com/CasparCG/server/issues/1480).


tools/linux/image-versions
```diff
@@ -4,5 +4,5 @@ export IMAGE_PREFIX=ghcr.io/casparcg/server/
 
 export IMAGE_BASE=${IMAGE_PREFIX}base:focal-2
 export IMAGE_CEF=${IMAGE_PREFIX}cef:95.0.4638.69-obs-1
-export IMAGE_BOOST=${IMAGE_PREFIX}boost:1.66-8
+export IMAGE_BOOST=${IMAGE_PREFIX}boost:1.67-8
```
Julusian, line 7:
```suggestion
export IMAGE_BOOST=${IMAGE_PREFIX}boost:1.67-0
```

Julusian:
We could probably get away with using 1.71, but I am happy with 1.67.
I say probably because while oldstable debian (buster) uses 1.67, the next oldest uses 1.71. I dont see why anyone would want to try and run this on an os which is EOL in a year. Besides, they would probably be able to perhaps with some slight import tweaks to get it to compile. 
But as there is no reason to go newer yet, there probably isnt any point in doing so.

jpc0:
I didn't check if there would be any breaking changes up until 1.71. I figured since 1.67 fixed being able to go to bump the language version to C++17 if we ever need to there wasn't much reason to go above that.

I suppose if someone needs something from a newer boost version or if one of the bugs effects us then someone can look into whether there are any breaking changes that are an issue.

PR #1486, opened 2023-05-27
Move from custom Precompiled header implementation to Cmake native

jpc0:
This include pull request [https://github.com/CasparCG/server/pull/1485](https://github.com/CasparCG/server/pull/1485). If this gets commited the other does not need to be committed.


Julusian:
I think I need to do a bit of testing of this one before merging. I wonder what the experience is like to edit the precompiled header lists, particularly on windows. I suspect this will be just as awkward as adding new files, which would be acceptable.

jpc0:
Yeah it is a bit of a beast. I also am not currently using that workflow but would want to test things in an IDE on linux like clion or something if someone can do that. Not sure how that integrates with cmake PCH.

It's actually a pretty big change...

jpc0:
Up to you but I think the other one should be merged and not this one. I think that one follows convention better.

jpc0:
Going to close the PR since the other seems to be the preferred.

PR #1487, opened 2023-05-27
Alternative to the other CMake precompiled header

jpc0:
This is an alternative to the way I did it before, this one keeps the StdAfx headers.

Not sure if the StdAfx.cpp/StdAfx.c files would still be needed and should still be included in the build. Will check that but this version builds right now.


jpc0:
That linux build failing doesn't seem related to my code?

Julusian:
> That linux build failing doesn't seem related to my code?

Yeah, thats my fault

Julusian:
Yeah I prefer this option

In this you have moved some files to `src/common/include/common`, is there a reason for that? its making for some weird import statements.

Other than that I think this is good

src/modules/bluefish/StdAfx.h
```diff
@@ -37,12 +37,12 @@
 #include <string>
 #include <vector>
 
-#include <common/memory.h>
-#include <common/utf.h>
+#include "../common/memory.h"
```
Julusian, line 40: Is there a reason for changing these imports from common?

src/modules/newtek/StdAfx.h
```diff
@@ -35,9 +35,9 @@
 #include <string>
 #include <vector>
 
-#include <common/memory.h>
+#include "../common/memory.h"
 
-#include <common/except.h>
-#include <common/log.h>
+#include "../common/except.h"
```
Julusian, line 40: same question for these

src/protocol/StdAfx.h
```diff
@@ -51,10 +51,10 @@
 #include <boost/filesystem.hpp>
 #include <boost/range/algorithm.hpp>
 
-#include <common/memory.h>
-#include <common/utf.h>
+#include "../common/memory.h"
+#include "../common/utf.h"
```
Julusian, line 55: and these

jpc0:
Guess not. Let me confirm the rest of the StdAfx.g files build if I change to `<common/...>` then will update.

dimitry-ishenko:
Hate to rain on your guys' parade, but I've always been taught that angle brackets are used for system headers.

Strictly speaking, the standard says that the difference between `#include <...>` and `#include "..."` is implementation-defined, but the common consensus seems to be:

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

Julusian:
I don't disagree with that, but this is consistent with what we are already doing. I'm happy for that to be improved, but it felt odd for just a few of those places to be changed in an 'unrelated' PR

dimitry-ishenko:
Makes sense. I guess this would be a fight for another day.

jpc0:
I'm with @Julusian on this one, core guidelines agree as well [https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform) which you can see is what I was following but since it worked without the relative includes I didn't really have an opinion either way. I would much rather not include `..` in the cmake file to start with and rather export the includes explicitly where needed in the target but that wouldn't change how the includes are written.

dimitry-ishenko:
@jpc0 that link is saying the same thing I am saying (or, rather, the gcc docs). 😛 I agree with you on not using `..` though.

jpc0:
From what I'm reading it says to use <> for everything except relative imports...

But yeah with some of the cmake clean-up we can get rid of the relative include path, just need to convince @Julusian on the weird include directory or find a different approach, I generally think that the includes should be namespaced though.

dimitry-ishenko:
@jpc0 from your Core Guidelines link (emphasis mine):

> Nevertheless, the guidance is to use the quoted form for including files that exist at a relative path to the file containing the #include statement (_**from within the same component or project**_) and to use the angle bracket form everywhere else, where possible.

In other words, files from your project, which are all located at some path relative to the current file, should use quotes. All others, meaning system files and other libs not part of your project, should use angle brackets.

They are saying the same thing the gcc link is saying.

PR #1491, opened 2023-07-19
Add Artnet consumer

firecraftgaming:
I added a DMX consumer that is capable of taking a region of a channel and outputing that onto dmx channels in different forms. It is capable of outputing three modes. These modes are DIMMER, RGB, RGBW, which will average the color in the region and change its colorspace to then output 1, 3, 4 channels on the configured artnet universe starting at the configured start address. 

I also created a UDP Heartbeat service that lets you configure a name for a caspar server, and it will broadcast that name, letting people on the network easier find caspar servers. Could create for a system where clients can automatically identify servers without the need to specify an ip. This is really nice for headless servers that get moved around alot. Not needing the pain of figuring out its ip address. This is an optional settings and is turned off by default. The destination and send interval can also be configured, it defaults to send to the networks broadcast ip over port 3468 every second.


jpc0:
> I added a DMX consumer that is capable of taking a region of a channel and outputing that onto dmx channels in different forms. It is capable of outputing three modes. These modes are DIMMER, RGB, RGBW, which will average the color in the region and change its colorspace to then output 1, 3, 4 channels on the configured artnet universe starting at the configured start address.
 
I don't necessarily see where DMX being built in to CasparCG is useful, this would be up to one of the other's to comment on. Probably more useful to implement it independently with an API that can be tied into existing systems.

> I also created a UDP Heartbeat service that lets you configure a name for a caspar server, and it will broadcast that name, letting people on the network easier find caspar servers. Could create for a system where clients can automatically identify servers without the need to specify an ip. This is really nice for headless servers that get moved around alot. Not needing the pain of figuring out its ip address. This is an optional settings and is turned off by default. The destination and send interval can also be configured, it defaults to send to the networks broadcast ip over port 3468 every second.

I would like to query why this over an existing protocol such as mDNS? Having mDNS will likely be more useful in the future for implementing other standards such as Ravenna or ST2110.

firecraftgaming:
> I don't necessarily see where DMX being built in to CasparCG is useful, this would be up to one of the other's to comment on. Probably more useful to implement it independently with an API that can be tied into existing systems.

I think it is useful to have built in, as it would be really tedious to implement otherwise. I find it hard to see an api that would allow for the fine control needed to send dmx data from channels with great configuration.

On a sidenote: These features were implemented primarily for my own project, but it fealt as they would fit nicely in the official caspar version too, so I understand if some interests are unaligned.

> I would like to query why this over an existing protocol such as mDNS? Having mDNS will likely be more useful in the future for implementing other standards such as Ravenna or ST2110.

I suppose that would be nice, however for us it was a lot more useful to have it broadcasted as we are changing network often. We also need names to be included in the messages in an easy way, so it would be straight forward for our clients to pick up the names, and distinguish between our servers. As for the this repo it might indeed make more sense to implement it using mDNS.

vimlesh1975:
Can we get a build to test.On 20 Jul 2023 5:40 am, Eliyah Sundström ***@***.***> wrote:

I don't necessarily see where DMX being built in to CasparCG is useful, this would be up to one of the other's to comment on. Probably more useful to implement it independently with an API that can be tied into existing systems.

I think it is useful to have built in, as it would be really tedious to implement otherwise. I find it hard to see an api that would allow for the fine control needed to send dmx data from channels with great configuration.
On a sidenote: These features were implemented primarily for my own project, but it fealt as they would fit nicely in the official caspar version too, so I understand if some interests are unaligned.

I would like to query why this over an existing protocol such as mDNS? Having mDNS will likely be more useful in the future for implementing other standards such as Ravenna or ST2110.

I suppose that would be nice, however for us it was a lot more useful to have it broadcasted as we are changing network often. We also need names to be included in the messages in an easy way, so it would be straight forward for our clients to pick up the names, and distinguish between our servers. As for the this repo it might indeed make more sense to implement it using mDNS.

dimitry-ishenko:
Since OSC seems to be the go-to protocol for CCG, I also kind of agree that DMX should implemented externally. I've written and [OSC Gateway](https://github.com/dimitry-ishenko-casparcg/oscgard) which you can use to translate OSC messages to DMX for example.

firecraftgaming:
> Since OSC seems to be the go-to protocol for CCG, I also kind of agree that DMX should implemented externally. I've written and [OSC Gateway](https://github.com/dimitry-ishenko-casparcg/oscgard) which you can use to translate OSC messages to DMX for example.

I am sorry, but I don't really understand. How would one get pixel data from osc messages? If it would help to understand, the dmx consumer is inspired by Resolume arenas dmx output. However a bit simplified, but can be extended on later. https://resolume.com/support/en/dmx

jpc0:
> > Since OSC seems to be the go-to protocol for CCG, I also kind of agree that DMX should implemented externally. I've written and [OSC Gateway](https://github.com/dimitry-ishenko-casparcg/oscgard) which you can use to translate OSC messages to DMX for example.
> 
> I am sorry, but I don't really understand. How would one get pixel data from osc messages? If it would help to understand, the dmx consumer is inspired by Resolume arenas dmx output. However a bit simplified, but can be extended on later. https://resolume.com/support/en/dmx

I don't think either of us looked at the actual code before commenting, or at least not in depth since I only started the discussion on the meta idea not on any specific implementation details. Something that was left out of the pull request is that it does seem like you would be actually attempting to output the image from the channel onto the DMX controlled lighting or at least doing something like lightsync where the colour of the lighting matches the on screen visuals.

I know my use case is actually live production and at least one other company also uses Caspar for that but it is targeted as a CG and playout device. Likewise in our live production that functionality is covered by our lighting console with all major brands supporting pixel mapping.

> ... something like lightsync where the colour of the lighting matches the on screen visuals.

Might be interesting in a studio setting but generally the lighting isn't controlled from the playout server. Maybe an API can be exposed from CasparCG, possibly using OSC as mentioned that can be used to pass the information.

This is a rather significant piece of code to add to Caspar though and will likely need proper testing.

firecraftgaming:
> ... it does seem like you would be actually attempting to output the image from the channel onto the DMX controlled lighting or at least doing something like lightsync where the colour of the lighting matches the on screen visuals.

Yeah that is kind of exactly what I am trying to do.

> ...  generally the lighting isn't controlled from the playout server.

Yeah I know that is the usual case, but in our use case we have a lightning console that controls the regular lights and also sends one dmx signal to our caspar server to colorize the current background graphics on our screen. The screen however doesn't cover the entire stage so we wanted to extend it with some led strips, but not overcomplicate it with a lot of expensive equipment or fancy controllers. We did not just want to have the led strips be the same color that the caspar server receive but instead copy the outer pixels, so that if we overlay something on top of the background graphics they would also change. We felt that the easiest and cheapest option would be to build it into caspar, so that is what we did. I thought that would be a neat feature to have built into the official caspar version too, for people wanting the same thing as we did.

> ... and will likely need proper testing.

Yeah, I bet. Though I have tested the fixture config a lot, and the rasterisation too with all sort of config. I firstly tested it on a javascript canvas with directly translated code so I could see exactly what pixels it selected and so on. I also of course tested in caspar with different colored backgrounds. However I think it rightly would need some automated tests, to ensure it functions in the future too.

dimitry-ishenko:
> I am sorry, but I don't really understand. How would one get pixel data from osc messages? If it would help to understand, the dmx consumer is inspired by Resolume arenas dmx output. However a bit simplified, but can be extended on later. https://resolume.com/support/en/dmx

You don't. Disregards my comment. 😅 I wrote it late at night, while thinking about something else.

firecraftgaming:
> ... I firstly tested it on a javascript canvas with directly translated code so I could see exactly what pixels it selected and so on.

Yeah I did, but I just realised now when I was about to share the demo that one edge case doesn't work, so yeah testing is needed lol. However I think except that edge it should work fine on every config.

Julusian:
As this is an alternate video output 'device', I don't have a problem with it being included. It's not making any changes outside of adding a new module, so I don't see it as adding any maintenance cost. 

>  I find it hard to see an api that would allow for the fine control needed to send dmx data from channels with great configuration.

This would be possible, using something like NDI in the middle. CasparCG already supports that, so you would need a simple application to receive the NDI and transform it into artnet.

> I suppose that would be nice, however for us it was a lot more useful to have it broadcasted as we are changing network often.

All of that is part of mDNS too. It allows for connecting via the hostname (if you trust mDNS to be reliable enough for that), or it can be used for discovery where CasparCG could provide a name, ip, port and any other key+value data that would be useful.

> but it is targeted as a CG and playout device

It is, but some functionality (mostly more recent features) are more live-production based than strictly broadcast (eg https://github.com/CasparCG/server/pull/1433)

firecraftgaming:
> This would be possible, using something like NDI in the middle. CasparCG already supports that, so you would need a simple application to receive the NDI and transform it into artnet.

Yeah, I suppose, but NDI is often unreliable and really cpu intensive, so for us it fealt better to implement it as a module.

> All of that is part of mDNS too. It allows for connecting via the hostname (if you trust mDNS to be reliable enough for that), or it can be used for discovery where CasparCG could provide a name, ip, port and any other key+value data that would be useful.

Well, I am going to look into that and implement it then. I am going to create a commit now that removes the heartbeat so this pr can be merged. I'll create another pr later with mDNS implemented, so they can be merged independently.

firecraftgaming:
I have now removed the heartbeat part of the pr, and from my testing the dmx consumer is operating great.

src/modules/dmx/consumer/dmx_consumer.cpp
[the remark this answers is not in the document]
firecraftgaming: Isn't it better to use something other than `concurrent_bounded_queue`. I am opting for the first solution and passing the const frame to the calculation thread. However that thread will only pop the frame everytime it needs to send data, and that would usually be lower than the fps of the channel. It's probably better to just store the last frame?

firecraftgaming:
> Unrelated, should it be possible to specify the local ip address to open the socket on, for scenarios where the machine is connected to multiple networks?

I don't think this is necessary as the artnet data are outgoing UDP packets. The different networks have different local ip ranges right? Like a device on network A would have the ip say 192.168.xxx.yyy and a device on network b would have the ip 172.22.xxx.yyy. I am not sure, so please correct me if I am wrong.

src/modules/dmx/consumer/dmx_consumer.cpp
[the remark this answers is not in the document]
firecraftgaming: I used a variable to store the last variable and a mutex to access it. I don't know if that was the right approach. Do you have a better idea how to implement it, or is it okay?

jpc0:
> > Unrelated, should it be possible to specify the local ip address to open the socket on, for scenarios where the machine is connected to multiple networks?
> 
> 
> 
> I don't think this is necessary as the artnet data are outgoing UDP packets. The different networks have different local ip ranges right? Like a device on network A would have the ip say 192.168.xxx.yyy and a device on network b would have the ip 172.22.xxx.yyy. I am not sure, so please correct me if I am wrong.

If artnet allows multicast then I think it would be needed. This would definitely be needed for the mDNS implementation since most network libraries I've used uses device IP to bind to specific interfaces.

src/modules/artnet/consumer/artnet_consumer.cpp
```diff
@@ -0,0 +1,291 @@
...
+                                uint8_t dmx_data[512];
+                                memset(dmx_data, 0, 512);
+
+                                for (auto computed_fixture : computed_fixtures) {
+                                    auto color = average_color(frame, computed_fixture.rectangle);
+                                    uint8_t* ptr = dmx_data + computed_fixture.address;
+
+                                    switch (computed_fixture.type) {
+                                        case FixtureType::DIMMER:
+                                            ptr[0] = (color.r + color.g + color.b) / 3;
```
baltedewit, line 105: I believe this implementation to be slightly naive. It would be better to convert to luminance.
```suggestion
                                            ptr[0] =  0.279 * color.r + 0.547 * color.g + 0.106 * color.b
```

src/shell/casparcg.config
```diff
@@ -135,6 +138,30 @@
...
+                <host>127.0.0.1</host>
+                <port>6454</port>
+
+                <refresh-rate>30</refresh-rate>
+
+                <fixtures>
+                    <fixture>
+                        <type>RGB</type>
+                        <start-address>1</start-address>
+                        <fixture-count>10</fixture-count>
```
baltedewit, line 153: I don't know _too much_ about lighting, but would it be useful to have some stride parameters here? If I'm not mistaken lots of fixtures have have other parameters besides RGBW that we don't want to control from CasparCG.

baltedewit (commented):
I think it's okay for this live as a CasparCG consumer too. I do question how useful this implementation will be to anyone else but I can't comment too much on that with my limited knowledge of lighting applications.

src/shell/casparcg.config
firecraftgaming: Yeah, you are right. I am going to fix a stride setting to create an offset between fixtures.

firecraftgaming:
> I don't know too much about lighting, but would it be useful to have some stride parameters here? If I'm not mistaken lots of fixtures have have other parameters besides RGBW that we don't want to control from CasparCG.

I have created a setting for how many channels per fixture, which should suffice for most applications. Though I have in plan to make it so you can define your own additional fixture types so that it can match the fixtures real channels better.

firecraftgaming:
> If artnet allows multicast then I think it would be needed. This would definitely be needed for the mDNS implementation since most network libraries I've used uses device IP to bind to specific interfaces.

I still think it should work by which ip you define, as different network interfaces also have different starts of multicast addresses. At least I hope so, as I can't find any resource on how it is possible in boost to bind to a certain network interface, especially when you are sending UDP packets, and not just receiving them.

jpc0:
> ... as I can't find any resource on how it is possible in boost to bind to a certain network interface

Boost ASIO most definitely does support this

[](https://www.boost.org/doc/libs/1_82_0/doc/html/boost_asio/reference/ip__multicast__outbound_interface.html)

Julusian:
I'm pretty certain artnet does do both unicast and multicast. This doesn't need to support them both though. That can be done later if someone wants it. 

I am wondering about the sending up because I remember that some (typically older?) artnet devices use a ip range 2.0.0.0/8, which is not a private range so if being used on a machine which also has Internet/other networks, it will likely need some help to know to use the correct interface. 
Again, this could be not supported for now

jpc0:
> I'm pretty certain artnet does do both unicast and multicast. This doesn't need to support them both though. That can be done later if someone wants it. 
> 
> I am wondering about the sending up because I remember that some (typically older?) artnet devices use a ip range 2.0.0.0/8, which is not a private range so if being used on a machine which also has Internet/other networks, it will likely need some help to know to use the correct interface. 
> Again, this could be not supported for now
> 

I agree with this. YAGNI

firecraftgaming:
I am now done with all the requested changes except the one about choosing a specific network interface as it seems like that can be done later in the future if needed. However I am getting a weird build message only on ubuntu 22.04: `error: ‘sleep_for’ is not a member of ‘std::this_thread’`, except it is and its building fine on the other platforms, and other consumers use it and it is working there. I don't have a lot of c++ knowledge so if someone could help me, it would be really appreciated.

jpc0:
> I am now done with all the requested changes except the one about choosing a specific network interface as it seems like that can be done later in the future if needed. However I am getting a weird build message only on ubuntu 22.04: `error: ‘sleep_for’ is not a member of ‘std::this_thread’`, except it is and its building fine on the other platforms, and other consumers use it and it is working there. I don't have a lot of c++ knowledge so if someone could help me, it would be really appreciated.

Just stopped at home so will check in an hour or so. Did you try the docker build?

Also this could be some sort of UB presenting itself, I see at least in the last commit there is a Cstyle cast, maybe changing those to static_cast will bring it to the forefront. As I said I haven't look very deeply into the code so that is a total shot in the dark

jpc0:
> I am now done with all the requested changes except the one about choosing a specific network interface as it seems like that can be done later in the future if needed. However I am getting a weird build message only on ubuntu 22.04: `error: ‘sleep_for’ is not a member of ‘std::this_thread’`, except it is and its building fine on the other platforms, and other consumers use it and it is working there. I don't have a lot of c++ knowledge so if someone could help me, it would be really appreciated.

You need to `#include <thread>`, do not rely on transitive includes, if you use something from a header import the header directly.

firecraftgaming:
Oh, yeah... I forgot that. I am going to fix it soon.

src/modules/artnet/artnet.h
```diff
@@ -0,0 +1,22 @@
...
+
+#pragma once
+
+#include <core/module_dependencies.h>
+
+namespace caspar {
+    namespace artnet {
+
+        void init(const core::module_dependencies& dependencies);
+        void uninit();
```
Julusian, line 18:
```suggestion
```
This method is not implemented

src/modules/artnet/consumer/artnet_consumer.cpp
```diff
@@ -0,0 +1,298 @@
...
+                                            ptr[1] = color.g - w;
+                                            ptr[2] = color.b - w;
+                                            ptr[3] = w;
+                                            break;
+                                    }
+                                }
+
+                                send_dmx_data(dmx_data, 512);
+                            }
+                        } catch (...) {
```
Julusian, line 125: should this be outside or inside of the while loop?

Currently, when something throws it will be caught outside of the while loop, causing the thread to terminate. The consumer will then be left not outputting any data, appearing to casparcg, AMCP and OSC to be still running.

This be moved inside of the while loop, so that it will continue to try to send data on the next frame. Or alternatively, if you store and rethrow the error from `send()`, then we will see that the consumer errored and will automatically remove it. (I think, I am working from memory of the decklink consumer)

Julusian (commented):
All of the files should probably include the GPL copyright block that we use elsewhere. It may not be strictly necessary, but these files being different will likely make someone unsure if these files are covered under gpl or if this header implies they are licensed differently

```
/*
 * Copyright (c) 2011 Sveriges Television AB <info@casparcg.com>
 *
 * This file is part of CasparCG (www.casparcg.com).
 *
 * CasparCG is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * CasparCG is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with CasparCG. If not, see <http://www.gnu.org/licenses/>.
 *
 * Author: Eliyah Sundström eliyah@sundstroem.com
 */
```

firecraftgaming:
> All of the files should probably include the GPL copyright block that we use elsewhere. It may not be strictly necessary, but these files being different will likely make someone unsure if these files are covered under gpl or if this header implies they are licensed differently

I have fixed it now.

Issue #1235, opened 2019-09-20
CEF Upgrade


Julusian:
### Description
CEF often has breaking bugs which block us from upgrading to certain versions. The purpose of this issue is to track those bugs so that we know why we can't use a version and what to verify is fixed when we do upgrade.

#### CEF 73
Nothing known

#### CEF 75
* HTML video doesnt trigger repaints, causing some pages (such as youtube) to not render smoothly/at all. (Might affect 74 too)

#### CEF 76
* [New Viz compositor](https://bitbucket.org/chromiumembedded/cef/issues/2575/viz-implementation-for-osr). Removes shared_texture functionality. [Issue](https://bitbucket.org/chromiumembedded/cef/issues/1006/cef3-support-rendering-to-a-hardware-gl)

Jgaleas7:
in the version 77??

sirfnomi:
I think shared texture still not back in 77

sirfnomi:
Hi Julusian, Is CasparCG updating with 73 till new Viz gets shared texture ?

Julusian:
Yes, my intention is for #1217 to update to CEF73. Hopefully that will happen sometime this month

sirfnomi:
Thanks hopefully.

Jgaleas7:
any update?

Julusian:
@Jgaleas7 master is now updated to CEF71, with shared texture support

dotarmin:
> @Jgaleas7 master is now updated to CEF71, with shared texture support

I'm sure the builds has stoped to work due to this. I can try to fix it this weekend or on Monday morning.

grahamspr86:
> > @Julusian  master is now updated to CEF71, with shared texture support
> 
> I'm sure the builds has stoped to work due to this. I can try to fix it this weekend or on Monday morning.

That explains why I couldn't see a build! Wasn't sure how long they took following the commit

dotarmin:
> That explains why I couldn't see a build! Wasn't sure how long they took following the commit

I just noticed it, I can maybe even try to fix it tonight if I'm free. It takes around 7 minutes for a build to finish.

Julusian:
@dotarmin It compiles in vs fine, so I guess it will be something with the packaging? The files needed to be copied from CEF has changed in this version so that could be it, but its hard to say without seeing the build output

dotarmin:
> The files needed to be copied from CEF has changed in this version so that could be it

[...]

Issue #1390, opened 2021-08-31
‘LOAD’ command 2.0.7 vs 2.3.x


Sidonai-1:
Hello all,

I recently noticed a difference in the behaviour of the LOAD command in version 2.0.7 vs 2.3.x and I am wondering if it has any reason behind it or maybe something changed and I need to use another command now.

In 2.0.7 if you had something playing/paused on a channel and you used “LOAD 1-10 Clip1.mp4” the first frame of Clip1 would appear in the channel inmediately, nothing in between the previous content and the new.

In 2.3.x the LOAD function ‘clears the layer’ before showing the first frame of Clip1, which brings the signal to black for at least 1 frame.

The PLAY command doesn’t behave like this, it does the same in 2.0.7 and 2.3.x

Does this have any reason to be or could we request to bring it back the way it was in 2.0.7?

Any thoughts on a workaround in the meantime?

Thank you!
----

### Environment
* Server version: 2.3.x
* Operating system: Windows 10

PR #1499, opened 2023-10-06
feat: Update CEF to 117

### LINKED ISSUE #1265: GPU memory leak on HTML producer
### Expected behaviour
VRAM usage should be freed upon closing of HTML templates or HTML producers.

### Current behaviour
GPU memory usage keeps increasing when removing and adding HTML templates until VRAM is full and starts using shared video memory, slowing down rendering.

----

### Steps to reproduce
1. Specify `<html><enable-gpu>true</enable-gpu></html>` in config
2. `CG 1-1 ADD 0 "template1" 1 "data"`
3. `CG 1-1 STOP`
3. Repeat steps 2 and 3 indefinitely.
4. Watch casparcg.exe GPU memory usage go up in task manager.

### Environment
* Commit: 6985aac9cd164dea5c1049272d005caace55b2c3
* Server version: 2.3 dev

---

### Screenshots
Timelapse: https://img.mehir.ar/template.gif

### LINKED ISSUE #1387: With "GPU enable" server don't start
### Expected behaviour
I guess start normally and play html + threejs templates

### Current behaviour
does't start, shows a log with some errors and at the end a line to report here this issue

`[2021-07-21 12:00:59.271] [info]    ############################################################################
[2021-07-21 12:00:59.272] [info]    CasparCG Server is distributed by the Swedish Broadcasting Corporation (SVT)
[2021-07-21 12:00:59.272] [info]    under the GNU General Public License GPLv3 or higher.
[2021-07-21 12:00:59.272] [info]    Please see LICENSE.TXT for details.
[2021-07-21 12:00:59.272] [info]    http://www.casparcg.com/
[2021-07-21 12:00:59.272] [info]    ############################################################################
[2021-07-21 12:00:59.272] [info]    Starting CasparCG Video and Graphics Playout Server 2.3.2 4de6d18f Dev
[2021-07-21 12:00:59.456] [info]    Initializing OpenGL Device.
[2021-07-21 12:00:59.459] [info]    Initialized OpenGL 4.5.13587 Core Profile Context 20.2.2 26.20.15019.19000 ATI Technologies Inc.
[2021-07-21 12:00:59.492] [info]    D3D11: Selected adapter: Radeon RX Vega M GH Graphics
[2021-07-21 12:00:59.492] [info]    D3D11: Selected feature level: 45312
[2021-07-21 12:00:59.517] [info]    [asio] Shutting down global io_service.
[2021-07-21 12:00:59.517] [info]    [asio] Global io_service uninitialized.
[2021-07-21 12:00:59.517] [error]   Exception: C:\Program Files (x86)\Jenkins\workspace\casparcg-server-dep\master\src\accelerator\ogl\util\device.cpp(132): Throw in function __cdecl caspar::accelerator::ogl::device::impl::impl(void)
[2021-07-21 12:00:59.517] [error]   Dynamic exception type: class boost::exception_detail::clone_impl<struct caspar::gl::ogl_exception>
[2021-07-21 12:00:59.517] [error]   [struct caspar::tag_msg_info * __ptr64] = Failed to initialize d3d interop.
[2021-07-21 12:00:59.517] [error]   [struct caspar::tag_stacktrace_info * __ptr64] =  0# 0x00007FF6965BA37E in casparcg
[2021-07-21 12:00:59.517] [error]    1# 0x00007FF6965E9690 in casparcg
[2021-07-21 12:00:59.517] [error]    2# 0x00007FF696622510 in casparcg
[2021-07-21 12:00:59.517] [error]    3# 0x00007FF696621033 in casparcg
[2021-07-21 12:00:59.517] [error]    4# 0x00007FF6965F057D in casparcg
[2021-07-21 12:00:59.517] [error]    5# 0x00007FF6965F0475 in casparcg
[2021-07-21 12:00:59.517] [error]    6# 0x00007FF6965D54FC in casparcg
[2021-07-21 12:00:59.517] [error]    7# 0x00007FF6965D619A in casparcg
[2021-07-21 12:00:59.517] [error]    8# 0x00007FF6965BD2E4 in casparcg
[2021-07-21 12:00:59.517] [error]    9# 0x00007FF6965C111D in casparcg
[2021-07-21 12:00:59.517] [error]   10# 0x00007FF6969F6A88 in casparcg
[2021-07-21 12:00:59.517] [error]   11# BaseThreadInitThunk in KERNEL32
[2021-07-21 12:00:59.517] [error]   12# RtlUserThreadStart in ntdll
[2021-07-21 12:00:59.517] [error]   
[2021-07-21 12:00:59.517] [error]   

[2021-07-21 12:00:59.517] [error]    0# 0x00007FF6965BA37E in casparcg
[2021-07-21 12:00:59.517] [error]    1# 0x00007FF6965B9CAF in casparcg
[2021-07-21 12:00:59.517] [error]    2# 0x00007FF696AA09E7 in casparcg
[2021-07-21 12:00:59.517] [error]    3# 0x00007FFE692C1030 in VCRUNTIME140
[2021-07-21 12:00:59.517] [error]    4# is_exception_typeof in VCRUNTIME140
[2021-07-21 12:00:59.517] [error]    5# RtlCaptureContext2 in ntdll
[2021-07-21 12:00:59.517] [error]    6# 0x00007FF6965C111D in casparcg
[2021-07-21 12:00:59.517] [error]    7# 0x00007FF6969F6A88 in casparcg
[2021-07-21 12:00:59.517] [error]    8# BaseThreadInitThunk in KERNEL32
[2021-07-21 12:00:59.517] [error]    9# RtlUserThreadStart in ntdll
[2021-07-21 12:00:59.517] [error]   
[2021-07-21 12:00:59.630] [fatal]   Unhandled exception in main thread. Please report this error on the GitHub project page (www.github.com/casparcg/server/issues).`

### Steps to reproduce
1. use hades canyon pc [Intel® NUC Kit NUC8i7HVK] tried also with a  Core M3-7Y30with Intel HD 615 GPU
2. enable gpu in the config file

### Environment
Commit: 4de6d18f
Server version: v2.3.2
Operating system: Windows 10
PC: Intel® NUC Kit NUC8i7HVK with Radeon™ RX Vega M GH graphics (integrated)

### LINKED ISSUE #1441: Consumer shows countdown clock *skipping* times in the output
I created an html template that uses three.js to produce a countdown clock that has 3d text.

The template code has been created so that rather than re-rendering on every frame, the rendering only happens when the time changes (ie: once per second) as a way of reducing the GPU load.

Problem is: the clock output appears to be skipping some times entirely in the output!

### Expected behaviour
Every second of the countdown should appear in the output

### Current behaviour
Some seconds of the countdown output are not shown in the consumer output

----

### Steps to reproduce
1. Install the clock template to your CasparCG server: [clock.zip](https://github.com/CasparCG/server/files/9848422/clock.zip)
2. Make sure your CasparCG configuration file has the html graphics enabled for access to the GPU.
3. Create a rundown that uses the template
4. Play the countdown.
5. Your consumer (whatever you configured it to go to) should start showing you a countdown to a distant date & time.
6. Observe the consumer output, and you will probably see, occasionally, that the output *skips over* some times being shown
7. This behavior does not happen in a regular chrome browser with graphics acceleration enabled.
8. The template outputs to the CasparCG log the time that it's rendering, so you'll be able to see that it's not skipping over any times.

### Environment
* Commit: [abc3582](https://github.com/CasparCG/server/commit/abc35821b797bbcc9cfedabf396fd2e2ef5c9b8c)
* Server version: 'master'

Note: this is not a new problem introduced with the above commit, but, since the CEF has been updated, I figured I'd mention that I've tested it with the above version too, and find it's skipping the output too.

Also, I've seen this behavior happen where the template is much simpler, and it's just doing 2D text output.

---

### Screenshots

Screen recording of a web browser (top), then the screen consumer (bottom) and the CasparCG log in the background,
showing that it's not skipping over any times, but, watch long enough, and you'll see when the CasparCG output does *not appear*, and the time sticks for an extra second.

Check out the clock at 00:07 of the recording, and you'll see what I'm talking about!

https://user-images.githubusercontent.com/10079005/197439846-4e294992-dda7-46a5-81cb-0455c10eb319.mp4

### LINKED ISSUE #1454: HTML animations sometimes get stuck on the last frame
### Expected behaviour

Animations always play to completion

### Current behaviour

When playing an animation with the HTML producer, the animation will sometimes freeze on the final frame, which will stay on-screen until another update to the HTML page occurs. This only happens on Windows with `enable-gpu` set to true.

My best guess for what's happening is that when CEF calls `OnAcceleratedPaint`, sometimes the GPU hasn't yet executed the write to the shared texture, so caspar might end up reading an old frame when it goes to copy the frame into its queue. Because `OnAcceleratedPaint` is only called whenever the content changes, caspar stays on this old frame until there's another update to the HTML page.

I suspect that #1431 and #1441 are caused by the same root issue.

### Steps to reproduce
1.  Create a HTML file with the following content:
<details>
<summary>index.html</summary>

```html
<!DOCTYPE html>
<style>
@keyframes grow  {
    0%, 25% {
        transform: scale(0);
    }

    75%, 100% {
        transform: scale(1);
    }
}

.test {
    animation: grow 2s cubic-bezier(0, 1, 1, 0) infinite alternate;
    width: 500px;
    height: 500px;
    background-color: red;
}
</style>
<div class="test"></div>
```
</details>

2. Start caspar with `enable-gpu` set to true on the HTML producer

3. Play the above file: `PLAY 1-1 [HTML] file:////path/to/index.html`

4. Observe that sometimes the red square will stop mid-animation. It seems highly variable how often this happens but should occur at least once within a couple of minutes.

### Environment
* Commit: abc35821b
* Server version: 2.3.3
* Operating system: Windows 10 21H2 (19044.2251)

We have also reproduced this on the 2.3.3 stable release, but it seems more common on the above commit.

---

### Screenshots
See 0:09 and 0:18 in the video:

https://user-images.githubusercontent.com/17673205/208012707-6468a9b4-b024-478a-9e4c-784ccc7df207.mp4

### LINKED ISSUE #1498: Update of CEF chrome
### Description
is it possible to update the CEF lib to te latest version that is used in OBS? 
### Solution suggestion

https://github.com/obsproject/obs-studio/pull/6867/files 

---

Julusian:
This work was sponsored by NRK.

Related #1235
Closes #1498
Fixes #1265
Fixes #1454 
Fixes #1387
Fixes #1441
And others?

It has been a few years since I last evaluated whether the shared-texture flow in CEF was necessary. I started off looking into this as I was curious how linux performed, which has never had shared-texture support.

Initial tests were run on a machine with a 7950x (dual channel ddr5) and GTX1060, using https://www.testufo.com/ as the page to produce constant fluid motion. I did try some webgl demos at a couple of points, but results were not impacted noticably. Using a single channel of 1080i5000. From memory, the cost is the same no matter how many channels the producers are spread across, but this needs verifying.

Using the CEF95 in master:
| | CEF95 windows | CEF95 linux | CEF117 windows | CEF117 linux |
|----|----|----|----|----
| gpu disabled | 18 | 12 | 18 | 17 |
| gpu enabled (with shared) | 8 | - | - | - |
| gpu enabled (no shared) | 5 | 12 | 13 | 15 |

Also worth noting, CEF95 gpu-enabled without shared-textures flooded the console with errors from CEF.
At one point I was having issues with CEF117 crashing on linux, that appears to have stopped happening now.

From this, it looks like there is no big change in gpu-disabled behaviour (a change of 1 could be due to my testing method), but gpu-enabled (no shared) is performing much better, and looks to negate the need for shared-textures.

To verify this, I tried this on another machine (i9-10920X & Quadro P4000. quad channel DDR4)
| | CEF71 windows | CEF95 windows | CEF95 linux | CEF117 windows | CEF117 linux | 
|----|----|----|----|----|----
| gpu disabled | 7 | 15 | 16 | 14 | 16 |
| gpu enabled (with shared) | - | 8 | - | - | - |
| gpu enabled (no shared) | 7 | 2 | 12 | 9 | 14 |

And another (i7-7800X & RTX 3080Ti. dual channel DDR4)
| | CEF71 windows | CEF95 windows | CEF117 windows |
|----|----|----|----
| gpu disabled | 4 | 6 | 5 |
| gpu enabled (with shared) | 6 | 7 | - |
| gpu enabled (no shared) | ? | ? | 5 |

And another (xeon W-2133 & Quadro P4000. dual channel DDR4)
| | CEF71 windows | CEF95 windows | CEF117 windows |
|----|----|----|----
| gpu disabled | 3 | 4 | 4 |
| gpu enabled (with shared) | 5 | 6 | - |
| gpu enabled (no shared) | 0 | ? | 3 |

While not a perfect match, depending on the machine used performance can be better than before.

Investigating this, I discovered that the slowness with `gpu enabled (no shared)` was due to the memcpy we are performing, to copy the CEF pixel buffer into our gpu backed buffer. It was notably slower in this vs `gpu disabled`, but wrapping the copy with a `tbb:parallel_for` improves performance such that it rivals the `gpu disabled` flow. Making the same change when `gpu disabled` or on linux has no noticable impact, so it is not done in either of those cases.
This performance optimisations was made for only the CEF117 builds, as this PR is aiming to ensure it doesnt degrade performance compared to the current builds/releases of casparcg. 

For this test I have been focusing on how the number of producers, and not focusing at all on the actual resource usage of this. I expect that there will be more cost on system memory especially for this change, but at this point this is a question of do we want to accept the increased resource requirement, or stick with a 2021 version of CEF for the forseeable future. There are known issues with the current CEF integration (#1454), causing some users to be sticking with the previous version we had, CEF71 from 2018.  
The unfortunate situation is that the shared-texture support in CEF is broken, and has been for over 2 years. Various people have been fixing it for each new release of CEF, most recently OBS. But as of CEF 104 from mid 2022, the patch needs to be rewritten from scratch, which is likely to take quite a while to happen. As it looks like we do not need (but we would like it, once possible again) the shared-texture support, I think it best to drop it for now so that we can update CEF.  
I have tried again at making CEF 103 work for us, starting from the work done by OBS, but getting the D3D to GL interop to not lose frames is proving too challenging. And as that version is already over a year old, how long would we stick to it for?

Windows build: https://builds.julusian.dev/casparcg/casparcg-server-cef117-upstream-2842bc616bb9817547b775eae466d330050d0025-windows.zip


Sidonai-1:
Thank you for the research, I am very interested in trying the build once its merged.

By your post and what I could understand googling a bit: the missing "shared texture" will only affect performance in certain cases, is that correct? Or is there any HTML features/ workflows that will be directly affected or non-functional by its absence?

I am trying to understand what this change entails to remain in the old one with shared-texture support or just update with no fear of the old templates breaking.

Thanks!

Julusian:
When `gpu-enabled` is set to true in the config, that means that CEF will be doing its compositing on the GPU (on windows using D3D). 

When using the `shared textures` flow, CEF passes us a handle to that frame (still on the GPU). We are then able to tell the GPU to adopt that frame into OpenG and make a copy of the frame. During this, the frame will stay entirely within the GPU.

Without shared textures, CEF is downloading the frame to CPU memory and passing us that CPU located pixels. We then perform a copy and send that back to the GPU.

So the difference becomes:
With shared textures: GPU -> GPU (with OBS CEF 103, that might is GPU -> GPU -> GPU, and might need another step adding)
Without shared textures: GPU -> CPU -> CPU -> GPU
In both cases there could be extra copies going on that I am not aware of, but they should only be within a 'zone'

Copying via the CPU is problematic because firstly it requires more copies to be done, which costs CPU time. And going between the GPU and CPU will consume PCIE bandwidth. 
Doing some quick maths, a gen 3 16lane PCIE slot (typical GPU) can do 15760MB/s. With a single 1080p frame being ~8MB, at 50p that means that a little over 39 producers will saturate the PCIE bus (assuming no other overhead). If doing 4k50, that becomes a little over 9.  
More load will also be put on system memory, but as that is a lot faster than PCIE, it is less likely to be an issue. If it does, make sure that every lane (not slot, but lane) of memory is populated, as that could double CPU memory bandwidth.
If using a PCIE 4 cpu and motherboard, the bandwidth is doubled, as is the number of producers. Double it again for gen5.

I should probably repeat the same test in 4K50, to make sure that it remains similarly usable.

-----

As far as Im aware, nothing should break with this, other than a degradation in performance due to the change in copying. Whether that change is noticable is what my measurements in the PR description were trying to figure out.  
It is entirely possible that some templates may break a little, due to the jump in chromium version, but if that happens it is unavoidable and should be unrelated to whether shared-textures is in use or not. If concerned, they will behave the same as in normal chrome/chromium

rrebuffo:
I've been testing some templates in **2.4-beta1** vs **2.4-dev** and it seems that `enable-gpu` no longer works in beta1.
I tried playing a hefty template that uses a lot of GPU and I found that it doesn't use 3D features of the GPU at all:

- 3D GPU usage in 2.4-beta1: 18%
- 3D GPU usage in 2.4-dev: 67%

I captured a video to show the results:
https://www.youtube.com/watch?v=5xDlA70Fyj8
The beginning of the video shows the template working properly (2.4-dev with GPU enabled), after that is playing at ~4fps.

Julusian:
@rrebuffo are you running on windows or linux? 
what gpu are you using?
what resolution is this running at?

CEF should still be gpu enabled, but I did tell it to use a different backend which could have an impact on this. It seemed fine in my testing though

rrebuffo:
Sorry for the lack of context.

My developement rig is:
Intel Core i9-10900X
4 channel 64GB 3600Mhz
Nvidia 1050Ti GPU
Windows 10 22H2

I'm running a custom resolution: 3584x768

I could prepare the template for you to test. It's heavy on SVGs and masks so it's a good candidate to stress chrome.

Sidonai-1:
> @rrebuffo I've taken a look at this, and it seems to be a consequence of some of the performance tweaks made. Primarily, the switch of the angle backend used by CEF is the problem here. I suspect its a case of the GL backend works better when running many simpler templates, but the default backend (d3d11?) is better for fewer more heavy templates. So this is now exposed as a config field.
> 
> Try using this with the next beta build, and it should run better:
> 
> ```
> <html>
>     <enable-gpu>true</enable-gpu>
> 	<angle-backend></angle-backend>
> </html>
> ```

I tried to test this with the 2023-12-22 15:16 build but I couldn't make it show aything on screen. The HTML shows up in the DIAG, but there is no output. I tried GPU true and false, and backend blank, gl, d9 and d11. I didn't have an SDI output at the moment, just the screen consumer.

Julusian:
@Sidonai-1 should be better now. I appear to have been clumsy when doing #1390 and broke most producers (all except ffmpeg).

PR #1506, opened 2023-11-14
feat: update cef to 103

Julusian:
An alternative to #1499, updating to CEF 103 which maintains the shared-texture flow, and hopefully the same performance as CEF95/CEF73 that current builds are using.

The main focus here has been to look into the reports of skipped frames that the current CEF95 has caused. (eg #1441)

CEF is built from https://github.com/Julusian/cef/tree/5060-shared-textures which is https://github.com/obsproject/cef/pull/1 rebased on top of some more recent changes to the OBS 103 branch.

Windows builds:

* ~~[casparcg-server-8548acb3dee0283f303ee2d3310cf9ef0bfb39ba-windows.zip](https://github.com/CasparCG/server/files/13377637/casparcg-server-8548acb3dee0283f303ee2d3310cf9ef0bfb39ba-windows.zip)~~ Flawed, fails the test for #1441


Julusian:
https://github.com/CasparCG/server/pull/1499 looks to be a more viable route forward

PR #1517, opened 2024-01-05
feat: route audio from CEF

Julusian:
Inspired by/based upon a start made by nxtedition at https://github.com/nxtedition/casparcg

This needs a bit more testing to ensure it works smoothly and without glitches.  
In particular, when audio starts/stops within the page, and when motion starts/stops

This also needs a sync test, to check that the audio is in sync and remains in sync with both video and audio pausing and resuming

### DIFF
--- a/src/modules/html/producer/html_producer.cpp
+++ b/src/modules/html/producer/html_producer.cpp
@@ -81,8 +81,12 @@ struct presentation_frame
     bool has_video = false;
     bool has_audio = false;
 
-    explicit presentation_frame()
+    explicit presentation_frame(core::draw_frame video = {})
     {
+        if (video) {
+            frame = std::move(video);
+            has_video = true;
+        }
     }
 
     presentation_frame(presentation_frame&& other)
@@ -256,12 +260,13 @@ class html_client
     {
         if (!try_pop(field)) {
             graph_->set_tag(diagnostics::tag_severity::SILENT, "late-frame");
+            return core::draw_frame::still(last_frame_);
+        } else {
+            return last_frame_;
         }
-
-        return last_frame_;
     }
 
-    core::draw_frame last_frame() const { return last_frame_; }
+    core::draw_frame last_frame() const { return core::draw_frame::still(last_frame_); }
 
     bool is_ready() const
     {
@@ -366,10 +371,7 @@ class html_client
             core::draw_frame new_frame = core::draw_frame(std::move(frame));
             last_generated_frame_ = new_frame;
 
-            presentation_frame wrapped_frame;
-            wrapped_frame.add_video(std::move(new_frame));
-
-            frames_.push(std::move(wrapped_frame));
+            frames_.push(presentation_frame(std::move(new_frame)));
             while (frames_.size() > 4) {
                 frames_.pop();
                 graph_->set_tag(diagnostics::tag_severity::WARNING, "dropped-frame");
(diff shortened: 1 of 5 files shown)


ComingHOME-Darmstadt:
Sorry to contact you this way, I don't know an other way, so I will try it once:
Pull request https://github.com/bitfocus/companion-module-casparcg-server/pull/16
waits for over a year and is so usefull.
Can you merge this for Companion?

Julusian:
@qu1rk3y I did this prototype implementation a while back because I was curious on how hard it would be, but haven't done as much testing as I think it needs before it can be merged.
So it has sat waiting for that testing to be done before it can progress.

If anyone wants to give it a try, there is a build based on top of the master branch at the time of writing at https://builds.julusian.dev/casparcg/casparcg-server-b52edd40855b659798c9b78643e66d1d8e6aebdc-windows.zip
Please report back if it works fine for you or if you had any issues, that will help with confidence to get this merged.

Sidonai-1:
> @qu1rk3y I did this prototype implementation a while back because I was curious on how hard it would be, but haven't done as much testing as I think it needs before it can be merged.
> So it has sat waiting for that testing to be done before it can progress.
> 
> If anyone wants to give it a try, there is a build based on top of the master branch at the time of writing at https://builds.julusian.dev/casparcg/casparcg-server-b52edd40855b659798c9b78643e66d1d8e6aebdc-windows.zip
> Please report back if it works fine for you or if you had any issues, that will help with confidence to get this merged.

I tried to test this but I was unable to run this server, it was complaining about an asio io_service.


This happens in b52edd4 as well as latest master build e4e9ed2

src/modules/html/producer/html_producer.cpp
```diff
@@ -302,7 +369,10 @@ class html_client
         {
             std::lock_guard<std::mutex> lock(frames_mutex_);
 
-            frames_.push(std::make_pair(now(), core::draw_frame(std::move(frame))));
+            core::draw_frame new_frame = core::draw_frame(std::move(frame));
+            last_generated_frame_      = new_frame;
```
niklaspandersson, line 373: @Julusian Could you talk me through how the move semantics works here? As far as I can tell, the draw_frame assignment operator moves the content of new_frame to last_generated_frame_ which would leave new_frame invalid.

src/modules/html/producer/html_producer.cpp
```diff
@@ -399,6 +471,51 @@ class html_client
...
+    }
+    void OnAudioStreamPacket(CefRefPtr<CefBrowser> browser, const float** data, int samples, int64_t pts) override
+    {
+        if (!audioResampler_)
+            return;
+
+        auto audio       = audioResampler_->convert(samples, reinterpret_cast<const void**>(data));
+        auto audio_frame = core::mutable_frame(this, {}, std::move(audio), core::pixel_format_desc());
+
+        {
```
niklaspandersson, line 494: This crashes on Linux. If i comment out this entire block, it runs.

src/modules/html/producer/html_producer.cpp
```diff
@@ -399,6 +471,51 @@ class html_client
...
+            std::lock_guard<std::mutex> lock(frames_mutex_);
+            if (frames_.empty()) {
+                presentation_frame wrapped_frame(last_generated_frame_);
```
niklaspandersson, line 497: Again, I don't understand how this would work. When you construct a presentation_frame from a draw_frame, you move the draw_frame into the presentation_frame. Doesn't that invalidate last_generated_frame_?

niklaspandersson (commented):
It runs on WIndows for me, but I don't understand how it could (see the other comments). The audio doesn't sound right at all though. It's not choppy, it more like every other audio frame is missing is something like that.
It crashes on Linux.

src/modules/html/producer/html_producer.cpp
Julusian: Perhaps this is from the hardcoded 8 in https://github.com/CasparCG/server/blob/72b5ef2c7ad129474f6a519cda0abe922f1e696d/src/modules/ffmpeg/util/audio_resampler.cpp#L31?  
I appear to have opened this PR a few hours before I made a change to support 16 channel audio https://github.com/CasparCG/server/commit/28c4dcd415d0a2d01b11d789ddd97e8e9dfdd178

That would explain the slowness/broken audio you are experiencing too

src/modules/html/producer/html_producer.cpp
Julusian: > As far as I can tell, the draw_frame assignment operator moves the content of new_frame to last_generated_frame_ which would leave new_frame invalid.

The parameter of the assignment operator is not a reference, so will be a copy of `new_frame` making it safe to mutate. You have made me question this now, but this same pattern has been in place for `draw_frame` for years, with the `core::frame_producer` doing it with any `draw_frame` that gets produced by every producer.

src/modules/html/producer/html_producer.cpp
Julusian: Similar to the other question, the constructor parameter isnt a r-value or l-value, so I think it is fine

src/modules/html/producer/html_producer.cpp
niklaspandersson: Ok, I see it now. I missed the implicit call to the copy constructor during the assignment. It looks OK.

src/modules/html/producer/html_producer.cpp
niklaspandersson: I'll try that

Julusian:
Replaced by #1590

PR #1523, opened 2024-01-17
Allow configuring diag font path at build time

oscarwcl:
This allows for better integration with distros that put the system font in a different directory.

As suggested in https://github.com/CasparCG/server/pull/1308#issuecomment-1868345931.

The equivalent of the old `-DUSE_SYSTEM_DIAG_FONT=ON` option is now `-DDIAG_FONT_PATH=/usr/share/fonts/truetype/liberation/LiberationMono-Regular.ttf`


src/CMakeLists.txt
```diff
@@ -12,6 +12,14 @@ set(CASPARCG_DOWNLOAD_CACHE ${CMAKE_CURRENT_BINARY_DIR}/external CACHE STRING "D
 
 option(ENABLE_HTML "Enable HTML module, require CEF" ON)
 
+set(DIAG_FONT_PATH "LiberationMono-Regular.ttf" CACHE STRING
+    "Path to font that will be used to load diag font at runtime. By default
+    this loads the font distribtued with the application from the working
```
nerijus, line 17: typo - distribtued, should be distributed

dimitry-ishenko:
I am a bit late to scene, but what if the font path is not known at build-time? Or, what if it's not the same on the build machine compared to the target machine?...

Julusian:
The aim of this is to be friendly for distributing as a debian/arch/whatever package. In which case it can be reasonably assumed that the font will be found at a predictable path.

Yeah I didn't think about using a system library to locate it. But I probably wouldn't have done it anyway, because it seems like quite an effort for what it is.
I'll accept a pr that uses fontconfig for this, as long as it isn't too big of a heap of code to do so

dimitry-ishenko:
I've looked at the fontconfig API and it's a bit of a mess... One lazy way of doing it would be something like this:

```c++
std::string font_path;
auto pipe = popen("fc-match --format=%{file} LiberationSans-Regular.ttf", "r");
char buf[128];
while (fgets(buf, sizeof(buf), pipe)) font_path += buf;
```

It's marginally less hacky than hard-coding the path, but if you are cool with it, I can spin off a PR shortly.

dimitry-ishenko:
@Julusian see #1559 

One thing I saw that I didn't like was that the `get_default_font()` function contains a static variable (`DEFAULT_FONT`). While it has a nice feature of being initialized on first use, it does involve a hidden lock (since C++11) and a check on each invocation.

In this case it would make more sense to instead create a class static variable (inside `graph`) and initialize it once. There are also a few other statics inside `graph::render()` with the same problem. I believe this function gets called multiple times per second, so there could be some performance hit.

dimitry-ishenko:
Here are test results:

Single-threaded: https://coliru.stacked-crooked.com/a/ba4ab9ed90d28352
```
Static  local: 5824ms
Static global: 4856ms (-16.62%)
```
Multi-threaded: https://coliru.stacked-crooked.com/a/ac8424b20ed46f5a
```
Static  local: 5573ms
Static global: 4738ms (-14.98%)
```

On my local machine I get even more dramatic results:

Single-threaded:
```
Static  local: 5983ms
Static global: 3992ms (-33.28%)
```
Multi-threaded:
```
Static  local: 3697ms
Static global: 2064ms (-44.17%)
```

Issue #1542, opened 2024-06-12
Bug: NDI Playback distorted


jpc0:
### Observed Behavior

It seems like the resolution being fed from NDI into the backend is incorrect. Could also the the pixel format but unlikely. I have attached pictures of the expected output vs what is actually being output. I see the same behaviour on both the decklink consumer output and output from screen consumer so it seems like the issues is before the consumer at the very least. The FFMPEG, IMAGE and COLOUR producers seem to work fine, this does seem restricted to the NDI producer.

I have made some code changes to try faultfinding:
I threw an exception at https://github.com/CasparCG/server/blob/e4e9ed299e1a0f9c87c8709d19b0585a5703102a/src/modules/newtek/producer/newtek_ndi_producer.cpp#L199 and it did not trigger so that confirms that it is in fact not taking that branch.

I also logged the width and height here: https://github.com/CasparCG/server/blob/e4e9ed299e1a0f9c87c8709d19b0585a5703102a/src/modules/newtek/producer/newtek_ndi_producer.cpp#L204 and I can confirm the with and height is 1920x1080 which is the same width and height as the consumers & channel is running at. There are no errors reported from casparcg. A behaviour I did notice was that it takes quite a while (close to a minute) for the image to display and for the logging I placed for width and height to start printing. On windows there is never a delay.

I may try bumping to NDI V6 but that needs some code changes in casparcg, and seeing if that solves the problem.

![expected](https://github.com/CasparCG/server/assets/12384344/8687d67d-4f62-491e-9d71-932f2b3f2abc)

![actual](https://github.com/CasparCG/server/assets/12384344/5545ab35-da80-4a3c-a79a-26d6ac7734f2)

### Expected behaviour

NDI Playback works as expected, and is it does on windows.

### Steps to reproduce

Playback NDI on linux

### Environment

* Commit: HEAD
* Server version: v2.4.0
* Operating system: Ubuntu 22.04

jpc0:
Some more information:

As mentioned in https://github.com/CasparCG/server/pull/1543 discussion it is also distorted on Windows builds. However when I tested it is working correctly on v2.4.0 release. Upon bisecting I found that commit e3289ecac25fca89221f258051222ecdefa08d92 is the commit that broke the expected behaviour.

niklaspandersson:
I'll have a look at the (my) offending commit

PR #1543, opened 2024-06-12
Fix NDI video stretched

jpc0:
This is a potential fix for https://github.com/CasparCG/server/issues/1542, unfortunately I cannot test on windows right now but it does work on linux with and without alpha and at multiple resolutions. Since I suspect there were no issues on windows I rather used an ifdef to not touch the code that runs on windows.

### DIFF
--- a/src/modules/newtek/CMakeLists.txt
+++ b/src/modules/newtek/CMakeLists.txt
@@ -1,51 +1,31 @@
-cmake_minimum_required (VERSION 3.16)
-project (newtek)
+cmake_minimum_required(VERSION 3.16)
+project(newtek)
 
-set(SOURCES
-		consumer/newtek_ndi_consumer.cpp
-
-		producer/newtek_ndi_producer.cpp
-
-		util/ndi.cpp
-
-		newtek.cpp
-
-)
+set(SOURCES consumer/newtek_ndi_consumer.cpp producer/newtek_ndi_producer.cpp
+            util/ndi.cpp newtek.cpp)
 set(HEADERS
-		consumer/newtek_ndi_consumer.h
+    consumer/newtek_ndi_consumer.h
+    producer/newtek_ndi_producer.h
+    util/ndi.h
+    newtek.h
+    interop/Processing.NDI.compat.h
+    interop/Processing.NDI.deprecated.h
+    interop/Processing.NDI.DynamicLoad.h
+    interop/Processing.NDI.Find.h
+    interop/Processing.NDI.FrameSync.h
+    interop/Processing.NDI.Lib.cplusplus.h
+    interop/Processing.NDI.Lib.h
+    interop/Processing.NDI.Recv.ex.h
+    interop/Processing.NDI.Recv.h
+    interop/Processing.NDI.Routing.h
+    interop/Processing.NDI.Send.h
+    interop/Processing.NDI.structs.h
+    interop/Processing.NDI.utilities.h
+    StdAfx.h)
 
-		producer/newtek_ndi_producer.h
-
-		util/ndi.h
-
-		newtek.h
-
-		interop/Processing.NDI.compat.h
-		interop/Processing.NDI.deprecated.h
-		interop/Processing.NDI.DynamicLoad.h
-		interop/Processing.NDI.Find.h
-		interop/Processing.NDI.FrameSync.h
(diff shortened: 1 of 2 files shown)


Julusian:
This bug/fix confuses me, so I want to understand what is happening before merging. I won't be able to do that this week

jpc0:
No problem it confuses me too.

The only reason I even though to use the stride in bytes is because in ffmpeg::make_frame the width is devided by the pixel stride for some reason.

In reality I am thinking about directly creating a casparcg frame from the libndi frame and avoiding the intermediate ffmpeg frame all together but I don't have time right now for such a refactor.

This is more a hacky solution that fixes the problem.

jpc0:
If it helps at all. It is similarly broken on windows. Seems like something changed with NDI library sometime? I am reasonably certian this worked fine on 2.4.0 on release on windows since that was what I was running but with whatever NDI version was running then but when I run it now with the latest NDI it is in fact broken on windows as well. The same fix fixes the issue on linux as well,

jpc0:
Some more testing. It work correctly on commit https://github.com/CasparCG/server/commit/55e3a04b8bc0aac89c0fea3b10458b38210d2e86 however at master it does not. 

It is a pain for me to bisect on my windows PC at home (CEF path fails to apply and AMD GPU BUG so need to apply 2 patches to codebase to get it to just build and run) so tomorrow I will try bisect on the linux box at work and hopefull get you to the exact commit that causes the break. I do suspect there was a change introduced when implementing HDR that broke it and I was running a custom build pre v2.4.0 on windows at work before.

Julusian:
> I do suspect there was a change introduced when implementing HDR that broke it and I was running a custom build pre v2.4.0 on windows at work before.

Once you said it didn't work on windows either, I was wondering about this. Definitely plausible that it could have caused this. I wonder if a few ffmpeg codecs/formats could be affected too

jpc0:
Looks like the commit that broke the behaviour is https://github.com/CasparCG/server/commit/e3289ecac25fca89221f258051222ecdefa08d92. I am happy to close this pull request if we can determine that there is actually other issues affecting ffmpeg but as is seen in the offending commit's change. Before we were referencing the linesize in ffmpeg::pixel_format_desc and after that commit we are referencing width.

I also found the release v2.4.0 is working correctly. So I cherry picked https://github.com/CasparCG/server/commit/e4e9ed299e1a0f9c87c8709d19b0585a5703102a ontop of the v2.4.0 tag and am running that

Two things I would like to see but if we want to open issues to make tracking them easier I'm happy to do that:
- The testing mentioned above, I will happily try to assist in implementation
- A v2.4 branch which only contains bugfixes and no new features. Semver effectively

niklaspandersson:
I would really like to get some automated tests in place as well. After working on other codebases for a long time, it felt really shaky editing shared code without any kind of verification of correctness.

In theory it should be pretty easy to get some tests up and running considering the modular architecture, but my knowledge of build- and test tools for C++ are all but non-existing.

Ffmpeg has a separate setup script for the tests that download all required assets from a location separated from the repository. That might be a possible way to handle test videos for us as well

Julusian:
We already have various build dependencies being fetched from releases in https://github.com/CasparCG/dependencies/ to avoid baking them into the repository, so it would be easy to add them there and get cmake to download them in the same way as the other dependencies when tests are wanted. 

I am open to having some tests, but I too don't know much about C++ test tooling. And I think it needs some thought on what the tests should be aiming to cover, especially as a lot of the bits that could do with tests have hardware/system requirements

Julusian:
@jpc0 
> A v2.4 branch which only contains bugfixes and no new features. Semver effectively

There is https://github.com/CasparCG/server/tree/v2.4.x 

If you look at the version numbers of builds, the `master` branch identifies itself as `2.5.0-dev` currently.

On a related note, I should be thinking about tagging 2.4.1 at some point, but I haven't figured out the criteria for doing so.

dimitry-ishenko:
> On a related note, I should be thinking about tagging 2.4.1 at some point, but I haven't figured out the criteria for doing so.

Yes please. If possible, I'd like to have my scaling mode PR and cef PR backported, so I can roll another Debian package... 😎

niklaspandersson:
PR #1580 will hopefully resolve this issue without the duplicated code

PR #1522, opened 2024-01-16
Fix build using system ffmpeg when ffmpeg include dirs are empty and fix undefined reference to boost::log

nerijus:

Issue #1489, opened 2023-06-01
CMake cleanup


jpc0:
This would just a thread to discus what needs to be cleaned up in the cmake configs and what direction we would want to go with some things. Pinging @dimitry-ishenko in this since he may be able to assist in the near future.

Some ideas:

1. Move PCH from custom implementation to CMake native implementation
    - https://github.com/CasparCG/server/pull/1487
2. Seperate out public and private includes in cmake
    - Currently all the includes are using the older include_directories command
    - It would be beneficial to have the public headers used by other targets exported as public and those that are implementation detail exported as private
3. Fix windows  and linux build process being dramatically different
    - This is unlikely to be entirely fixed while we still use NuGet on windows but we can definitely get this closer to the same
    - https://github.com/CasparCG/server/pull/1502
    - https://github.com/CasparCG/server/pull/1504
4. Cmake install targets?

Julusian:
For 3, I have been looking at vcpkg [in a branch](https://github.com/CasparCG/server/compare/master...Julusian:CasparCG-Server:feat/vcpkg). I'm not convinced on it yet, because it has the cost of making compile times much longer, unless I can find a good way to do binary caching (vcpkg does support this, but there isnt a good easy and free option). 

An overview of the packages we use for each of vcpkg and nuget:
For the 9 dependencies we have, 3 are committed to the repository, 3 are published by me, 1 is up to date, 2 are a bit stale and look abandoned.  
Switching to vcpkg, 8 of those are in the repository and maintained. The last, we can use a custom port. Realistically to make compiling bearable, ffmpeg may need to be avoided in vcpkg (compiling that in github actions takes 2+ hours, and it runs out of disk space), but we could use a custom port to provide a precompiled version. Compiling the other 8 dependencies takes about 20 minutes in github actions without caching.

jpc0:
Yeah I am aware of the painful recompiles of ffmpeg, have a project on RPI where I need to do so and generally need half a day, boost isn't much better though if you need all of boost.

I don't think NuGet is necessarily bad, will just need some config in cmake but there is stuff that can be cleaned up and generalized.

Julusian:
3 is affected by #1502, now using cmake external packages instead of nuget (and caching in docker images?).  
It doesn't clean anything up just yet, but does simplify it a bit by removing nuget. however, there is still a difference of using external packages on windows and system dependencies on linux so it doesnt really affect the cleanup all that much

Julusian:
[...]

PR #1556, opened 2024-09-08
Add FindCEF module & use it for Linux system builds

dimitry-ishenko:


dimitry-ishenko:
@Julusian I was thinking of using `FindCEF` for manually fetched version of libcef as well, but this is a bit tricky since CEF doesn't get pulled until the compile stage. This can partially be worked around by replacing `ExternalProject_Add` with `FetchContent_Declare/FetchContent_MakeAvailable`, which will make `libcef.so` and the headers available for `FindCEF`, but it still leaves `libcef_dll_wrapper.so`... Any thoughts?

dimitry-ishenko:
Another option would be to patch CEF itself and make it export targets that can be pulled in by the Server, whether it's manually-built or provided by the system. This might be a better version. You are already patching it anyway. And I can apply the patch to my CEF packages as well.

BTW, if you are interested, I have CEF packages in my PPA here: https://launchpad.net/~ppa-verse/+archive/ubuntu/casparcg

Julusian:
I don't have strong feelings on this really, just that I find the `Find*` scripts a bit annoying. Partly because they don't work with dependencies pulled in through `ExternalProject_Add`, and partly because you have to bring your own if cmake doesn't bundle it, and then maintain that. (eg #1522 #1418)
When doing this cmake rework, I removed the `FindBoost` for a similar reason, but for that we can at least assume that there will be one already on the system (on linux, windows does it manually).

So I am fine with it being done differently when using system vs downloaded version of the library.

I see you have already found my ppa, the packaging is probably not as properly done as yours, but it works well enough. I shall take a look at what you've done for inspiration

dimitry-ishenko:
OK yeah. CMake people seem to agree (as far as not liking the `FindXXX` scripts) and suggest that packages should export their own config files. I am going to try to patch the CEF package and make it install the config file. This way it should play nicely with both fetched and installed version.

> I see you have already found my ppa, the packaging is probably not as properly done as yours, but it works well enough. I shall take a look at what you've done for inspiration

If you care to give me access to your PPA, I can push my package there. Or, you can just grab mine here:

dimitry-ishenko:
@Julusian I've added build instructions to [my repo](https://github.com/dimitry-ishenko-casparcg/chromium-embedded-117) so you can re-create the package and upload to yours.

Or, if you want, I can be the maintainer for the Debian packages.

I am planning to integrate my CEF patches with the Server, which should help simplify the build system a bit. I will also try to do some work on the build system as discussed in #1489. So, expect some PRs from me shortly. 😅

dimitry-ishenko:
Closing this in favor of #1557

PR #1562, opened 2024-09-15
fix: compilation on 32bit platforms

dimitry-ishenko:
On 32bit platforms passing `long` to `data_t` causes ambiguity, because it doesn't map directly into any of the types held by `data_t`.


Julusian:
I would say that I don't see the value in supporting armhf, or other 32bit platforms, but as the change is so simple I don't object to it

PR #1608, opened 2025-03-25
feat: replace freeimage with ffmpeg

### LINKED ISSUE #1607: Bug:  .tif 'per channel' image consumer crashes server
### Observed Behavior

When trying to play a .tif that has been exported in RRGGBB (per channel in Photoshop) the server closes without error.

Image

Interleaved .tif work no problem.

Example of 'per channel' tif:
[tif per chanel.zip](https://github.com/user-attachments/files/19371707/tif.per.chanel.zip)

In mediainfo the files that crash are labelled "big" in "format settings"

Image

The ones that work normally are labelled "little".

### Expected behaviour

Just for the server to throw an error but not to crash.

### Steps to reproduce

1. Play 'per channel' .tif

### Environment

* Server version: 2.4.3 Stable

Julusian:
closes #1607

This seems to work for every image format I can think of (files generated in gimp). And consumer works in my testing

~~Seems to work for PNGs, havent tested other formats.~~

~~The code isn't very tidy yet~~
~~The consumer needs looking at too. First draft of the consumer is written~~

cc @dimitry-ishenko


dimitry-ishenko:
@Julusian Can we do 2.5 after this, please?

PR #1609, opened 2025-03-26
feat: add frame_converter and provide consumers opaque frame type

Julusian:
This is the first polished portion of https://github.com/CasparCG/server/pull/1581

This removes the `image_data` on `const_frame` and replaces it with an opaque `image_ptr`. Consumers can use the `frame_converter` to convert that into a pixel buffer.
For now this is the same basic texture to buffer copy as before, a follow up will replace that with a format conversion step

Most of the consumers need testing to make sure they work and don't crash immediately. ndi and ffmpeg appear to not work for me, but the same is true with master so is probably something weird on my machine today..

### DIFF
--- a/src/accelerator/CMakeLists.txt
+++ b/src/accelerator/CMakeLists.txt
@@ -5,6 +5,7 @@ set(SOURCES
 	ogl/image/image_kernel.cpp
 	ogl/image/image_mixer.cpp
 	ogl/image/image_shader.cpp
+	ogl/image/frame_converter.cpp
 
 	ogl/util/buffer.cpp
 	ogl/util/device.cpp
@@ -17,6 +18,7 @@ set(HEADERS
 	ogl/image/image_kernel.h
 	ogl/image/image_mixer.h
 	ogl/image/image_shader.h
+	ogl/image/frame_converter.h
 
 	ogl/util/buffer.h
 	ogl/util/device.h
--- a/src/accelerator/ogl/image/frame_converter.cpp
+++ b/src/accelerator/ogl/image/frame_converter.cpp
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2011 Sveriges Television AB <info@casparcg.com>
+ *
+ * This file is part of CasparCG (www.casparcg.com).
+ *
+ * CasparCG is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * CasparCG is distributed in the hope that it will be useful,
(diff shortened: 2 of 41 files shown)


Julusian:
Yeah I am a little worried about potential performance impact, but I also wonder if it will just be fine? I havent done any performance testing though, just some light usage.

The theory I have in my mind from the code is that this should be minimally different, a key thing is that I have made sure to queue the download before buffering the frames in consumers.
As for the downloads themselves, they are being done in the ogl thread and shouldn't be reliant on being awaited by the consumers for the download to progress. So unless the system is struggling, I think this should mean that frames will download in plenty of time. If the frame isn't ready and the consumer is awaiting the frame, this is no different to existing behaviour of blocking waiting for a frame to be pushed to the buffer.

---

I am currently thinking of releasing 2.5.0 any day now.  
This will merge into master(aka 2.6) shortly after, with any follow ups when they look ready.

Julusian:
I did consider doing something like that, but decided against it for some reasons that I don't 100% remember.
I think one reason for this was to avoid having to deal with race conditions when the consumer reinitialises and starts wanting a different format (although thinking about it, maybe that could still be a problem in this implementation if a consumer isnt careful). The other case I can think of is that depending on where this bulk downloading happens, could another consumer be added while this is happening?

The other part of the reason was that I think it will a little complexity to consumers, as they will need to declare the array of frames they need. And then when consuming the frames will need to guard against the frames vector being the wrong length. At the time, I didn't have an idea on how to do this and it felt like it would become a mess, but maybe it won't be as bad as I was worried about.

> It would also allow for optimizations in the case where multiple consumers want the same "view" of the rendered frame 

Yeah it would do that better/cleaner, I have tried to handle this today with some caching, but I acknowledge that it could be prone to race conditions resulting in an unnecessary cache miss, because of the downloads being scheduled on different thread. I felt that a cache miss was better than wrapping it in a mutex.

> It would take one step closer to bringing the ports-functionality to more consumers

I don't think it would change anything. The current approach of this PR allows consumers to download as many versions of the frame they want. All the subregion logic is being moved into this download step, so I don't see what would be any better 

> The total complexity would not be greater than the current implementation in the PR.

I'm not 100% convinced that will be the case. I suspect its going to add a bunch of complexity to the channel (output?) at the point where it needs to do the conversions.  
Sure, this PR does require some changes in every consumer, but they are pretty small and while it does require some thinking about async, not much.

I am also curious what the performance implications would be. Doing the conversions in the channel would mean that (unless a new thread is introduced here for this purpose, with some handling for race conditions) some part of the channel will need to block while these downloads are being done, which goes against the way the mixer is pipelining the compositing?  
Perhaps it would be best to do these downloads as part of that compositing (like the old rgba8 download used to be), but that really raises the risk of race conditions (consumers being added/reconfigured)

PR #1326, opened 2020-07-02
Fix dropped and repeated audio samples for 1001-based standards

scriptorian:
Fixes #1227 with further work to ensure that routes work correctly.
Many thanks to ducthiem90 for most of the work!

PR #1615, opened 2025-04-19
fix audio handling for fractional framerates with stream tags

### LINKED ISSUE #838: Implement new nb_samples API for NTSC support

### LINKED ISSUE #1036: Audio Slowed down and glitching at 59.94 in Caspar CG
Hi Everyone

I am having the following problem when trying to playback any file in caspar that is 59,94 framrate. The audio sounds slowed down, and glitchy as well as the video has micro stutters. This is extremely weird as just last week i played back 59,94 content with absolutely no problem.

I also noticed i could get CG to play back 29,97 content and it would use its converter to double the frame rate and play out correctly at 59,94. However this has also stopped working as well.

I need to play media with alpha channel as well as inserts that don’t require an alpha. My Setup is as follows:

PC:
CPU - i7 3770K
GPU - Gt 740
Ram - DDR 3 1600 mghz 12 Gigs
Drive Adata SSD 500 Gig
Playout Card - Black Magic Decklink 4k Extreme

All files are H.264 Quicktime with a bit rate of 4500 kbps, have tried both AAC audio at 128 kbps as well as PCM audio at 48kHz and 44.1 kHz at both 16 and 24Bit. Changes nothing.

My current work around is to set my channels on CG server to 29,97, output that out of the deckling card and then use 2x UpdownCross HD’s for key and fill to convert it back to 59,94 for the ATEM. it works fine and the audio and video playback is perfect, however i don’t want to sacrifice two converters just for the purpose of ensuring Caspars playback works, when i have had Caspar playback 59,94 files just last week with no issue whats so ever.

Please can someone point me in the right direction!

CasparCG Server Version: 2.0.7
Caspar ClientL 2.0.8

### LINKED ISSUE #1145: 2.2.0 stable Sound issues 1080p5994

### LINKED ISSUE #1319: Decklink audio output is broken with NTSC or 59.94 setup
### Expected behaviour
Test Tone singal playback should be clean

### Current behaviour
Test Tone singal playback is not clean when casparCG is setup with Decklink NTSC or 1080i5994
Playback with PAL or 50, 60 configuration is fine.

### Steps to reproduce
1. Create test tone signal with ffmpeg
  ffmpeg -f lavfi -i testsrc2=rate=29.97:duration=60:size=1920x1080 -f lavfi -i sine=frequency=1000:sample_rate=48000:duration=60 -af volume=-30dB -c:v mpeg2video -b:v 20000k -c:a pcm_s16le tone.mov
2. setup CasparCG with Decklink, NTSC
3. playback generated test signal
  play 1-1 tone
4. Listen HDSDI or AES audio from Decklink (Volume up)
5. You can hear abnormal tone sound with NTSC or 5994 setup

### Environment
* Commit: d442a8f84f68838f91c82f5f57c9634e84f7a23a and later
* Server version: [v 2.3.0]
* Operating system: [Windows 10]
 Decklink Driver [10.11.4 ~ 11.5.1]

with PAL or 50, 60 configuration, tone signal playback is perfect

You can check recorded decklink output here

Normal [PAL, 6000]
https://youtu.be/77crz2iHc5g
https://youtu.be/yaKXQ8g2YE0

Abnormal [NTSC, 5994]
https://youtu.be/ftGc4xQ8OgE
https://youtu.be/Xvi4Weer888

TomKaltz:
This is an adaption of #1326 for the modern codebase.  Ongoing discussion: #1319

This PR fixes variable audio cadence handling by tagging each frame that comes from a producer so that the mixer can distinguish between audio streams.  When a new stream is detected while running a variable cadence channel format, one sample of silence is inserted at the beginning of the stream.  This guarantees (if the producer is producing the proper audio packet cadence) that even when the channel and producer cadences are not synced/aligned that the mixer will always have the desired amount of samples to mix.

This PR also makes sure that routed frames are tagged with a unique tag as to not confuse the mixer.

closes #1319
closes #1326
closes #838
closes #1145
closes #1036 
related #1227
related #1192


TomKaltz:
@scriptorian how does this look?  Cross channel route audio is not working (very garbled) but decklink and oal works perfectly.

Julusian:
Cross channel routes arent working because of a major flaw in `fix_stream_tag`. It is setting each frame being routed to have the same tag, which means the mixer will see them as one 'source' and will share the same buffer of extra samples across them all.  
Trying to give these unique ids might be doable without too much complexity, but will be some more state for the remapper to ensure the unique ids are persistent when the routed content changes.

Perhaps another problem could be that frames can have a null tag, which means they skip this realignment logic. I can't tell where those frames could be coming from (just by looking at the pr), maybe those are just the ones produced from mixing the channel?

The route producer should also be fixing up `frame2b`. For interlaced channels it will only be adjust tags for the upper fields and not lower.

I am still wondering about making the route_producer mix the audio instead of modifying the tags. Not sure how best to do it, but it would be great if the samples mixed for the channel compositing could be used for the route too, which would require some refactoring.


Picking more into the mixer; 
It might be better to check for more than 1 cadence value instead of checking for 1001. That should ensure this would work for custom resolutions too (and shouldnt impact the complexity)  
Will this correctly handle cases such as routing 25p into 50p? or similar frame doubling cases? Another case for the routing logic to consider I suppose, producers could be expected to follow the cadence, even if offset.  
I think the line `last_ptr = buf.data();` is unsafe, as `buf` goes out of scope before this pointer is used.

------

I am not too against the approach in this PR, but I do think it would be nicer (and a lot more work) for the producers to be producing the correct samples. But as I said in the other thread, if getting this working is the best we can do for now then I am fine with it (once the bugs are ironed out)

TomKaltz:
@Julusian I hacked on it a little more and may have covered all of your suggestions.  I have not extensively tested this but seems to work with 5994 files played out to decklink or through channel route out to decklink.

TomKaltz:
[Here are windows binaries](https://drive.google.com/file/d/1kvZjX_KoIw1vjCl4SsBJM5EwMrHmUjAM/view?usp=sharing) if anyone in 5994 land wants to give b17f35ed00c93aad3b54841a5e56450bff498eca a test drive.

sendust:
@TomKaltz
Thank you very much for the improvement work on the 5994 standard audio algorithm. After testing with the binary, I noticed a loud 'tick' sound when performing actions such as play, pause, and unload. It's fine during 1x playback, but the 'tick' sound becomes bothersome when repeatedly performing seek or pause operations.

TomKaltz:
@sendust please email me regarding your test setup: tomkaltz at gmail.

TomKaltz:
Also @sendust the slight pop/tick you’re hearing at start and pause/stop is most likely expected due the nature(or anti-nature) of electronic audio. It’s very easy to start/stop playback at the top or bottom of a waveform and it’s entirely content dependent how much that unnatural audio break is perceived.  This could be mitigated in a couple of ways but it’s outside the scope of this PR.

src/core/producer/route/route_producer.cpp
```diff
@@ -100,11 +171,18 @@ class route_producer
...
                         frame1b = core::draw_frame::push(frame1);
                     }
+
+                    // Update the tag in the frame to allow the audio mixer to distinguish between the source frame and the routed frame
+                    frame1b = self->tag_fix_(frame1b);
+
                     auto frame2b = frame2;
                     if (!frame2b) {
                         // Ensure that any interlaced channel will repeat frames instead of showing black.
                         // Note: doing 50p -> 50i will result in dropping to 25p and frame doubling.
```
Julusian, line 181: I think this comment means that audio in this situation would have already been broken. So I guess this case we shouldn't worry about in this PR. 

Maybe fixing that won't be as bad as I feared it would be.. I might take a look at that

src/core/producer/route/route_producer.cpp
```diff
@@ -143,7 +221,12 @@ class route_producer
             std::pair<core::draw_frame, core::draw_frame> frame;
             if (!buffer_.try_pop(frame)) {
                 graph_->set_tag(diagnostics::tag_severity::WARNING, "late-frame");
+                // If we have variable cadence or interlacing and are cross-channel, preserve the previous frame
+                if ((has_variable_cadence_ || is_interlaced_) && previous_frame_ && is_cross_channel_) {
```
Julusian, line 225: I don't follow why we need to store `previous_frame_` and not just whatever is already in `frame_`?

As this is written, if we end up here, that means the buffer has no frame, so frame X is 'missing'. frame_ currently contains X-1, previous_frame contains X-2. so why do we want to promote X-2 to be the current frame?

It is worth noting that both frame_ and previous_frame are pairs of 'frames' (ie a pair of fields)

-----

The result of this (and the line in the else), is that when routing a interlaced channel into a progressive channel, the second field gets stuck.

src/core/producer/route/route_producer.cpp
```diff
@@ -53,8 +111,13 @@ class route_producer
     std::shared_ptr<route> route_;
 
     std::optional<std::pair<core::draw_frame, core::draw_frame>> frame_;
+    std::optional<std::pair<core::draw_frame, core::draw_frame>> previous_frame_;
     int                                                          source_channel_;
     int                                                          source_layer_;
+    fix_stream_tag                                               tag_fix_;
+    bool                                                         has_variable_cadence_ = false;
+    bool                                                         is_interlaced_ = false;
```
Julusian, line 119: It would be good to prefix these with `source_` to make it easier know what these refer to when reading their usages

Julusian (changes requested):

src/core/mixer/audio/audio_mixer.cpp
```diff
@@ -92,25 +91,68 @@ struct audio_mixer::impl
...
+                    auto offset = int(item_size) - (channels - (n % channels));
                     mixed[n]    = static_cast<double>(ptr[offset]) * item.transform.volume + mixed[n];
                 }
             }
+
+            if (has_variable_cadence && item.tag) {
+                if (item_size + last_size > dst_size) {
+                    auto                 buf_size = item_size + last_size - dst_size;
+                    std::vector<int32_t> buf(buf_size);
+                    std::memcpy(buf.data(), item.samples.data() + dst_size - last_size, buf_size * sizeof(int32_t));
```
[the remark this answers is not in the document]
TomKaltz: @Julusian Where should the resampling occur?  The route producer or the mixer?

Perhaps we restrict cross-format routing as proper sampling might be a tall order for this PR.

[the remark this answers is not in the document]
TomKaltz: @Julusian I'm not sure cross-format routing ever had correct audio handling, did it?  To make this correct, the route producer would need to repackage frames with the correct audio samples I presume.  That's outside the scope of this PR.

I will add a buffer safety check to prevent the segfault and a warning when a cross-channel route is attempted with differing formats that audio may be unstable.

src/core/producer/route/route_producer.cpp
TomKaltz: I'm pretty sure audio is already broken in cross-format scenarios like this.  I'm going to leave it alone for now as I think we will have to fix the route producer to repackage frames/samples to fix cross-format audio routing/mixing.

TomKaltz:
[Here are windows binaries](https://drive.google.com/file/d/1kvZjX_KoIw1vjCl4SsBJM5EwMrHmUjAM/view?usp=sharing) if anyone in 5994 land wants to give c3338750d24f8b5dc011c567efd07f4d3955d996 a test drive.

src/core/mixer/audio/audio_mixer.cpp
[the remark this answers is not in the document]
Julusian: yeah I don't think cross-format routing has had much attention. So I'm not worried about it in this PR as long as it doesn't crash.

PR #1626, opened 2025-05-02
Compile with ffmpeg master

programmerjake:
ffmpeg master removed a bunch of fields that were deprecated, add `#if` version checks so casparcg compiles with ffmpeg master.

~~I just disabled a workaround for DVCPRO HD, since afaict it was fixed in ffmpeg a while ago. I put it in an `#if` version check so it can still be used for old versions of ffmpeg.~~

I have not verified that it was fixed, so I'm not sure how to proceed here...~~should I figure out how to verify that?~~ should I change the code to apply the workaround to new versions of ffmpeg too? (edit: I just applied the workaround for new versions too)


Julusian:
Yeah I think that keeping the fix makes sense as it isn't hard to reimplement. 

Thanks for this!

PR #1630, opened 2025-05-13
fix: Fix typos in comments and strings

rjmunro:
Found using typos tool.

### DIFF
--- a/src/modules/flash/producer/FlashAxContainer.cpp
+++ b/src/modules/flash/producer/FlashAxContainer.cpp
@@ -603,6 +603,7 @@ void STDMETHODCALLTYPE FlashAxContainer::OnFlashCall(BSTR request)
         CASPAR_LOG(debug) << print_() << L" [activity]     " << str;
 
         // this is how templatehost 1.7 reports that a command has been received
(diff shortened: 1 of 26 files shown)


src/modules/flash/producer/FlashAxContainer.cpp
```diff
@@ -603,7 +603,7 @@ void STDMETHODCALLTYPE FlashAxContainer::OnFlashCall(BSTR request)
         CASPAR_LOG(debug) << print_() << L" [activity]     " << str;
 
         // this is how templatehost 1.7 reports that a command has been received
-        if (str.find(L"Command recieved") != std::wstring::npos || str.find(L"Command received") != std::wstring::npos)
+        if (str.find(L"Command received") != std::wstring::npos || str.find(L"Command received") != std::wstring::npos)
```
Julusian, line 606: This one looks like it was intentional, and handling a typo in another system/component/library

rjmunro: fixed. I've also added a comment and a config for the typos tool to ensure it doesn't happen again.

PR #1635, opened 2025-05-27
remove vestigial dependency on boost chrono

programmerjake:
I noticed that boost chrono isn't used anywhere, so remove the remaining references to it.

(found while trying to figure out why I get a deprecated header warning on Debian 12, turns out the actual git commit boost used for the boost 1.74.0 release doesn't match the 1.74.0 tag in boost coroutines, which includes a fix, annoying...)


dimitry-ishenko:
If only would someone also nuke boost filesystem as well... 😅

Issue #92, opened 2013-04-02
Playback/record of 608/708 CC and VANC


hummelstrand:
This request comes from a LinkedIn discussion:
All US broadcast stations, cable systems and satellite to home broadcasters -- even those that don't originate their own programming -- must have the ability to pass through any captions present in the programming they air. Most stations also have obligations to air at least some of their locally-produced programming with closed captions. 

Equipment that doesn't at the very least pass CEA-708 captions (preferably in SMPTE-334 caption data packets) can't be considered by even the smallest broadcasters.

TomKaltz:
@kierank Can BMD Decklink output 708 SDI embedded captioning?

kierank:
It lets you set VANC data, so if you write the code to support 708, then yes.

euphoricz:
I am currently attempting to implement 708 pass through in CasparCG server for Decklink devices as a proof of concept, it is easy enough to obtain the ancillary data using function call GetAncillaryData in the producer and passing it in the basic_frame to the consumer where I set it with SetAncillaryData.

I also had to change fill_frame in the schedule_next_video function within the decklink consumer to IDecklinkMutableVideoFrame so I can have access to ancillary functions.

The main issue I'm running into however is the producer input pixel format which is BGRA 8 bit needs to be a 10 bit pixel format since this is required by 708 closed captioning.

I have changed the pixel format to 10 bit YUV for output in the decklink consumer however changing this in the producer creates incorrectly colored video, this is most likely due to the stride being incorrect however I'm still trying to track down where this is being set.

euphoricz:
This issue is much more complex than it sounds.

708 is stored in the vanc which requires a 10 bit pixel format.

Caspar currently passes 8 bit through the pipeline.

Converting an input to 8 bit strips out the vanc data, not to mention ffmpeg will also strip out any closed captions on encoding.

In order to use ffmpeg to encode a 10 bit pixel format such as YUV422P ffmpeg must be compiled to do so, based on this post on stackexchange it sounds like if ffmpeg can encode 10 bit it will no longer encode 8 bit:  http://video.stackexchange.com/questions/13164/encoding-422-in-10-bit-with-libx264

[...]

PR #1637, opened 2025-06-10
Add support for closed captions

programmerjake:
Adds support for ASTC A/53 Part 4 Closed Captions (essentially EIA-608 and CTA-708).

This adds a extensible side-data mixer (side-data naming is based on FFmpeg's `AVFrameSideDataType`), which for now just handles closed captions. For closed captions, for each layer, it chooses which input layer to take the closed captions from based on a priority associated with each layer. That priority can be any positive floating-point number. Zero, NaN, or negative numbers mean that there is no priority, so that layer won't be used for closed captions, allowing you to turn off closed captions.

The side-data mixer currently doesn't properly support switching the closed captions source in the middle of an input video, currently it'll just stop the previous video's side-data and start the next one's side data, rather than e.g. building a proper EIA-608 transition sequence. It does warn when you change the source like that though.

The side-data mixer also has functionality to use side-data from dropped frames in the input -- the code allows different side-data types to handle that differently since some types of side-data may only want to use the side-data from the frames that are actually sent to the output (e.g. `AV_FRAME_DATA_DYNAMIC_HDR_PLUS` I guess), and some may not want to miss any frames (closed captions, where missing a frame of data can cause the captions to be garbled).

So far, I've implemented closed captions support in:
* FFmpeg producer: it supports getting closed captions from either the video stream's side data, or from a separate subtitles stream (e.g. in `.mxf` files). It also has support for reading a subtitles-only file (e.g. `.scc` or `.mcc`).
* FFmpeg consumer: it supports putting closed captions into the video stream's side data. It doesn't yet support generating the VANC stream for `.mxf` files, or other file types that need the closed captions to be a separate subtitles stream.
* Decklink consumer: supported only with FFmpeg 8.0 or later. Prints a warning and disables closed captions output on earlier versions.

I added a ACMP command (`MIXER CLOSED_CAPTIONS_PRIORITY`) for changing a layer's closed-captions priority.

Since information on which inputs to use for closed captions or not is stored in the frame transforms, and those are all cleared by the `MIXER CLEAR` command, I decided to add a per-producer closed-captions priority setting that is associated with the producer rather than only with frame transforms. to set that, you have to pass the new `CLOSED_CAPTIONS_PRIORITY` option to whatever `LOAD`/`LOADBG`/`PLAY` command you use to create that producer. That has the benefit of also following the transition between different producers in a layer, rather than you having to be quick and sending `MIXER CLOSED_CAPTIONS_PRIORITY` at the right time. An additional benefit of having the new `CLOSED_CAPTIONS_PRIORITY` argument for `PLAY` is that you can enable closed captions directly from the config file if you didn't want to start up an ACMP client or type any commands into stdin.

# FIXMEs for future contributors

I'm no longer doing substantial work on this since I'm busy with other stuff, so there is stuff that I'm leaving for someone else to do:
* after writing the whole system around A53 CC packets (EIA-608/708), I realized that it would probably be better to base it around something equivalent to FFmpeg's `AVSmpte436mCodedAnc` instead, since that retains all the information around line numbers, frame rates, and other information that is stripped when converting to A53 CC packets. I wrote the code in FFmpeg for that.


programmerjake:
I'm not sure this PR will fix everything wanted in #92, so I'm not marking this as fixing that issue.

programmerjake:
I'll work on adding the proper `#ifdef`s for older versions of ffmpeg tomorrow, I was testing on Debian 12 and it turns out Ubuntu 22.04 uses an even older version of ffmpeg than I was expecting.

programmerjake:
rebased on #1642 again, I also added EIA-708 VANC generation code for the decklink consumer based on libklvanc and the FFmpeg decklink code, afaict it generates VANC packets that are sent to @niklaspandersson's code successfully, but the SDI monitors I have hooked up don't show any closed captions -- either I'm not using them correctly or something is broken or both.

stuff from the log:
```
[2025-06-13 19:10:13.669] [trace]   decklink consumer: got A53_CC side data: fc 97 23 fd 80 80 fa 00 00 fa 00 00 fa 00 00 fa and 44 bytes more
...
[2025-06-13 19:10:13.669] [trace]   decklink consumer: generated VANC packet from A53_CC side data: 96 69 49 4f 43 00 70 72 f4 fc 97 23 fd 80 80 fa 00 00 fa 00 00 fa 00 00 fa 00 00 fa 00 00 fa 00 and 41 bytes more
```

src/modules/decklink/consumer/vanc_frame_side_data_strategy.cpp
```diff
@@ -0,0 +1,311 @@
...
+        LKV_HANDLE_ERR(klvanc_convert_EIA_708B_to_packetBytes(eia708_cdp.get(), &bytes_p, &byte_count));
+
+        struct bytes_deleter final
+        {
+            void operator()(std::uint8_t* bytes) const noexcept { free(bytes); }
+        };
+
+        std::unique_ptr<std::uint8_t[], bytes_deleter> bytes(bytes_p);
+
+        vanc_packet retval = {0x01, 0x61, 11, std::vector(bytes.get(), bytes.get() + byte_count)};
```
niklaspandersson, line 237: This might be your problem. I think you've swapped the DID & SDID values.

programmerjake: libklvanc has `sdid = 0x1, did = 0x61`:
https://github.com/stoth68000/libklvanc/blob/b409fc2b0b8051c871f89367a3489f8aa2b6ed37/src/core-packet-eia_708b.c#L532
```c++
klvanc_sdi_create_payload(0x01, 0x61, buf, byteCount, words, wordCount, 10);
```
https://github.com/stoth68000/libklvanc/blob/b409fc2b0b8051c871f89367a3489f8aa2b6ed37/src/libklvanc/vanc.h#L243
```c++
int klvanc_sdi_create_payload(uint8_t sdid, uint8_t did,
```

but `vanc_packet` puts `did` first. so they are swapped. well spotted!

dimitry-ishenko:
> turns out the bugs in interlaced modes were actually partially ffmpeg bugs, fixed in [FFmpeg/FFmpeg@61a9f3c](https://github.com/FFmpeg/FFmpeg/commit/61a9f3c0ce7f74ba869f44a49e16f3cd1b79e18d)
> 
> so if we want this to work we'll need a newer version of ffmpeg than is available from apt on the Ubuntu versions we use.

@programmerjake that patch was merged into ffmpeg 6.1 and in Ubuntu 24.04 (Noble) we are on ffmpeg 6.1.1. Are you referring to Jammy?

programmerjake:
> @programmerjake that patch was merged into ffmpeg 6.1 and Ubuntu 22.04 (Jammy) is the only distro that doesn't have it. Which Ubuntu versions are you referring to?

ah, I thought it wasn't available on noble last I checked. regardless we'll need to change it to build ffmpeg rather than use the system ffmpeg when the system ffmpeg is too old.

Julusian:
> regardless we'll need to change it to build ffmpeg rather than use the system ffmpeg when the system ffmpeg is too old.

we used to do this, but it was a pain to maintain, and there was push from users to use the system version.
I am open to having a cmake config field so that it can be built against a non-system version, but I don't think we will ship builds which do that

programmerjake:
> we used to do this, but it was a pain to maintain, and there was push from users to use the system version.

then we can't use ffmpeg for framerate conversion when there are closed captions, since it messes up closed captions by duplicating and/or deleting frames with their closed captions side data, which was fixed in that commit.

dimitry-ishenko:
> regardless we'll need to change it to build ffmpeg rather than use the system ffmpeg when the system ffmpeg is too old.

@programmerjake sorry I don't follow. Like I said, Ubuntu Jammy is the only version that has FFmpeg older than 6.1. Are you on Debian Stable maybe? I believe they still use FFmpeg 5.1

programmerjake:
I thought the point of supporting Ubuntu Jammy is that it's supposed to just work, requiring you to compile your own ffmpeg and not providing one with the released binary packages doesn't sound like being supported to me.

I'm using Debian 12 for development, but that's not what I'm complaining about since you never claimed to support it.

dedicatedbroadcastsolutions:
> Oh, FFmpeg with DeckLink can't be statically linked to CasparCG since they both include the DeckLink loader.

Why are we using ffmpeg with decklink enabled?  I thought casparCG has its own decklink implementation...

dimitry-ishenko:
> The commit I linked to isn't the whole fix, it's just the latest in a sequence of commits that fix the bug, since they apparently had a partial fix before.

OK I got it.

> I was thinking that since we're using custom ffmpeg, we'd compile the same version of ffmpeg used on windows (iirc 7.0.2) and just statically link it to casparcg.

Static linking is an infernal abomination, a heretical curse spat in the face of divine code. For large projects, it bloats binaries to grotesque, unholy sizes, rips open vulnerabilities like a gaping wound in the digital abyss, squanders memory with the ravenous greed of a demon, and chains developers to an endless, torturous grind of recompilation for every trivial shift in a subproject, dragging their souls through a screaming void of build cycles.

Let's not do static linking.

programmerjake:
> > I was thinking that since we're using custom ffmpeg, we'd compile the same version of ffmpeg used on windows (iirc 7.0.2) and just statically link it to casparcg.
> 
> Static linking is an infernal abomination...

I think that's greatly exaggerated and dynamic linking has significant downsides too, but I don't care enough to argue about it here.
> 
> Let's not do static linking.

ok. Debian has a [potentially working branch backporting ffmpeg 7.1 to debian 12](https://salsa.debian.org/multimedia-team/ffmpeg/-/tree/debian/bookwoorm-backports?ref_type=heads), which is pretty close to jammy, maybe that could be adapted into a jammy backport?

programmerjake:
> * It is only a problem if using interlaced. (This might not be worth considering, I don't have any idea about distribution of usage on this)

it is broken whenever the ffmpeg `fps` filter is used to change frame rates, so whenever the source file's framerate doesn't match casparcg's internal progressive frame rate.

> So as long as it is communicated well that this is known to be broken in some scenarios, I am ok with that

ok, in that case I'll add compile-time detection and log an error at runtime whenever it would be broken, and then just disable closed captions support in the ffmpeg producer.

dimitry-ishenko:
> I think that's greatly exaggerated and dynamic linking has significant downsides too, but I don't care enough to argue about it here.

Yep. I ran it through grok twice to make sound extra gloomy.

> Debian has a [potentially working branch backporting ffmpeg 7.1 to debian 12](https://salsa.debian.org/multimedia-team/ffmpeg/-/tree/debian/bookwoorm-backports?ref_type=heads), which is pretty close to jammy, maybe that could be adapted into a jammy backport?

Unfortunately that doesn't help. For example, it needs libsrist-dev, libjxl-dev and librav1e-dev none of which are in jammy. I've tried backporting libjxl-dev, but it needed 4 other other packages that are not in jammy, and librav1e-dev needed another 30 or so rust-related packages.

programmerjake:
> if it isn't in jammy's ffmpeg why go to a lot of effort to enable all the new codecs and stuff in a backport to jammy? imo just disable the ffmpeg configure features that need stuff not in jammy.

I backported ffmpeg 7.1.1-1 from debian to jammy, I disabled `libjxl-dev`, `librav1e-dev`, `librist-dev`, and `libvpl-dev`:
https://salsa.debian.org/programmerjake-guest/ffmpeg-casparcg-backports/-/tree/jammy-casparcg-backport?ref_type=heads

The only missing dependency is `libjs-bootstrap5` which is only needed by `ffmpeg-doc`, the version from ubuntu noble works fine since it is only javascript and the only dependency it has hasn't changed at all.

programmerjake:
> > if it isn't in jammy's ffmpeg why go to a lot of effort to enable all the new codecs and stuff in a backport to jammy? imo just disable the ffmpeg configure features that need stuff not in jammy.

> That's why I've opted to back-port noble's version of ffmpeg (6.1.1) into jammy only. That way noble and oracular are not impacted.

Backporting only to jammy was exactly what I meant, I see no reason to have our own ffmpeg for noble or oracular or any other distro if their ffmpeg already has everything we need.

dimitry-ishenko:
> > > if it isn't in jammy's ffmpeg why go to a lot of effort to enable all the new codecs and stuff in a backport to jammy? imo just disable the ffmpeg configure features that need stuff not in jammy.
> 
> > That's why I've opted to back-port noble's version of ffmpeg (6.1.1) into jammy only. That way noble and oracular are not impacted.
> 
> Backporting only to jammy was exactly what I meant, I see no reason to have our own ffmpeg for noble or oracular or any other distro if their ffmpeg already has everything we need.

@programmerjake I've back-ported ffmpeg into jammy in my PPA and recompiled the server with it. Not that it matters right now I guess, but when server 2.5 is ready for release (ahem @Julusian) it should work with your patches.

programmerjake:
> > So as long as it is communicated well that this is known to be broken in some scenarios, I am ok with that
> 
> ok, in that case I'll add compile-time detection and log an error at runtime whenever it would be broken, and then just disable closed captions support in the ffmpeg producer.

Added in the latest commit.

programmerjake:
rebased on top of #1633 (well actually I copied the changes but reorganized them into a few commits since rebasing every commit would have been painful because of the merge conflicts)

programmerjake:
Switched to using the new FFmpeg 8 API for manipulating VANC packets instead of using libklvanc since last I checked libklvanc doesn't really work with MSVC.

dedicatedbroadcastsolutions:
FYI: this PR's `av_opt_set_bin`/`av_opt_array_ref` changes to `decklink_producer.cpp` also happen to fix the `archlinux:latest` CI job, which is currently broken on `master` due to Arch's rolling FFmpeg package removing `av_opt_set_int_list` (see #1771 for a minimal standalone fix, and the failing run at https://github.com/CasparCG/server/actions/runs/31319015756/job/93258882705). Worth noting as it's an extra reason to get this reviewed — it's currently the only thing fixing that CI failure for anyone rebasing onto latest master.

dedicatedbroadcastsolutions:
Can we please get some reviews of this pull request?

programmerjake:
rebased on #1633, I guess I'll have to wait on that one being merged so I can then rebase on the merge commit to fix the conflicts here.

niklaspandersson:
Some initial comments. I'm sorry it has taken so long. But, as I'll explain below, this PR is difficult to review and that's probably why no one has done it yet.

There is a lot of things going on in this PR, and specifically in the 0f28852c174c5bbfdb3b048a76435bc476d964c2 commit. Makes it very difficult to review, and by extension, approve.

1. There are two major changes to core interfaces, forcing you to touch a lot of files, that I am not sure are necessary.
   1. Changing out the `frame_producer` type for the new `frame_producer_and_attr`. To me, just adding some kind of `producer_attributes` struct to the frame_producer type would have been a lot less invasive while still allowing you to add arbitrary attributes to the frame_producers.
   2. Adding the side_data argument to the `mutable_frame` constructor. You yourself assign the side_data to the frame after construction in `ffmpeg::make_frame`, so forcing everyone to pass an empty side_data argument on construction seems unnecessary. I'm fine with a constructor overload, or if you put it as the last argument with a default value of an empty side_data, but I don't think it should be a required argument.
      Related to this, I also think that the `ffmpeg::make_frame` function should accept the "old" signature, again to avoid forcing every callsite to be updated. And in this case also to have a `core::frame_side_data_queue`, even if it is completely irrelevant for that producer. Calling it with a (default valued) nullptr as side_data_queue should be fine, and would avoid a lot of churn in the codebase.
2. There are also some changes that are not relevant for this feature that I think would be better off in separate PRs. I approve of striving to make the code "nicer" or "better", but those kinds of changes need to be small and isolated to not make it difficult to review the actual feature changes.
   1. The channel abstraction for cross-thread communication is a good idea, but it is not relevant to this PR and should be in a separate PR. If you want it for the new code, alright. But it is a bit of a big refactor to replace existing code with it as part of this PR.
   2. The same goes for `create_buffersink` and the various `get_supported_XXX` helpers. They are definitely not trivial nor relevant to this PR and should be in a separate PR.

I have stop here for now. I will address more specifics in a follow-up comment, hopefully tomorrow, but I wanted to get this out of the way first.

src/core/frame/frame_transform.h
```diff
@@ -131,13 +131,98 @@ struct audio_transform final
...
+    {
+        return !(l < r);
+    }
+
+  private:
+    /// larger values mean higher priority; zero, negative, or NaN means don't use closed captions
+    float value_;
+};
+
+struct side_data_transform final
```
niklaspandersson, line 204: This side data transform class is big enough to motivate a separate file

src/core/mixer/side_data/side_data_mixer.cpp
```diff
@@ -0,0 +1,237 @@
...
+frame_side_data_in_queue side_data_mixer::mixed()
+{
+    std::vector<const_frame_side_data> mixed;
+
+    auto mix_result = impl_->mix_node(impl_->transform_stack_[0], mixed);
+
+    if (mix_result.a53_cc_side_data) {
+        if (impl_->last_a53_cc_source_queue_ && impl_->last_a53_cc_source_queue_ != mix_result.cc_source_queue) {
+            // TODO: implement proper handling for switching between different a53 cc sources
+            CASPAR_LOG(warning) << L"side_data_mixer: A/53 closed captions source changed,\n"
```
niklaspandersson, line 203: Isn't the whole thing with mixing priorities and stuff moot if the resulting cc could be corrupted if mixing is used? 
If this is diffucult to fix, would it be possible to start with a simpler mixing scheme to avoid this issues? For instance to always prefer a source with cc that is already playing?

src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -116,35 +132,263 @@ core::color_space get_color_space(const std::shared_ptr<AVFrame>& video)
...
+    static inline constexpr int           eia608_video_height  = 16;
+    static inline constexpr AVPixelFormat eia608_video_pix_fmt = AV_PIX_FMT_RGB24;
+
     Decoder() = default;
 
-    explicit Decoder(AVStream* stream)
+    explicit Decoder(AVStream* stream, Decoder* eia608_subtitles_decoder, bool remove_a53_cc)
         : st(stream)
     {
+        auto debug_stream_kind = stream->codecpar->codec_id == AV_CODEC_ID_EIA_608       ? "eia-608"
```
niklaspandersson, line 325: This is quite a lot of new, CC specific code in this already large file. Consider extracting this and the SubstitutedA53CCQueue class to a separate file

src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -116,35 +132,263 @@ core::color_space get_color_space(const std::shared_ptr<AVFrame>& video)
     return result;
 }
 
+/// queue for A53 CC side data, needed to avoid EIA-608 and CEA-708 side data getting out of sync or split up in ffmpeg
+/// filters
```
niklaspandersson, line 136: Please provide a sentence or two describing how this works

niklaspandersson (changes requested):
I agree with the general architecture of this PR.

But like I said above in previous comments, I would like to see some changes to confidently be able to approve this:
- you need to reduce the footprint and don't touch more files than necessary
- Extract side-data / cc-data related code to separate files in some places
- Fix the mixer issue. Although, that is isolated to cc related code, so I'm fine with getting that fixed in a later PR

dedicatedbroadcastsolutions:
Regarding mixing of CC, switching mid stream leads to corruption until the next display clear command which usually happens every few seconds unless it is in endless scroll mode which is less used.  Properly implementing this requires some amount of encode/decode of captions or at least deep inspection of the stream.  Most implementations I'm planning on will not involve switching between files mid stream.  With the current priority system it is up to the client to set which file has the highest priority and this is the source.  Having multiple priorities set allows situations like playing a shorter file in the middle of a larger file or join in progress functionality where captions revert to another source when the playback finishes.  That won't be fully corruption safe until at least a few extra clear commands are encoded and added to the stream but for now it is up to the client to avoid overlapping files that could have problems.  One workaround is to have the client play a file with just clear commands in CC for a few seconds with high priority in switching.

The amount of work required for a proper solution to mid stream switching without momentary issues caused us to choose to put that feature off for a future project.  Your other points seem reasonable though.

src/core/mixer/side_data/side_data_mixer.cpp
dedicatedbroadcastsolutions: Mixing priority is still useful even once encode/decode and proper mixing is implemented for some of the reasons I stated below

programmerjake:
> Some initial comments. I'm sorry it has taken so long. But, as I'll explain below, this PR is difficult to review and that's probably why no one has done it yet.
> 
> There is a lot of things going on in this PR, and specifically in the [0f28852](https://github.com/CasparCG/server/commit/0f28852c174c5bbfdb3b048a76435bc476d964c2) commit. Makes it very difficult to review, and by extension, approve.

Sorry, the commits were split out more cleanly before, but rebasing that onto more recent code was a giant mess so I ended up just manually merging the final state and re-concocting all the commits but that's a lot of code to split into separate commits so I gave up after a while and put the rest in that one commit.

> 
> 1. There are two major changes to core interfaces, forcing you to touch a lot of files, that I am not sure are necessary.
>    
>    1. Changing out the `frame_producer` type for the new `frame_producer_and_attr`. To me, just adding some kind of `producer_attributes` struct to the frame_producer type would have been a lot less invasive while still allowing you to add arbitrary attributes to the frame_producers.

ok, I had it separate since those attributes aren't logically part of the frame producer -- they're for mixing the frames, not producing them. also that would allow me to avoid having to change all `frame_producer` implementations, many of which I can't test since I don't have the right hardware and/or software.

>    2. Adding the side_data argument to the `mutable_frame` constructor. You yourself assign the side_data to the frame after construction in `ffmpeg::make_frame`, so forcing everyone to pass an empty side_data argument on construction seems unnecessary. I'm fine with a constructor overload, or if you put it as the last argument with a default value of an empty side_data, but I don't think it should be a required argument.

ok, I had it as a required argument to make it much easier for producers to actually think about if there should be side data or not. imo having implicit defaults is a good way to miss critical spots when making changes.

>  Related to this, I also think that the `ffmpeg::make_frame` function should accept the "old" signature, again to avoid forcing every callsite to be updated. And in this case also to have a `core::frame_side_data_queue`, even if it is completely irrelevant for that producer. Calling it with a (default valued) nullptr as side_data_queue should be fine, and would avoid a lot of churn in the codebase.

ok, though that breaks whenever `ffmpeg::make_frame` finds side-data since it needs a consistent `frame_side_data_queue` to avoid deleting/duplicating side-data when frames are deleted/duplicated when the source's frame rate doesn't match the output frame rate. so it can't create a new `frame_side_data_queue` for every frame.

That's why I designed `ffmpeg::make_frame` to always require a valid `frame_side_data_queue`.

> 2. There are also some changes that are not relevant for this feature that I think would be better off in separate PRs. I approve of striving to make the code "nicer" or "better", but those kinds of changes need to be small and isolated to not make it difficult to review the actual feature changes.
>    
>    1. The channel abstraction for cross-thread communication is a good idea, but it is not relevant to this PR and should be in a separate PR. If you want it for the new code, alright. But it is a bit of a big refactor to replace existing code with it as part of this PR.

iirc I replaced the existing code in ffmpeg/producer/av_producer.cpp with the channels because the amount of reconfiguring where packets and frames go would make a mutex/condvar version very confusing whereas channels are generally easy to understand. so if it's separated out into a separate PR, this PR would have to depend on it. it does actually need to change `Decoder`'s logic a bunch to route all the different parts properly.

>    2. The same goes for `create_buffersink` and the various `get_supported_XXX` helpers. They are definitely not trivial nor relevant to this PR and should be in a separate PR.

I created `create_[a]buffersink` because there's like 6 different places where they all do nearly the same thing with duplicated code that iirc was all buggy in different ways, so I figured it would be better to use a common implementation where the bugs can be fixed instead of having to remember to fix all the other spots in different ways every time. I can split it out into a separate PR if you like, but this PR will depend on it.

src/modules/ffmpeg/producer/av_producer.cpp
programmerjake: I'll add some comments describing that, but basically it replaces the side data that ffmpeg sees with a counter (since ffmpeg likes to split-up/merge side data which can mess it up when trying to get the output to match the input file), and then figures out which counts it sees when the side-data comes back out of ffmpeg and uses that to reconstruct what the proper side data should be.

programmerjake:
> * Fix the mixer issue. Although, that is isolated to cc related code, so I'm fine with getting that fixed in a later PR

yeah, like @dedicatedbroadcastsolutions says, properly mixing two cc sources is quite difficult, so that's not happening in this PR. Also, I'm not planning on doing more work once this PR is merged since I'm busy with other unrelated projects, so someone else will have to work on cc mixing, though based on how difficult that is, that may just not happen (and that's probably mostly fine).

programmerjake:
I'll work on rebasing this and splitting out some of the changes into separate commits probably tomorrow.

programmerjake:
I split out some of the changes into separate commits so they can be separate PRs (once the conflicts are fixed), I still need to continue working on rebasing though.

programmerjake:
I finished splitting everything into separate commits while keeping the overall diff identical, I still need to rebase on latest `master`, and add the changes requested.

programmerjake:
I rebased on master and rearranged the commits into a few separate branches that are then merged together at the appropriate points, that way I can make separate PRs for those branches: #1791 #1792 #1793

sadly github doesn't show the branching structure on the PR.

I still need to address the review comments and the windows build failures

programmerjake:
> >    1. Changing out the `frame_producer` type for the new `frame_producer_and_attr`. To me, just adding some kind of `producer_attributes` struct to the frame_producer type would have been a lot less invasive while still allowing you to add arbitrary attributes to the frame_producers.
> 
> ok, I had it separate since those attributes aren't logically part of the frame producer -- they're for mixing the frames, not producing them. also that would allow me to avoid having to change all `frame_producer` implementations, many of which I can't test since I don't have the right hardware and/or software.

I tried to refactor it to have `cc_priority` be a field of `frame_producer` as you suggested, however when trying to do that, making sure `cc_priority` stays up-to-date when the field changes and your `frame_producer` is just supposed to be a thin wrapper over one or more other `frame_producer`s makes it imo even more invasive since I have to go through and change nearly all `frame_producer`s to correctly propagate changes so we don't end up with a inconsistent mess that causes strange behavior.

That makes me think that having `frame_producer_and_attr` is really a much cleaner way to do it, or maybe have a new `frame_producer` that just wraps an internal `frame_producer` and changes the cc-priority of the returned frames, and then have the commands that change cc-priority auto-insert that wrapper as needed.

@niklaspandersson what do you think?

I will instead refactor for the other suggestions while I wait.

programmerjake:
> 2\. Adding the side_data argument to the `mutable_frame` constructor. You yourself assign the side_data to the frame after construction in `ffmpeg::make_frame`, so forcing everyone to pass an empty side_data argument on construction seems unnecessary. I'm fine with a constructor overload, or if you put it as the last argument with a default value of an empty side_data, but I don't think it should be a required argument.

refactored to having `side_data` as the last argument of `mutable_frame::mutable_frame()` with a default value, as suggested.

programmerjake:
> 2\. Related to this, I also think that the `ffmpeg::make_frame` function should accept the "old" signature, again to avoid forcing every callsite to be updated. And in this case also to have a `core::frame_side_data_queue`, even if it is completely irrelevant for that producer. Calling it with a (default valued) nullptr as side_data_queue should be fine, and would avoid a lot of churn in the codebase.

refactored to also accept the old signature, now `AVProducer::Impl` is the only place that passes in a non-null `side_data_queue`

src/modules/ffmpeg/producer/av_producer.cpp
programmerjake: I expanded the documentation comment and moved `SubstitutedA53CCQueue` to its own file.

dedicatedbroadcastsolutions:
I integrated this into a test branch with AI filling some of the gaps.  It identified an infinite loop that was causing a bug during my testing where playback would stall.

I wasn't sure how to make inline change requests with that tool on here without seeing the diff so I had it make a pull request against the branch this is based on.

https://github.com/programmerjake/casparcg-server/pull/3/changes

src/modules/ffmpeg/producer/substituted_a53_cc_queue.h
```diff
@@ -0,0 +1,195 @@
...
+                        p += sizeof(k.data);
+                        size -= sizeof(k.data);
+                        continue;
+                    }
+                }
+                if (!corrupted_.exchange(true, std::memory_order_relaxed)) {
+                    CASPAR_LOG(error) << "ffmpeg producer: SubstitutedA53CCQueue: closed captions corrupted by ffmpeg, "
+                                         "removing them.";
+                    break;
+                }
```
dedicatedbroadcastsolutions: Needs extra break here to prevent infinite loop, or some other fix.

programmerjake:
> I integrated this into a test branch with AI filling some of the gaps. It identified an infinite loop that was causing a bug during my testing where playback would stall.
> 
> I wasn't sure how to make inline change requests with that tool on here without seeing the diff so I had it make a pull request against the branch this is based on.
> 
> [programmerjake#3](https://github.com/programmerjake/casparcg-server/pull/3)

I pushed a fix crediting you for finding it: [652780a85](https://github.com/CasparCG/server/pull/1637/changes/652780a85e616e76d04cf5420cd635a92927ede4)

PR #1642, opened 2025-06-12
Decklink consumer VANC

niklaspandersson:
Add support for adding VANC packets to decklink frames

* Separate Vanc class that handles vanc related logic
* Strategy based approach to isolate different types of ancilliary data 
* Only support for a push based workflow atm (via amcp APPLY, se separate PR #1638 ), but should be fairly easy to extend with support for data arriving from other sources (ie. the producer/consumer-pipeline)

### DIFF
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -1058,6 +1058,16 @@ struct decklink_consumer final : public IDeckLinkVideoOutputCallback
                     CASPAR_LOG(error) << print() << L" Failed to add ancillary packet.";
                 }
             }
+
+            bool isInterlaced = mode_->GetFieldDominance() != bmdProgressiveFrame;
+            if (isInterlaced) {
+                auto field2_packets = vanc_->pop_packets(true);
+                for (auto& packet : field2_packets) {
+                    if (FAILED(ancillary_packets->AttachPacket(get_raw(packet)))) {
+                        CASPAR_LOG(error) << print() << L" Failed to add ancillary packet.";
+                    }
+                }
+            }
         }
         if (FAILED(output_->ScheduleVideoFrame(
                 get_raw(fill_frame), display_time, decklink_format_desc_.duration, decklink_format_desc_.time_scale))) {
--- a/src/modules/decklink/consumer/decklink_consumer.cpp
+++ b/src/modules/decklink/consumer/decklink_consumer.cpp
@@ -27,6 +27,7 @@
 #include "decklink_consumer.h"
 #include "format_strategy.h"
 #include "monitor.h"
+#include "vanc.h"
 
 #include "../util/util.h"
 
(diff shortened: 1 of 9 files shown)


src/modules/decklink/consumer/decklink_consumer.cpp
```diff
@@ -264,6 +273,9 @@ class decklink_frame
         } else if (hdr_ && std::memcmp(&iid, &IID_IDeckLinkVideoFrameMetadataExtensions, sizeof(REFIID)) == 0) {
             *ppv = static_cast<IDeckLinkVideoFrameMetadataExtensions*>(this);
             AddRef();
+        } else if (vanc_ && std::memcmp(&iid, &IID_IDeckLinkVideoFrameAncillaryPackets, sizeof(REFIID)) == 0) {
```
programmerjake, line 276: this has to return a pointer to the same class, not a separate class. returning a pointer to an independent class violates COM's guarantees for `IUnknown::QueryInterface`:
https://learn.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void)#remarks

> - It must be symmetric—if a client holding a pointer to one interface queries successfully for another, then a query through the obtained pointer for the first interface must succeed.

returning a separate `vanc_` object violates that because querying the returned `vanc_` object for e.g. `IDeckLinkVideoFrame` will fail.

programmerjake: the basic idea is that `QueryInterface` only ever returns pointers to this class'es base classes.

src/modules/decklink/consumer/decklink_consumer.cpp
```diff
@@ -406,13 +418,11 @@ class decklink_frame
         return E_INVALIDARG;
     }
 
-    /*
     HRESULT STDMETHODCALLTYPE GetBytes(BMDDeckLinkFrameMetadataID metadataID, void* buffer, uint32_t* bufferSize)
```
programmerjake, line 421:
```suggestion
    HRESULT STDMETHODCALLTYPE GetBytes(BMDDeckLinkFrameMetadataID metadataID,
                                       void*                      buffer,
                                       uint32_t*                  bufferSize) override
```
missing `override`

src/modules/decklink/consumer/vanc_op47_strategy.cpp
```diff
@@ -0,0 +1,130 @@
...
+        : line_number_(line_number)
+        , line_number_2_(line_number_2)
+        , sd_line_(21)
+        , counter_(1)
+        , dummy_header_(dummy_header.empty() ? std::vector<uint8_t>() : base64_decode(dummy_header))
+    {
+    }
+
+    virtual bool        has_data() const { return !packets_.empty() || !dummy_header_.empty(); }
+    virtual vanc_packet pop_packet(bool field2)
```
programmerjake, line 41:
```suggestion
    virtual bool        has_data() const override { return !packets_.empty() || !dummy_header_.empty(); }
    virtual vanc_packet pop_packet(bool field2) override
```
missing `override` for all virtual methods here, also in `vanc_scte104_stategy.cpp`

programmerjake: oh, also you can't read from `packets_` without locking `mutex_`.

src/modules/decklink/consumer/vanc_scte104_strategy.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    std::vector<uint8_t> payload_ = {};
+
+  public:
+    explicit vanc_scte104_strategy(uint8_t line_number)
+        : line_number_(line_number)
+    {
+    }
+
+    virtual bool        has_data() const { return !payload_.empty(); }
```
programmerjake, line 26: you also need to lock `mutex_` here.

TomKaltz:
@niklaspandersson what's the status of this?

src/modules/decklink/util/base64.hpp
Julusian, line 1: we already have a base64 implementation in src/common/base64.h/cpp (some wrapping around boost), do we need another?

If this new one is superior, perhaps it should replace the other one?

Julusian (approved):

PR #1650, opened 2025-06-26
implement layer level LENGTH override to support AUTO play for all producers

### LINKED ISSUE #1644: RFC: auto play for infinite length producers
### Description

Using LOADBG AUTO is the most reliable way to get frame-accurate playlist functionality.  For infinite length producers (FFMPEG with LOOP, ROUTE, DECKLINK, IMAGE, etc) AUTO is not an option.  It would be very useful if the user could set the auto play length for any producer...even an infinite length one.

### Solution suggestion

I see two options for implementing this...

1. modify all producers to honor a length property for auto play purposes (even ffmpeg with loop), or
2. Add a new AMCP token to set the auto play duration in LOADBG

What do you all think?

TomKaltz:
This PR extends the existing LENGTH parameter to work with infinite length producers (decklink, bluefish, looping content, etc.) for autoplay functionality. Previously, infinite length producers couldn't use timed autoplay since their nb_frames() returns infinity. With this change, users can now specify LENGTH with any producer to control autoplay timing, enabling commands like LOADBG 1-10 decklink DEVICE 1 LENGTH 500 AUTO or LOADBG 1-10 video.mp4 LOOP LENGTH 300 AUTO to automatically transition after the specified frame count.

The implementation tracks separate length overrides for foreground and background producers at the layer level. When a LENGTH parameter is provided in AMCP commands, the override is stored and used in autoplay calculations instead of the producer's natural nb_frames(). Length overrides transfer correctly during play transitions and are properly cleared/replaced with subsequent load operations, preserving all existing functionality while cleanly extending autoplay support to infinite time producers.

## notable autoplay behavior
- If a producer's actual length is shorter than the supplied LENGTH (ffmpeg producer with short file or IN, OUT creating a shorter than LENGTH clip), then the last frame is repeated until LENGTH frame count is achieved.
- You can use IN, OUT, LOOP and LENGTH together with ffmpeg producer to create a custom looping clip that will loop until  LENGTH frame count is achieved.

closes #1644


dimitry-ishenko:
Sorry for jumping in late, but perhaps `TOTAL_LENGH` or `OVERALL_LENGTH` would be better names?

src/core/producer/layer.cpp
```diff
@@ -101,6 +127,8 @@ struct layer::impl
     {
         foreground_ = frame_producer::empty();
         auto_play_  = false;
+        foreground_auto_length_override_.reset();
+        background_auto_length_override_.reset();
```
Julusian: I don't think this is correct, as we aren't disposing of background_
```suggestion
```

src/core/producer/layer.cpp
```diff
@@ -159,6 +189,66 @@ struct layer::impl
             return draw_frame{};
         }
     }
+
+    std::future<std::wstring> call(const std::vector<std::wstring>& params)
```
Julusian: All of the AMCP responses from this are unusual. They say things like `AUTO PLAY NOT ACTIVE` which will be annoying to parse in a library in any kind of generic way.
It would be better to simply return `FRAMES_LEFT TRUE 123`. While less human readable, it will be much easier to parse without needing to consider every route through the code.

src/core/producer/stage.h
```diff
@@ -82,6 +83,11 @@ class stage_base
     virtual std::future<frame_transform> get_current_transform(int index)                                          = 0;
     virtual std::future<void>
     load(int index, const spl::shared_ptr<frame_producer>& producer, bool preview = false, bool auto_play = false) = 0;
```
Julusian: It would be nicer to replace these other methods, instead of adding more with extra parameters.
The new parameter could have a default value of none, so to the couple of callsites that havent been touched in this PR could continue that way

Julusian (changes requested):

PR #1651, opened 2025-06-26
feat: use gpu texture in screen consumer

niklaspandersson:
* Add texture property to const_frame that references the texture on the gpu
* Replace GPU-CPU-GPU copy with GPU texture in screen consumer
* With this, a screen consumer is more or less free if used in combination with any other consumer

### DIFF
--- a/src/modules/screen/consumer/screen_consumer.cpp
+++ b/src/modules/screen/consumer/screen_consumer.cpp
@@ -111,15 +111,29 @@ struct configuration
     bool            always_on_top = false;
     colour_spaces   colour_space  = colour_spaces::RGB;
     bool            high_bitdepth = false;
+    bool            gpu_texture   = false;
 };
 
 struct frame
 {
-    GLuint pbo   = 0;
-    GLuint tex   = 0;
-    char*  ptr   = nullptr;
-    GLsync fence = nullptr;
+    GLuint                         pbo = 0;
+    GLuint                         tex = 0;
+    char*                          ptr = nullptr;
+    std::shared_ptr<core::texture> texture;
+    GLsync                         fence = nullptr;
 };
+
+struct screen_consumer;
+
+struct display_strategy
+{
+    virtual ~display_strategy() {}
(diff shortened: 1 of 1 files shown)


niklaspandersson:
Doesn't work on Linux as is, because on linux we use different opengl "wrappers" in the accelerator and screen consumer.
If I build the accelerator using the sfml context like on windows (just removing the conditions in the accelerator cmakelists file, it works. 

@Julusian, why do we have a separate context (using EGL) on Linux? I'm guessing there is a reason...

Julusian:
That was a recent change to support environments where X11 is not available https://github.com/CasparCG/server/pull/1624

But maybe that can be reshuffled to choose at runtime based on whether it is found to be available rather than being hardcoded this way.

niklaspandersson:
> That was a recent change to support environments where X11 is not available #1624
> 
> But maybe that can be reshuffled to choose at runtime based on whether it is found to be available rather than being hardcoded this way.

Yea, I just had a look in the commit history and realized I should have looked there before asking :)

A runtime check sounds like the correct fix. I'll have a go at that...

niklaspandersson:
I think this works now. I'm on vacation and haven't got access to my linux workstation. It runs fine(ish) in WSL, and selects context implementation based on the presens of a DISPLAY env var. 

The screen consumer actually works in WSL now, with the sfml context. ...when using the EGL context it crashes in WSL, but that is not specific to this PR.

TomKaltz:
@niklaspandersson what issues were you having?

niklaspandersson:
> @niklaspandersson what issues were you having?

My client had issues with glitching when he used this together with ndi- and decklink producers. I haven't been able to reproduce, but I've seen a video. 

I just noticed that I was missing a fence to protect the gpu textures while rendering. I'll have him try it again now...

Julusian:
My only question here is whether there is value in keeping the host mode or if this should be simplified to just the gpu mode?
And as part of that, should this gpu mode be the default?

I shall merge it now as is, but something to consider and fix in a follow up.

PR #1652, opened 2025-06-29
feat: decklink support for chosing yuv output i 8-bit mode

niklaspandersson:
Enables users to opt-in to using yuv (v210) instead of rgba in the decklink consumer on standard 8-bit channels. This enables the maximum usage of the decklink cards output bandwidth.


Julusian:
I think this is going to be important, as the blackmagic 2110 ecosystem is pretty picky and many receivers require 10bit even if their specs claim to support 8bit (and the decklink auto converts rgb to 8bit yuv)

PR #1654, opened 2025-07-04
add websocket servers for amcp and monitor data

TomKaltz:
WIP implement websocket server using boost beast for AMCP and monitoring

Exposes 2 websocket ports…One for AMCP…the other port for monitoring.  When a client connects to the monitor websocket, the full state is sent and then subsequent updates will be the charge deltas in RFC6902 format.

The monitor client can request a full_state message by sending {“command”:”request_full_state”}

### DIFF
--- a/src/protocol/util/websocket_server.cpp
+++ b/src/protocol/util/websocket_server.cpp
@@ -1,0 +1,525 @@
+/*
+ * Copyright (c) 2024 CasparCG (www.casparcg.com).
+ *
+ * This file is part of CasparCG (www.casparcg.com).
+ *
+ * CasparCG is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * CasparCG is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with CasparCG. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "../StdAfx.h"
+
+#include "websocket_server.h"
+
+#include <common/log.h>
+#include <common/utf.h>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/beast/core.hpp>
+#include <boost/beast/websocket.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include <map>
+#include <mutex>
+#include <set>
+#include <sstream>
+#include <thread>
+
+namespace beast     = boost::beast;
+namespace http      = beast::http;
+namespace websocket = beast::websocket;
+namespace net       = boost::asio;
+using tcp           = net::ip::tcp;
+
+namespace caspar { namespace IO {
+
+// Convert monitor state to JSON string
+std::string monitor_state_to_json(const core::monitor::state& state)
+{
+    boost::property_tree::wptree json_tree;
+
+    // Add timestamp
+    json_tree.put(
+        L"timestamp",
+        std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())
+            .count());
+
+    // Convert monitor state to JSON structure
+    // The monitor state contains various performance and status information
+    boost::property_tree::wptree channels_tree;
+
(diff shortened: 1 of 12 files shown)


Julusian:
I haven't read the code or thought about this much yet, but a couple of ideas:

Perhaps this should operate in a way of letting the client say what they want to listen to (and perhaps potentially the format of the data?)  
That way a client can listen to only the channels it is interested in, and optionally turn on/off noisier messages. I don't know how granular this should be, there are a few ways this could be approached.
In a follow up change it could allow for opt into the full diag data needed for building the diag window.

And while I like that this is using json-patch, I wonder if that might be frustrating for some uses because of the extra complexity needed (while this works well for something like js, I dont know how well supported it is in other languages). so it would be nice to be able to specify this, perhaps as part of some 'subscribe' calls the client makes. json-patch can be the default though.

ronag (approved):
LGTM

Though I think the patching is unecessary. It's not so much data. Why not just always send the whole state and the server diff and apply changes?

ronag:
Another thing. Add a server timestamp to every update.

TomKaltz:
> @TomKaltz just curious, is there a problem with using OSC?

@dimitry-ishenko I've used OSC very reliably for many years with great success but there were things that always bothered me.  One big issue I have is that you can't immediately definitively know when a stage is empty/cleared since that stage will just stop producing OSC messages.  This required awkward checks with timeouts to detect when a stage/layer stopped producing data.  With this implementation, as long as you aren't using a wildcard for the layer subscriptions, the layer will transmit ```producer: "empty"``` immediately when a stage/layer is cleared.

The advantages I see to adding websockets are as follows...
- more reliable tcp transport (udp data delivery is not guaranteed)
- immediate cleared/empty stage status when subscribing to specific layers
- support for control and realtime status for web browser clients (or html cg templates)
- bandwidth reduction due to optional per message compression and/or path subscription model

I'm hoping hoping this PR will add value to CasparCG Server.  I always suspected it would.  I will be testing it heavily with a client project I'm working on right now so I hope to get it as bullet-proof as possible ASAP.

sirfnomi:
Hi @TomKaltz,

First of all, thank you for your work on this pull request!

I have a question regarding the scope of the information that will be exposed via this new WebSocket. Beyond general server status, would it be possible to include more granular internal state information for individual layer? Specifically, I'm referring to states such as:

* **Loading:** When a clip is being loaded into server.
* **Loaded:** When a clip is loaded into server and ready to play.
* **Playing:** When a clip is actively being played out.
* **Paused:** If a clip is currently paused.
* **Stopped:** When a clip has finished playing or has been explicitly stopped.

Currently, when developing clients that interact with CasparCG, determining the precise state of a clip using OSC is quite challenging. We often have to infer a clip's state through a combination of its name, time, and duration, and then manage multiple conditional checks on the client-side. This approach is prone to inaccuracies and adds significant complexity to client development.  **This is especially problematic when replaying the same clip or play another clip on same layer, as the server often initially sends outdated time and duration information for a brief period before providing the actual, current details of the clip.**

Exposing explicit clip states via the WebSocket would greatly simplify client-side logic and enable more robust and responsive UIs. It would eliminate the need for clients to "guess" the state and instead allow them to react directly to definitive state changes from the server.

Is this something that you've considered, or do you think it would be feasible to incorporate into this WebSocket implementation?

Thanks again for your efforts!

src/protocol/util/websocket_server.h
```diff
@@ -0,0 +1,76 @@
...
+
+namespace caspar { namespace IO {
+
+/**
+ * WebSocket server that handles both AMCP protocol commands and monitor data streaming.
+ */
+class websocket_server
+{
+  public:
+    // Constructor for AMCP + Monitor (legacy)
```
Julusian, line 45: a 'legacy' and a 'new' in some new code? sounds like one of these can be removed

src/protocol/util/websocket_server.cpp
```diff
@@ -0,0 +1,541 @@
...
+    std::shared_ptr<boost::asio::io_context> context_;
+    protocol_strategy_factory<wchar_t>::ptr  amcp_protocol_factory_;
+    uint16_t                                 amcp_port_;
+    uint16_t                                 monitor_port_;
+
+    tcp::acceptor amcp_acceptor_;
+    tcp::acceptor monitor_acceptor_;
+
+    std::set<spl::shared_ptr<websocket_monitor_session>>                                           monitor_sessions_;
+    std::vector<std::function<std::pair<std::wstring, std::shared_ptr<void>>(const std::string&)>> lifecycle_factories_;
```
Julusian, line 375: as I said to some of this in server.cpp, I think a lot of this lifecycle logic can be removed

TomKaltz:
@Julusian I marked this as draft once again.  There was almost a complete rewrite in my fork and I haven't had time to contribute the work back here.  For now this PR is in a holding pattern until I can devote time into the cleanup effort.

PR #1663, opened 2025-08-14
Add possibility to bind AMCP port to specific IP address

nerijus:
Currently AMCP port 5250 listens on all interfaces:
```
ss -nlp|grep 5250
tcp   LISTEN 0      4096                                              0.0.0.0:5250             0.0.0.0:*    users:(("casparcg",pid=3534544,fd=112))
```
So it is possible for everyone from internet to connect to it and send commands (unless it is firewalled).
With this PR it is possible to specify IP addr like this:
```
--- casparcg.config.orig	2025-07-18 15:26:29.000000000 +0000
+++ casparcg.config	2025-08-14 15:31:52.387490375 +0000
@@ -19,6 +19,7 @@
     </channels>
     <controllers>
         <tcp>
+            <host>127.0.0.1</host>
             <port>5250</port>
             <protocol>AMCP</protocol>
         </tcp>
```
Now it listens on `127.0.0.1` only:
```
ss -nlp|grep 5250
tcp   LISTEN 0      4096                                            127.0.0.1:5250             0.0.0.0:*    users:(("casparcg",pid=965101,fd=113))
```
This PR was made with the help of AI. If `<host>127.0.0.1</host>` is omitted, then it behaves like before. The value should be IP addr, not hostname.


src/protocol/util/AsyncEventServer.h
```diff
@@ -38,7 +38,13 @@ class AsyncEventServer
   public:
     explicit AsyncEventServer(std::shared_ptr<boost::asio::io_context>    io_context,
                               const protocol_strategy_factory<char>::ptr& protocol,
+                              const std::string&                          host,
                               unsigned short                              port);
+
+    AsyncEventServer(std::shared_ptr<boost::asio::io_context>             io_context,
```
Julusian, line 44: I dont think this constructor without the host is being used or is helpful, and could be removed

src/protocol/util/AsyncEventServer.cpp
```diff
@@ -280,9 +280,34 @@ struct AsyncEventServer::implementation : public spl::enable_shared_from_this<im
...
+                  CASPAR_LOG(fatal) << "Invalid host address: " << host << " - " << e.what();
+                  throw;
+              }
+          }
+          tcp::acceptor a(*io_context_);
+          a.open(endpoint.protocol());
+          a.set_option(boost::asio::ip::tcp::acceptor::reuse_address(true));
+          a.bind(endpoint);
+          a.listen();
+          return a;
```
Julusian, line 309: is there a reason for this constructor being more verbose than before? 
Previously it was doing `acceptor_(*io_context_, tcp::endpoint(tcp::v4(), port))`, so it  looks like this could be done with something along the lines of `return tcp::acceptor(*io_context_, endpoint);` instead?

Julusian (changes requested):
I was going to say that it would also be good to have this documented in the examples block in the config file, but it appears that this section of config is not expressed there, only in the active portion of the file, so maybe not

PR #1668, opened 2025-09-24
fix: remove unnecessary sws scale forcing conversion to YUVA422

deadbeef84:
There's no need to convert pixel format before we pass frames to ffmpeg, it should handle conversion as necessary. If your output pixel format is not YUVA422 ffmpeg would make an additional conversion. With this change, a single conversion should be possible.

It may have been required at some point to do parallel slices, but that should no longer be necessary as ffmpeg/swscale should be able to do that natively.


Julusian:
Looks like it came from 7b94bc6544b620583263bee411f88be99ab6eda2 originally, but I am willing to try removing it to find out if it is still needed

PR #1677, opened 2025-10-19
Vulkan

niklaspandersson:
Initial support for Vulkan

* Drop in replacement of the opengl accelerator
* Image mixing should be on feature parity with the opengl image mixer
* Very limited compatibility and performance testing has been done
* Support for configuring which accelerator backend to use (defaults to opengl on linux and windows)
* New feature flag "ENABLE_VULKAN" (defaults to OFF on linux and windows) to  opt in to building with vulkan support (since it requires new build dependencies)
* ~~Initial support for macOS~~
* Very limited updates of the BUILDING documentation

### DIFF
--- a/src/CMakeModules/Bootstrap_Linux.cmake
+++ b/src/CMakeModules/Bootstrap_Linux.cmake
@@ -40,6 +40,24 @@ find_package(SFML 3 COMPONENTS Graphics System Window QUIET)
 if(NOT SFML_FOUND)
     find_package(SFML 2 COMPONENTS graphics system window REQUIRED)
 endif()
+
+find_package(Vulkan REQUIRED)
+
+FetchContent_Declare(vk_bootstrap
+        URL ${CASPARCG_DOWNLOAD_MIRROR}/vk-bootstrap/vk-bootstrap-1.4.328.tar.gz
+        URL_HASH SHA256=3be0220de218dc3e692aeac552b2953860a0e0a48257f4a61c3f1c1472674744
+        DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+        )
+FetchContent_MakeAvailable(vk_bootstrap)
+
+FetchContent_Declare(vma
+        URL ${CASPARCG_DOWNLOAD_MIRROR}/VulkanMemoryAllocator/VulkanMemoryAllocator-3.3.0.tar.gz
+        URL_HASH SHA256=c4f6bbe6b5a45c2eb610ca9d231158e313086d5b1a40c9922cb42b597419b14e
+        DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+)
+FetchContent_MakeAvailable(vma)
+
+
 find_package(X11 REQUIRED)
 
 if (ENABLE_HTML)
@@ -139,7 +157,7 @@ ENDIF ()
 
 ADD_COMPILE_OPTIONS (-fnon-call-exceptions) # Allow signal handler to throw exception
 
-ADD_COMPILE_OPTIONS (-Wno-deprecated-declarations -Wno-write-strings -Wno-multichar -Wno-cpp -Werror)
+ADD_COMPILE_OPTIONS (-Wno-deprecated-declarations -Wno-write-strings -Wno-multichar -Wno-cpp -Werror -Wno-nonnull -Wno-nullability-completeness)
 IF (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     ADD_COMPILE_OPTIONS (-Wno-terminate)
 ELSEIF (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
@@ -149,3 +167,5 @@ ELSEIF (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     message(STATUS "ADDING: -DTBB_USE_GLIBCXX_VERSION=${TBB_USE_GLIBCXX_VERSION}")
     add_definitions(-DTBB_USE_GLIBCXX_VERSION=${TBB_USE_GLIBCXX_VERSION})
 ENDIF ()
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
--- a/src/CMakeModules/Bootstrap_Linux.cmake
+++ b/src/CMakeModules/Bootstrap_Linux.cmake
@@ -16,6 +16,7 @@ set(USE_STATIC_BOOST OFF CACHE BOOL "Use shared library version of Boost")
 set(USE_SYSTEM_CEF ON CACHE BOOL "Use the version of cef from your OS (only tested with Ubuntu)")
 set(CASPARCG_BINARY_NAME "casparcg" CACHE STRING "Custom name of the binary to build (this disables some install files)")
 set(ENABLE_AVX2 OFF CACHE BOOL "Enable the AVX2 instruction set (requires a CPU that supports it)")
+set(ENABLE_VULKAN OFF CACHE BOOL "Enable Vulkan support")
 
 # Determine build (target) platform
 SET (PLATFORM_FOLDER_NAME "linux")
@@ -41,21 +42,23 @@ if(NOT SFML_FOUND)
     find_package(SFML 2 COMPONENTS graphics system window REQUIRED)
 endif()
 
-find_package(Vulkan REQUIRED)
+IF (ENABLE_VULKAN)
+    find_package(Vulkan REQUIRED)
 
-FetchContent_Declare(vk_bootstrap
-        URL ${CASPARCG_DOWNLOAD_MIRROR}/vk-bootstrap/vk-bootstrap-1.4.328.tar.gz
-        URL_HASH SHA256=3be0220de218dc3e692aeac552b2953860a0e0a48257f4a61c3f1c1472674744
-        DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
-        )
-FetchContent_MakeAvailable(vk_bootstrap)
+    FetchContent_Declare(vk_bootstrap
+            URL ${CASPARCG_DOWNLOAD_MIRROR}/vk-bootstrap/vk-bootstrap-1.4.328.tar.gz
+            URL_HASH SHA256=3be0220de218dc3e692aeac552b2953860a0e0a48257f4a61c3f1c1472674744
+            DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+            )
+    FetchContent_MakeAvailable(vk_bootstrap)
 
-FetchContent_Declare(vma
-        URL ${CASPARCG_DOWNLOAD_MIRROR}/VulkanMemoryAllocator/VulkanMemoryAllocator-3.3.0.tar.gz
-        URL_HASH SHA256=c4f6bbe6b5a45c2eb610ca9d231158e313086d5b1a40c9922cb42b597419b14e
-        DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
-)
-FetchContent_MakeAvailable(vma)
+    FetchContent_Declare(vma
+            URL ${CASPARCG_DOWNLOAD_MIRROR}/VulkanMemoryAllocator/VulkanMemoryAllocator-3.3.0.tar.gz
+            URL_HASH SHA256=c4f6bbe6b5a45c2eb610ca9d231158e313086d5b1a40c9922cb42b597419b14e
+            DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+    )
+    FetchContent_MakeAvailable(vma)
+endif()
 
 
 find_package(X11 REQUIRED)
@@ -157,7 +160,12 @@ ENDIF ()
 
 ADD_COMPILE_OPTIONS (-fnon-call-exceptions) # Allow signal handler to throw exception
 
-ADD_COMPILE_OPTIONS (-Wno-deprecated-declarations -Wno-write-strings -Wno-multichar -Wno-cpp -Werror -Wno-nonnull -Wno-nullability-completeness)
+ADD_COMPILE_OPTIONS (-Wno-deprecated-declarations -Wno-write-strings -Wno-multichar -Wno-cpp -Werror)
+
+IF (ENABLE_VULKAN)
+    ADD_COMPILE_OPTIONS (-Wno-nonnull -Wno-nullability-completeness)
+ENDIF()
+
 IF (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     ADD_COMPILE_OPTIONS (-Wno-terminate)
 ELSEIF (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/src/CMakeModules/Bootstrap_Windows.cmake
+++ b/src/CMakeModules/Bootstrap_Windows.cmake
@@ -151,6 +151,22 @@ target_link_directories(GLEW::glew INTERFACE ${glew_SOURCE_DIR}/lib/Release/x64)
 target_link_libraries(GLEW::glew INTERFACE glew32)
 casparcg_add_runtime_dependency("${glew_SOURCE_DIR}/bin/Release/x64/glew32.dll")
 
+find_package(Vulkan REQUIRED)
+
+FetchContent_Declare(vk_bootstrap
+		URL ${CASPARCG_DOWNLOAD_MIRROR}/vk-bootstrap/vk-bootstrap-1.4.328.zip
+		URL_HASH SHA256=10f257c30a0a49d30b28a72cf3a7942d93a61f977adaa04bee29304c6506dc12
+		DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+		)
+FetchContent_MakeAvailable(vk_bootstrap)
+
+FetchContent_Declare(vma
+		URL ${CASPARCG_DOWNLOAD_MIRROR}/VulkanMemoryAllocator/VulkanMemoryAllocator-3.3.0.zip
+		URL_HASH SHA256=81755d8fcb411b97292c6682e828501315db319374c7c34ba6e1226452c6c392
+		DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+)
+FetchContent_MakeAvailable(vma)
+
 # SFML
 FetchContent_Declare(sfml
 	URL ${CASPARCG_DOWNLOAD_MIRROR}/sfml/SFML-2.6.2-windows-vc17-64-bit.zip
--- a/src/CMakeModules/Bootstrap_Windows.cmake
+++ b/src/CMakeModules/Bootstrap_Windows.cmake
@@ -12,6 +12,7 @@ if(POLICY CMP0167)
 endif()
 
 set(BOOST_USE_PRECOMPILED ON CACHE BOOL "Use precompiled boost")
+set(ENABLE_VULKAN OFF CACHE BOOL "Enable Vulkan support")
 
 set(CASPARCG_RUNTIME_DEPENDENCIES_RELEASE "" CACHE INTERNAL "")
 set(CASPARCG_RUNTIME_DEPENDENCIES_DEBUG "" CACHE INTERNAL "")
@@ -151,21 +152,23 @@ target_link_directories(GLEW::glew INTERFACE ${glew_SOURCE_DIR}/lib/Release/x64)
 target_link_libraries(GLEW::glew INTERFACE glew32)
 casparcg_add_runtime_dependency("${glew_SOURCE_DIR}/bin/Release/x64/glew32.dll")
 
-find_package(Vulkan REQUIRED)
+IF(ENABLE_VULKAN)
+	find_package(Vulkan REQUIRED)
 
-FetchContent_Declare(vk_bootstrap
-		URL ${CASPARCG_DOWNLOAD_MIRROR}/vk-bootstrap/vk-bootstrap-1.4.328.zip
-		URL_HASH SHA256=10f257c30a0a49d30b28a72cf3a7942d93a61f977adaa04bee29304c6506dc12
-		DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
-		)
-FetchContent_MakeAvailable(vk_bootstrap)
+	FetchContent_Declare(vk_bootstrap
+			URL ${CASPARCG_DOWNLOAD_MIRROR}/vk-bootstrap/vk-bootstrap-1.4.328.zip
+			URL_HASH SHA256=10f257c30a0a49d30b28a72cf3a7942d93a61f977adaa04bee29304c6506dc12
+			DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+			)
+	FetchContent_MakeAvailable(vk_bootstrap)
 
-FetchContent_Declare(vma
-		URL ${CASPARCG_DOWNLOAD_MIRROR}/VulkanMemoryAllocator/VulkanMemoryAllocator-3.3.0.zip
-		URL_HASH SHA256=81755d8fcb411b97292c6682e828501315db319374c7c34ba6e1226452c6c392
-		DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
-)
-FetchContent_MakeAvailable(vma)
+	FetchContent_Declare(vma
+			URL ${CASPARCG_DOWNLOAD_MIRROR}/VulkanMemoryAllocator/VulkanMemoryAllocator-3.3.0.zip
+			URL_HASH SHA256=81755d8fcb411b97292c6682e828501315db319374c7c34ba6e1226452c6c392
+			DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
+	)
+	FetchContent_MakeAvailable(vma)
+ENDIF()
 
 # SFML
 FetchContent_Declare(sfml
--- a/src/accelerator/vulkan/image/fragment_shader.frag
+++ b/src/accelerator/vulkan/image/fragment_shader.frag
@@ -1,0 +1,578 @@
+#version 450
+
+#extension GL_ARB_fragment_shader_interlock : enable
+
+layout(location = 0) in vec4 TexCoord;
+layout(location = 1) in vec4 TexCoord2;
+
+layout(location = 0) out vec4 fragColor;
+
+layout(binding = 0) uniform sampler2D	plane[4];
+layout(input_attachment_index = 0, binding = 1) uniform subpassInput background;
+layout(binding = 2) uniform sampler2D local_key;
+layout(binding = 3) uniform sampler2D layer_key;
+
+
+const uint is_straight_alpha_mask = 1u << 0;
+const uint has_local_key_mask = 1u << 1;
+const uint has_layer_key_mask = 1u << 2;
+const uint invert_mask = 1u << 3;
+const uint levels_mask = 1u << 4;
+const uint csb_mask = 1u << 5;
+const uint chroma_mask = 1u << 6;
+const uint chroma_show_mask_mask = 1u << 7;
+
+layout(push_constant) uniform ParamsBlock {
+    uint color_space_index;
+    float precision_factor[4];
+    int blend_mode;
+    int keyer;
+    int pixel_format;
+    float opacity;
+
+/* levels */
+    float min_input;
+    float max_input;
+    float gamma;
+    float min_output;
+    float max_output;
+    
+/* contrast, saturation & brightness */
+    float brt;
+    float sat;
+    float con;
+
+/* Chroma */
+    float chroma_target_hue;
+    float chroma_hue_width;
+    float chroma_min_saturation;
+    float chroma_min_brightness;
+    float chroma_softness;
+    float chroma_spill_suppress;
+    float chroma_spill_suppress_saturation;
+
+    uint flags;
+};
+
+bool is_straight_alpha = (flags & is_straight_alpha_mask) == is_straight_alpha_mask;
+bool has_local_key = (flags & has_local_key_mask) == has_local_key_mask;
+bool has_layer_key  = (flags & has_layer_key_mask) == has_layer_key_mask;
+bool invert = (flags & invert_mask) == invert_mask;
+bool levels = (flags & levels_mask) == levels_mask;
+bool csb = (flags & csb_mask) == csb_mask;
+bool chroma = (flags & chroma_mask) == chroma_mask;
+bool chroma_show_mask = (flags & chroma_show_mask_mask) == chroma_show_mask_mask;
+
+const mat3[3] color_matrices = mat3[3](
+                    mat3(1.0, 0.0, 1.402, 1.0, -0.344, -0.509, 1.0, 1.772, 0.0),
+                    mat3(1.0, 0.0, 1.5748, 1.0, -0.1873, -0.4681, 1.0, 1.8556, 0.0),
+                    mat3(1.0, 0.0, 1.4746, 1.0, -0.16455312684366, -0.57135312684366, 1.0, 1.8814, 0.0)
+                ); 
+const vec3[3] luma_coefficients = vec3[3](
+                    vec3(0.299, 0.587, 0.114), // Rec. 601
+                    vec3(0.2126, 0.7152, 0.0722),  // Rec. 709
+                    vec3(0.2627, 0.6780, 0.0593)   // Rec. 2020
+                );
+
+mat3 color_matrix = transpose(color_matrices[color_space_index]);
+vec3 luma_coeff = luma_coefficients[color_space_index];
+
+/*
+** Contrast, saturation, brightness
+** Code of this function is from TGM's shader pack
+** http://irrlicht.sourceforge.net/phpBB2/viewtopic.php?t=21057
+*/
+
+vec3 ContrastSaturationBrightness(vec4 color, float brt, float sat, float con)
+{
+    const float AvgLumR = 0.5;
+    const float AvgLumG = 0.5;
+    const float AvgLumB = 0.5;
+
+    if (color.a > 0.0)
+        color.rgb /= color.a;
+
+    vec3 AvgLumin = vec3(AvgLumR, AvgLumG, AvgLumB);
+    vec3 brtColor = color.rgb * brt;
+    vec3 intensity = vec3(dot(brtColor, luma_coeff));
+    vec3 satColor = mix(intensity, brtColor, sat);
+    vec3 conColor = mix(AvgLumin, satColor, con);
+
+    conColor.rgb *= color.a;
+
+    return conColor;
+}
+
+/*
+** Gamma correction
+** Details: http://blog.mouaif.org/2009/01/22/photoshop-gamma-correction-shader/
+*/
+#define GammaCorrection(color, gamma)								pow(color, vec3(1.0 / gamma))
+
+/*
+** Levels control (input (+gamma), output)
+** Details: http://blog.mouaif.org/2009/01/28/levels-control-shader/
+*/
+
+#define LevelsControlInputRange(color, minInput, maxInput)				min(max(color - vec3(minInput), vec3(0.0)) / (vec3(maxInput) - vec3(minInput)), vec3(1.0))
+#define LevelsControlInput(color, minInput, gamma, maxInput)				GammaCorrection(LevelsControlInputRange(color, minInput, maxInput), gamma)
+#define LevelsControlOutputRange(color, minOutput, maxOutput) 			mix(vec3(minOutput), vec3(maxOutput), color)
+#define LevelsControl(color, minInput, gamma, maxInput, minOutput, maxOutput) 	LevelsControlOutputRange(LevelsControlInput(color, minInput, gamma, maxInput), minOutput, maxOutput)
+
+/*
+** Photoshop & misc math
+** Blending modes, RGB/HSL/Contrast/Desaturate, levels control
+**
+** Romain Dura | Romz
+** Blog: http://blog.mouaif.org
+** Post: http://blog.mouaif.org/?p=94
+*/
+
+
+/*
+** Desaturation
+*/
+
+vec4 Desaturate(vec3 color, float Desaturation)
+{
+    vec3 grayXfer = vec3(0.3, 0.59, 0.11);
+    vec3 gray = vec3(dot(grayXfer, color));
+    return vec4(mix(color, gray, Desaturation), 1.0);
+}
+
+
+/*
+** Hue, saturation, luminance
+*/
+
+vec3 RGBToHSL(vec3 color)
+{
+    vec3 hsl;
+
+    float fmin = min(min(color.r, color.g), color.b);
+    float fmax = max(max(color.r, color.g), color.b);
+    float delta = fmax - fmin;
+
+    hsl.z = (fmax + fmin) / 2.0;
+
+    if (delta == 0.0)
+    {
+        hsl.x = 0.0;
+        hsl.y = 0.0;
+    }
+    else
+    {
+        if (hsl.z < 0.5)
+            hsl.y = delta / (fmax + fmin);
+        else
+            hsl.y = delta / (2.0 - fmax - fmin);
+
+        float deltaR = (((fmax - color.r) / 6.0) + (delta / 2.0)) / delta;
+        float deltaG = (((fmax - color.g) / 6.0) + (delta / 2.0)) / delta;
+        float deltaB = (((fmax - color.b) / 6.0) + (delta / 2.0)) / delta;
+
+        if (color.r == fmax )
+            hsl.x = deltaB - deltaG;
+        else if (color.g == fmax)
+            hsl.x = (1.0 / 3.0) + deltaR - deltaB;
+        else if (color.b == fmax)
+            hsl.x = (2.0 / 3.0) + deltaG - deltaR;
+
+        if (hsl.x < 0.0)
+            hsl.x += 1.0;
+        else if (hsl.x > 1.0)
+            hsl.x -= 1.0;
+    }
+
+    return hsl;
+}
+
+float HueToRGB(float f1, float f2, float hue)
+{
+    if (hue < 0.0)
+        hue += 1.0;
+    else if (hue > 1.0)
+        hue -= 1.0;
+    float res;
+    if ((6.0 * hue) < 1.0)
+        res = f1 + (f2 - f1) * 6.0 * hue;
+    else if ((2.0 * hue) < 1.0)
+        res = f2;
+    else if ((3.0 * hue) < 2.0)
+        res = f1 + (f2 - f1) * ((2.0 / 3.0) - hue) * 6.0;
+    else
+        res = f1;
+    return res;
+}
+
+vec3 HSLToRGB(vec3 hsl)
+{
+    vec3 rgb;
+
+    if (hsl.y == 0.0)
+        rgb = vec3(hsl.z);
+    else
+    {
+        float f2;
+
+        if (hsl.z < 0.5)
+            f2 = hsl.z * (1.0 + hsl.y);
+        else
+            f2 = (hsl.z + hsl.y) - (hsl.y * hsl.z);
+
+        float f1 = 2.0 * hsl.z - f2;
+
+        rgb.r = HueToRGB(f1, f2, hsl.x + (1.0/3.0));
+        rgb.g = HueToRGB(f1, f2, hsl.x);
+        rgb.b= HueToRGB(f1, f2, hsl.x - (1.0/3.0));
+    }
+
+    return rgb;
+}
+
+
+
+
+/*
+** Float blending modes
+** Adapted from here: http://www.nathanm.com/photoshop-blending-math/
+** But I modified the HardMix (wrong condition), Overlay, SoftLight, ColorDodge, ColorBurn, VividLight, PinLight (inverted layers) ones to have correct results
+*/
+
+#define BlendLinearDodgef 					BlendAddf
+#define BlendLinearBurnf 					BlendSubstractf
+#define BlendAddf(base, blend) 				min(base + blend, 1.0)
+#define BlendSubstractf(base, blend) 		max(base + blend - 1.0, 0.0)
+#define BlendLightenf(base, blend) 		max(blend, base)
+#define BlendDarkenf(base, blend) 			min(blend, base)
+#define BlendLinearLightf(base, blend) 	(blend < 0.5 ? BlendLinearBurnf(base, (2.0 * blend)) : BlendLinearDodgef(base, (2.0 * (blend - 0.5))))
+#define BlendScreenf(base, blend) 			(1.0 - ((1.0 - base) * (1.0 - blend)))
+#define BlendOverlayf(base, blend) 		(base < 0.5 ? (2.0 * base * blend) : (1.0 - 2.0 * (1.0 - base) * (1.0 - blend)))
+#define BlendSoftLightf(base, blend) 		((blend < 0.5) ? (2.0 * base * blend + base * base * (1.0 - 2.0 * blend)) : (sqrt(base) * (2.0 * blend - 1.0) + 2.0 * base * (1.0 - blend)))
+#define BlendColorDodgef(base, blend) 		((blend == 1.0) ? blend : min(base / (1.0 - blend), 1.0))
+#define BlendColorBurnf(base, blend) 		((blend == 0.0) ? blend : max((1.0 - ((1.0 - base) / blend)), 0.0))
+#define BlendVividLightf(base, blend)		((blend < 0.5) ? BlendColorBurnf(base, (2.0 * blend)) : BlendColorDodgef(base, (2.0 * (blend - 0.5))))
+#define BlendPinLightf(base, blend) 		((blend < 0.5) ? BlendDarkenf(base, (2.0 * blend)) : BlendLightenf(base, (2.0 *(blend - 0.5))))
+#define BlendHardMixf(base, blend) 		((BlendVividLightf(base, blend) < 0.5) ? 0.0 : 1.0)
+#define BlendReflectf(base, blend) 		((blend == 1.0) ? blend : min(base * base / (1.0 - blend), 1.0))
+
+
+/*
+** Vector3 blending modes
+*/
+
+#define Blend(base, blend, funcf) 			vec3(funcf(base.r, blend.r), funcf(base.g, blend.g), funcf(base.b, blend.b))
+
+#define BlendNormal(base, blend) 			(blend)
+#define BlendLighten						BlendLightenf
+#define BlendDarken						BlendDarkenf
+#define BlendMultiply(base, blend) 		(base * blend)
+#define BlendAverage(base, blend) 			((base + blend) / 2.0)
+#define BlendAdd(base, blend) 				min(base + blend, vec3(1.0))
+#define BlendSubstract(base, blend) 		max(base + blend - vec3(1.0), vec3(0.0))
+#define BlendDifference(base, blend) 		abs(base - blend)
+#define BlendNegation(base, blend) 		(vec3(1.0) - abs(vec3(1.0) - base - blend))
+#define BlendExclusion(base, blend) 		(base + blend - 2.0 * base * blend)
+#define BlendScreen(base, blend) 			Blend(base, blend, BlendScreenf)
+#define BlendOverlay(base, blend) 			Blend(base, blend, BlendOverlayf)
+#define BlendSoftLight(base, blend) 		Blend(base, blend, BlendSoftLightf)
+#define BlendHardLight(base, blend) 		BlendOverlay(blend, base)
+#define BlendColorDodge(base, blend) 		Blend(base, blend, BlendColorDodgef)
+#define BlendColorBurn(base, blend) 		Blend(base, blend, BlendColorBurnf)
+#define BlendLinearDodge					BlendAdd
+#define BlendLinearBurn					BlendSubstract
+#define BlendLinearLight(base, blend) 		Blend(base, blend, BlendLinearLightf)
+#define BlendVividLight(base, blend) 		Blend(base, blend, BlendVividLightf)
+#define BlendPinLight(base, blend) 		Blend(base, blend, BlendPinLightf)
+#define BlendHardMix(base, blend) 			Blend(base, blend, BlendHardMixf)
+#define BlendReflect(base, blend) 			Blend(base, blend, BlendReflectf)
+#define BlendGlow(base, blend) 			BlendReflect(blend, base)
+#define BlendPhoenix(base, blend) 			(min(base, blend) - max(base, blend) + vec3(1.0))
+#define BlendOpacity(base, blend, F, O) 	(F(base, blend) * O + blend * (1.0 - O))
+
+
+vec3 BlendHue(vec3 base, vec3 blend)
+{
+    vec3 baseHSL = RGBToHSL(base);
+    return HSLToRGB(vec3(RGBToHSL(blend).r, baseHSL.g, baseHSL.b));
+}
+
+vec3 BlendSaturation(vec3 base, vec3 blend)
+{
+    vec3 baseHSL = RGBToHSL(base);
+    return HSLToRGB(vec3(baseHSL.r, RGBToHSL(blend).g, baseHSL.b));
+}
+
+vec3 BlendColor(vec3 base, vec3 blend)
+{
+    vec3 blendHSL = RGBToHSL(blend);
+    return HSLToRGB(vec3(blendHSL.r, blendHSL.g, RGBToHSL(base).b));
+}
+
+vec3 BlendLuminosity(vec3 base, vec3 blend)
+{
+    vec3 baseHSL = RGBToHSL(base);
+    return HSLToRGB(vec3(baseHSL.r, baseHSL.g, RGBToHSL(blend).b));
+}
+
+// Chroma keying
+// Author: Tim Eves <timseves@googlemail.com>
+//
+// This implements the Chroma key algorithm described in the paper:
+//      'Software Chroma Keying in an Imersive Virtual Environment'
+//      by F. van den Bergh & V. Lalioti
+// but as a pixel shader algorithm.
+//
+vec4  grey_xfer  = vec4(luma_coeff, 0);
+
+// This allows us to implement the paper's alphaMap curve in software
+// rather than a largeish array
+float alpha_map(float d)
+{
+    return 1.0 - smoothstep(1.0, chroma_softness, d);
+}
+
+// http://stackoverflow.com/questions/15095909/from-rgb-to-hsv-in-opengl-glsl
+vec3 rgb2hsv(vec3 c)
+{
+    vec4 K = vec4(0.0, -1.0 / 3.0, 2.0 / 3.0, -1.0);
+    vec4 p = mix(vec4(c.bg, K.wz), vec4(c.gb, K.xy), step(c.b, c.g));
+    vec4 q = mix(vec4(p.xyw, c.r), vec4(c.r, p.yzx), step(p.x, c.r));
+
+    float d = q.x - min(q.w, q.y);
+    float e = 1.0e-10;
+    return vec3(abs(q.z + (q.w - q.y) / (6.0 * d + e)), d / (q.x + e), q.x);
+}
+
+// From the same page
+vec3 hsv2rgb(vec3 c)
+{
+    vec4 K = vec4(1.0, 2.0 / 3.0, 1.0 / 3.0, 3.0);
+    vec3 p = abs(fract(c.xxx + K.xyz) * 6.0 - K.www);
+    return c.z * mix(K.xxx, clamp(p - K.xxx, 0.0, 1.0), c.y);
+}
+
+float AngleDiff(float angle1, float angle2)
+{
+    return 0.5 - abs(abs(angle1 - angle2) - 0.5);
+}
+
+float AngleDiffDirectional(float angle1, float angle2)
+{
+    float diff = angle1 - angle2;
+
+    return diff < -0.5
+            ? diff + 1.0
+            : (diff > 0.5 ? diff - 1.0 : diff);
+}
+
+float Distance(float actual, float target)
+{
+    return min(0.0, target - actual);
+}
+
+float ColorDistance(vec3 hsv)
+{
+    float hueDiff					= AngleDiff(hsv.x, chroma_target_hue) * 2;
+    float saturationDiff			= Distance(hsv.y, chroma_min_saturation);
+    float brightnessDiff			= Distance(hsv.z, chroma_min_brightness);
+
+    float saturationBrightnessScore	= max(brightnessDiff, saturationDiff);
+    float hueScore					= hueDiff - chroma_hue_width;
+
+    return -hueScore * saturationBrightnessScore;
+}
+
+vec3 supress_spill(vec3 c)
+{
+    float hue		= c.x;
+    float diff		= AngleDiffDirectional(hue, chroma_target_hue);
+    float distance	= abs(diff) / chroma_spill_suppress;
+
+    if (distance < 1)
+    {
+        c.x = diff < 0
+                ? chroma_target_hue - chroma_spill_suppress
+                : chroma_target_hue + chroma_spill_suppress;
+        c.y *= min(1.0, distance + chroma_spill_suppress_saturation);
+    }
+
+    return c;
+}
+
+// Key on any color
+vec4 ChromaOnCustomColor(vec4 c)
+{
+    vec3 hsv		= rgb2hsv(c.rgb);
+    float distance	= ColorDistance(hsv);
+    float d			= distance * -2.0 + 1.0;
+    vec4 suppressed	= vec4(hsv2rgb(supress_spill(hsv)), 1.0);
+    float alpha		= alpha_map(d);
+
+    suppressed *= alpha;
+
+    return chroma_show_mask ? vec4(suppressed.a, suppressed.a, suppressed.a, 1) : suppressed;
+}
+
+
+
+vec3 get_blend_color(vec3 back, vec3 fore)
+{
+    switch(blend_mode)
+    {
+    case  0: return BlendNormal(back, fore);
+    case  1: return BlendLighten(back, fore);
+    case  2: return BlendDarken(back, fore);
+    case  3: return BlendMultiply(back, fore);
+    case  4: return BlendAverage(back, fore);
+    case  5: return BlendAdd(back, fore);
+    case  6: return BlendSubstract(back, fore);
+    case  7: return BlendDifference(back, fore);
+    case  8: return BlendNegation(back, fore);
+    case  9: return BlendExclusion(back, fore);
+    case 10: return BlendScreen(back, fore);
+    case 11: return BlendOverlay(back, fore);
+//	case 12: return BlendSoftLight(back, fore);
+    case 13: return BlendHardLight(back, fore);
+    case 14: return BlendColorDodge(back, fore);
+    case 15: return BlendColorBurn(back, fore);
+    case 16: return BlendLinearDodge(back, fore);
+    case 17: return BlendLinearBurn(back, fore);
+    case 18: return BlendLinearLight(back, fore);
+    case 19: return BlendVividLight(back, fore);
+    case 20: return BlendPinLight(back, fore);
+    case 21: return BlendHardMix(back, fore);
+    case 22: return BlendReflect(back, fore);
+    case 23: return BlendGlow(back, fore);
+    case 24: return BlendPhoenix(back, fore);
+    case 25: return BlendHue(back, fore);
+    case 26: return BlendSaturation(back, fore);
+    case 27: return BlendColor(back, fore);
+    case 28: return BlendLuminosity(back, fore);
+    }
+    return BlendNormal(back, fore);
+}
+
+vec4 blend(vec4 fore)
+{
+    // attachments have bgra format to match opengl accelerator
+    vec4 back = subpassLoad(background).bgra;
+    if(blend_mode != 0)
+       fore.rgb = get_blend_color(back.rgb/(back.a+0.0000001), fore.rgb/(fore.a+0.0000001))*fore.a;
+   switch(keyer)
+   {
+       case 1:  return fore + back; // additive
+       default: return fore + (1.0-fore.a)*back; // linear
+   }
+}
+
+vec4 chroma_key(vec4 c)
+{
+    return ChromaOnCustomColor(c);
+}
+
+vec4 ycbcra_to_rgba(float Y, float Cb, float Cr, float A)
+{
+    const float luma_coefficient = 255.0/219.0;
+    const float chroma_coefficient = 255.0/224.0;
+
+    vec3 YCbCr = vec3(Y, Cb, Cr) * 255;
+    YCbCr -= vec3(16.0, 128.0, 128.0);
+    YCbCr *= vec3(luma_coefficient, chroma_coefficient, chroma_coefficient);
+
+    return vec4(color_matrix * YCbCr / 255, A);
+}
+
+vec4 get_sample(sampler2D texSampler, vec2 coords)
+{
+    return texture(texSampler, coords);
+}
+
+vec4 get_rgba_color()
+{
+    switch(pixel_format)
+    {
+    case 0:		//gray
+        return vec4(get_sample(plane[0], TexCoord.st / TexCoord.q).rrr * precision_factor[0], 1.0);
+    case 1:		//bgra,
+        return get_sample(plane[0], TexCoord.st / TexCoord.q).bgra * precision_factor[0];
+    case 2:		//rgba,
+        return get_sample(plane[0], TexCoord.st / TexCoord.q).rgba * precision_factor[0];
+    case 3:		//argb,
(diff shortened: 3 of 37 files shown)


niklaspandersson:
I've fixed:

- dynamic loading of the Vulkan libraries (so that it doesn't  depend on paths in the build environment)
- changed the required runtime version to 1.3 (but kept the required extension DYNAMIC_RENDERING_LOCAL_READ. According gpuinfo.org, this is supported on nvidia Turing-architecture and forward, 2018 and later and Intel 11th Gen+ (Tiger Lake+), which is even older)
- An rgba texture format bug in the vulkan fragment shader

niklaspandersson:
> > Related to the GPU format conversion PR. I think the Vulkan accelerator will simplify that effort. It especially makes the CPU/GPU synchronization at lot easier while maintaining parallellism, without having to explicitly wait.
> 
> Yeah,I've not had much motivation to continue on that recently, and that was made easy once you started on this 😆 . tbh I was expecting the sync to not be a problem, but I didn't test under much load so maybe it wasnt. But once this is merged maybe I can see if I can get that rebased onto this and see where it stands. The challenge now will be how to do it without needing to do the same work for both opengl and vulkan, as that will be tedious to do everything twice and just result in bugs. But maybe it is possible to keep things working similarly to today for opengl and keep the new stuff for vulkan

If you lack the motivation and/or time, I am very interested in this and will have both time and some budget during the months before midsummer.

niklaspandersson:
By the way. I didn't want to base this on the tip of master since there is build issues on windows there. Once the master compiles fine on all platforms, I'll rebase again.

niklaspandersson:
...and macOS support will be provided in a completely separate PR

dimitry-ishenko:
@niklaspandersson FWIW I've  included 2 patches in my OAL PR #1714 to fix windows builds. Hopefully, @Julusian accepts it soon...

src/CMakeModules/Bootstrap_Windows.cmake
```diff
@@ -151,6 +152,22 @@ target_link_directories(GLEW::glew INTERFACE ${glew_SOURCE_DIR}/lib/Release/x64)
 target_link_libraries(GLEW::glew INTERFACE glew32)
 casparcg_add_runtime_dependency("${glew_SOURCE_DIR}/bin/Release/x64/glew32.dll")
 
+IF(ENABLE_VULKAN)
+	find_package(Vulkan REQUIRED)
+	
+	FetchContent_Declare(fetch_vk_bootstrap
```
Julusian: I would prefer for these to come from tar.gz/zips added to https://github.com/CasparCG/dependencies/releases

Partly so that we don't have to worry about old repositories being deleted or servers shutting down, and a nice side effect is the ability to point it to a custom local cache which could be useful in some networks

src/CMakeModules/Bootstrap_Linux.cmake
```diff
@@ -40,6 +41,26 @@ find_package(SFML 3 COMPONENTS Graphics System Window QUIET)
 if(NOT SFML_FOUND)
     find_package(SFML 2 COMPONENTS graphics system window REQUIRED)
 endif()
+
+IF (ENABLE_VULKAN)
+    find_package(Vulkan REQUIRED)
+
+    FetchContent_Declare(
```
Julusian: Perhaps similar comment to windows, but for linux it is really nice to not need to download anything. some of the ubuntu packages I have rely on this, with their build environment not allowed internet access.

seems like debian has one of these already, but I don't see the other
https://packages.debian.org/sid/libvulkan-memory-allocator-dev

but perhaps something for a follow up?

tools/linux/Dockerfile
```diff
@@ -5,15 +5,21 @@ FROM docker.io/buildpack-deps:noble AS build-casparcg
...
+	ENV VULKAN_SDK=/tmp/vulkan-sdk/x86_64
+	ENV PATH=$VULKAN_SDK/bin:$PATH
+
+	COPY ./src /source
+
 	ARG CC
 	ARG CXX
 	ARG GIT_HASH
 
-	RUN cmake -GNinja /source -DUSE_STATIC_BOOST=ON -DUSE_SYSTEM_CEF=OFF
```
Julusian, line 16: Doesnt have to be part of this PR, but it would be nice if the `ubuntu:26.04` builds in the linux-system.yml workflow would enable this too. It looks like that release has vulkan 1.4 packages available. 

Perhaps for debian:sid and arch too, but thats definitely something for me to look at.

Julusian (approved):
I don't think it is worth me reading through this in much detail, I doubt I will spot anything wrong with my limited knowledge of vulkan (I was trying it for a project 5 years ago, and havent touched it since)
But maybe worth getting a review from some AI? I can trigger a review from copilot for you if you think that would be helpful

I was able to build this on ubuntu 24.04 by manually installing the vulkan debs from 26.04, as well as installing the `glslc` and `vulkan-validationlayers` packages.
It appeared to run ok in a very basic test.

It is spewing unending validation errors though. not sure if that is something to look at or noise from my mismatched vulkan install.
```
[2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b655560, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
[2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b655560, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
[2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
[2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x520000000052, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x500000000050, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x500000000050[] in imageView VkImageView 0x520000000052[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
[2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-VkImageMemoryBarrier2-oldLayout-parameter, Validation Error: [ VUID-VkImageMemoryBarrier2-oldLayout-parameter ] | MessageID = 0xf3ed55bf | vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].oldLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: oldLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-oldLayout-parameter)
[2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-VkImageMemoryBarrier2-newLayout-parameter, Validation Error: [ VUID-VkImageMemoryBarrier2-newLayout-parameter ] | MessageID = 0x18a78435 | vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].newLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: newLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-newLayout-parameter)
[2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-VkRenderingAttachmentInfo-imageLayout-parameter, Validation Error: [ VUID-VkRenderingAttachmentInfo-imageLayout-parameter ] | MessageID = 0x4d3f9c2c | vkCmdBeginRendering(): pRenderingInfo->pColorAttachments[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: imageLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkRenderingAttachmentInfo-imageLayout-parameter)
[2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
[2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
[2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
[2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
[2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkImageMemoryBarrier2-oldLayout-parameter, Validation Error: [ VUID-VkImageMemoryBarrier2-oldLayout-parameter ] | MessageID = 0xf3ed55bf | vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].oldLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: oldLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-oldLayout-parameter)
[2026-04-15 19:43:18.755] [info] async_event_server[:5250] Client 127.0.0.1 disconnected (0 connections).
```

src/CMakeModules/Bootstrap_Windows.cmake
niklaspandersson: For sure! But I'm not entirely sure how the releases in that repo works. I can see various tags, but I can't see anything that builds / packages anything

niklaspandersson:
> I don't think it is worth me reading through this in much detail, I doubt I will spot anything wrong with my limited knowledge of vulkan (I was trying it for a project 5 years ago, and havent touched it since) But maybe worth getting a review from some AI? I can trigger a review from copilot for you if you think that would be helpful

Yea, for sure. Let me rebase on master and see if I can fix some of these validation errors first. 

> I was able to build this on ubuntu 24.04 by manually installing the vulkan debs from 26.04, as well as installing the `glslc` and `vulkan-validationlayers` packages. It appeared to run ok in a very basic test.

Oh, without downloading the entire SDK from lunarg? That's great
 
> It is spewing unending validation errors though. not sure if that is something to look at or noise from my mismatched vulkan install.
>
> ```
> [2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b655560, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
> [2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b655560, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
> [2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
> [2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x520000000052, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x500000000050, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x500000000050[] in imageView VkImageView 0x520000000052[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
> [2026-04-15 19:43:18.727] [info] [ERROR: Validation] - VUID-VkImageMemoryBarrier2-oldLayout-parameter, Validation Error: [ VUID-VkImageMemoryBarrier2-oldLayout-parameter ] | MessageID = 0xf3ed55bf | vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].oldLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: oldLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-oldLayout-parameter)
> [2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-VkImageMemoryBarrier2-newLayout-parameter, Validation Error: [ VUID-VkImageMemoryBarrier2-newLayout-parameter ] | MessageID = 0x18a78435 | vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].newLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: newLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-newLayout-parameter)
> [2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-VkRenderingAttachmentInfo-imageLayout-parameter, Validation Error: [ VUID-VkRenderingAttachmentInfo-imageLayout-parameter ] | MessageID = 0x4d3f9c2c | vkCmdBeginRendering(): pRenderingInfo->pColorAttachments[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: imageLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkRenderingAttachmentInfo-imageLayout-parameter)
> [2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
> [2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
> [2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
> [2026-04-15 19:43:18.730] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718, Validation Error: [ VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x515e1bea | vkCmdPipelineBarrier2(): pDependencyInfo can not be called inside a dynamic rendering instance. This can be fixed by enabling the VK_EXT_shader_tile_image features. The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance and none of the shaderTileImageColorReadAccess, shaderTileImageDepthReadAccess, shaderTileImageStencilReadAccess features are enabled, the render pass must not have been started with vkCmdBeginRendering (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-shaderTileImageColorReadAccess-08718)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-vkCmdPipelineBarrier2-None-08719, Validation Error: [ VUID-vkCmdPipelineBarrier2-None-08719 ] Object 0: handle = 0x55c05b5ab1d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xe12198b2 | vkCmdPipelineBarrier2(): pDependencyInfo->pMemoryBarriers[0].dstAccessMask (VK_ACCESS_INPUT_ATTACHMENT_READ_BIT) is not from allowed access mask (VK_ACCESS_COLOR_ATTACHMENT_READ_BIT|VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT|VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT). The Vulkan spec states: If vkCmdPipelineBarrier2 is called within a render pass instance started with vkCmdBeginRendering, it must adhere to the restrictions in Explicit Render Pass Tile Image Access Synchronization (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdPipelineBarrier2-None-08719)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID_Undefined, Validation Error: [ VUID_Undefined ] | MessageID = 0x79de34d4 | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0].imageLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token.
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkWriteDescriptorSet-descriptorType-04151, Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-04151 ] Object 0: handle = 0x620000000062, type = VK_OBJECT_TYPE_IMAGE_VIEW; Object 1: handle = 0x600000000060, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x62405e2d | vkUpdateDescriptorSets(): pDescriptorWrites[1].pImageInfo[0] Descriptor update with descriptorType VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is being updated with invalid imageLayout Unhandled VkImageLayout for image VkImage 0x600000000060[] in imageView VkImageView 0x620000000062[]. Allowed layouts are: VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_ATTACHMENT_OPTIMAL, VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL, VK_IMAGE_LAYOUT_STENCIL_READ_ONLY_OPTIMAL. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT the imageLayout member of each element of pImageInfo must be a member of the list given in Input Attachment (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-04151)
> [2026-04-15 19:43:18.731] [info] [ERROR: Validation] - VUID-VkImageMemoryBarrier2-oldLayout-parameter, Validation Error: [ VUID-VkImageMemoryBarrier2-oldLayout-parameter ] | MessageID = 0xf3ed55bf | vkCmdPipelineBarrier2(): pDependencyInfo->pImageMemoryBarriers[0].oldLayout (1000232000) does not fall within the begin..end range of the core VkImageLayout enumeration tokens and is not an extension added token. The Vulkan spec states: oldLayout must be a valid VkImageLayout value (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-oldLayout-parameter)
> [2026-04-15 19:43:18.755] [info] async_event_server[:5250] Client 127.0.0.1 disconnected (0 connections).
> ```

Hmm, that is odd. I'm not seeing those. Can you see if you can find the validation layer settings that's being used for you. Also, what kind of gpu do you have, and which driver?

src/CMakeModules/Bootstrap_Windows.cmake
Julusian: Its manually uploaded archives taken from elsewhere, except for a few which are compiled in some way
so for those git repositories, probably appropriate to use the github option to download the source as a tar.gz and upload them to new tags there.

If you're unsure about how to do this, I can do this after its merged

Julusian:
> Hmm, that is odd. I'm not seeing those. Can you see if you can find the validation layer settings that's being used for you. Also, what kind of gpu do you have, and which driver?

ok, yeah lets ignore this for now then. It could well be some mismatched versions of libraries (I also observed today that webgl is broken in desktop chrome for me, I didnt notice that before today so that could be related)
It probably only shows in debug mode? not sure how I launched this yesterday

Ubuntu 24.04, nvidia 580.126.09 (from ubuntu apt), GTX 1060 6GB

niklaspandersson:
> > Hmm, that is odd. I'm not seeing those. Can you see if you can find the validation layer settings that's being used for you. Also, what kind of gpu do you have, and which driver?
> 
> ok, yeah lets ignore this for now then. It could well be some mismatched versions of libraries (I also observed today that webgl is broken in desktop chrome for me, I didnt notice that before today so that could be related) It probably only shows in debug mode? not sure how I launched this yesterday

Yes, it only uses validation layers in the debug build.

> Ubuntu 24.04, nvidia 580.126.09 (from ubuntu apt), GTX 1060 6GB

src/CMakeModules/Bootstrap_Linux.cmake
niklaspandersson: Fixed in the rebased history

src/CMakeModules/Bootstrap_Windows.cmake
niklaspandersson: fixed in the rebased history

niklaspandersson:
@Julusian, ok, let's see what copilot has to say about this

Julusian:
some other comments I got the other day when I asked claude to compare the review it compared to the opengl implementation. Ive only skim read what it said, not sure if it is all reasonable but it seems more thorough than the default copilot

---

Here's a thorough review of PR #1677 (Vulkan accelerator), compared against the OpenGL implementation, focused on correctness and behavioural parity.

---

## Overall Assessment

The implementation is a solid first pass with good structural fidelity to the OpenGL version. The core rendering pipeline, transform/crop system, and shader logic are closely matched. There are however several real behavioural issues and a few design concerns worth flagging.

---

## 🔴 Bugs / Behavioural Issues

### 1. `local_key` texture uses wrong component count (1 vs 4)
**File:** `vulkan/image/image_mixer.cpp` — `draw(item, ...)` method

```cpp
// Vulkan:
local_key_texture = local_key_texture ? local_key_texture : pass->create_attachment();
// create_attachment() defaults to components_count=4
```

```cpp
// OpenGL (image_mixer.cpp:196-198):
local_key_texture = local_key_texture ? local_key_texture
                  : ogl_->create_texture(target_texture->width(), target_texture->height(), 1, depth_);
//                                                                                          ^
//                                                                            Single channel for a key!
```

The OpenGL version creates a **1-channel** texture for the local key (only alpha/luma matters). The Vulkan version creates a full **4-channel RGBA** attachment. This wastes 4x memory for keys but more critically — the pipeline writes `fragColor.bgra` to the attachment, so when the key texture is read back, the shader reads it as `texture(local_key, ...).r`, which will give the **blue channel** (since the attachment is BGRA-ordered). This will cause silently broken keying. The key texture should be a single-channel texture, or at minimum the write path needs to output to `.r` consistently.

---

### 2. `renderpass::draw()` — texture array initialised with `params.textures[0]` for missing slots, but there may be zero textures
**File:** `vulkan/util/renderpass.cpp`

```cpp
std::array<vk::ImageView, 7> textures = {params.textures[0]->view(),  // repeated 4x
                                          params.textures[0]->view(),
                                          ...
                                          attachment->view(),
                                          params.textures[0]->view(),  // local_key default
                                          params.textures[0]->view()}; // layer_key default
```

If `params.textures` is empty (e.g. a key draw with no texture planes — theoretically guarded upstream but fragile), this crashes. More practically: when `params.local_key` or `params.layer_key` are null, the slots fall back to `params.textures[0]`, but the shader will still sample those slots if the `has_local_key`/`has_layer_key` flags are not set. This is fine behaviour-wise, but it means a valid image view must always exist for those slots. There's no `VK_NULL_HANDLE`-safe dummy texture, which can trigger Vulkan validation errors with descriptorBindingPartiallyBound **only** if the binding is actually accessed. Since the flags gate access in the shader this is borderline — but it makes Vulkan validation noisy and is fragile if the shader is ever changed.

---

### 3. Frame buffer ring size of 3 and no synchronisation between `image_kernel` vertex buffer and the GPU
**File:** `vulkan/image/image_kernel.cpp` — `upload_vertex_buffer`

The vertex buffer is CPU-mapped memory that is written *then* immediately submitted. With `frame_buffer_size = 3`, there's a fence wait at `create_renderpass` time before reusing a slot. However, the vertex buffer is allocated **per frame slot** but re-used across draw calls within a single frame (since `upload_vertex_buffer` is called once per `renderpass::commit()`). This is fine. But if `create_renderpass` is called again before the previous frame's fence fires and `_device.waitForFences` returns `eTimeout`, a `std::runtime_error` is thrown — which will propagate through the deferred future chain and crash the channel. The OGL version has no equivalent timeout concern since it's serial. **Consider whether a timeout abort is the right recovery path, or whether the channel should be restarted gracefully.**

---

### 4. `copy_async` (readback) always waits synchronously on the device thread
**File:** `vulkan/util/device.cpp` — `copy_async(texture)`

```cpp
submitSingleTimeCommands([&](vk::CommandBuffer cmd) {
    // transitions + copyImageToBuffer2
}, true); // wait=true
// TODO: return the fence and wait on it in the callers thread instead
```

The OpenGL version uses PBOs and an async readback with the wait deferred to the caller's thread. The Vulkan version **always stalls the device thread** until the copy is complete. This means every frame stalls the entire Vulkan dispatch thread waiting for a GPU→CPU readback. The TODO comment acknowledges this. At broadcast frame rates this will likely be a **significant latency/throughput bottleneck** depending on frame size. Not a crash, but a critical perf issue for production use.

---

### 5. `copy_async` (upload) also always waits synchronously
**File:** `vulkan/util/device.cpp` — `copy_async(array, ...)`

```cpp
submitSingleTimeCommands([&](...) {
    // transition + copyBufferToImage + transition
}, true); // wait=true
```

Same issue. OGL uploads textures and gets back a future that can be awaited separately. Here, **every texture upload blocks the dispatch thread**. For multi-plane formats like YCbCr 4:2:0 this means 3 sequential stalling uploads per frame per channel.

---

### 6. `renderpass` `commit()` — missing `endRendering()` when `layers_` is non-empty
**File:** `vulkan/util/renderpass.cpp`

The loop inside `commit()` calls `beginRendering()` per new attachment and ends up calling `endRendering()` at the very bottom unconditionally. The last rendering pass gets `endRendering()` called at the end of the function, which is correct. But trace through what happens when `layers_` is **empty**: there's a path where `beginRendering` is called to clear, then `cmd_buffer.setViewport/setScissor`, but the loop body never runs, so `endRendering` is called at line ~220 outside the if/else. This is fine. **However**, when `layers_` is non-empty, the final `endRendering()` call at the end of the function body is correct. Confirm this is always reached — it is (after the else block falls through) — so this is actually OK, just hard to follow. Consider restructuring for clarity.

---

## 🟡 Potentially Incorrect Behaviour

### 7. `blend_mode` comparison in shader uses `>= 0` which always true for `int`
**File:** `vulkan/image/fragment_shader.frag`

```glsl
if (blend_mode >= 0)
    color = blend(color);
```

`blend_mode` is declared as `int` in the push constant, and `blend_mode::normal = 0`, so this runs the blend function **on every draw**, including when blend_mode is 0 (normal). The `blend()` function handles this correctly (normal blend is a passthrough when background is considered), but this means `subpassLoad(background)` is **always executed**, requiring the background attachment to always be valid. Compare to how the OGL shader handles this — the OGL shader also always blends, so this is actually consistent. Fine, but the comment about "blend_mode >= 0" is confusing since it's always true — it should just be `blend(color)` unconditionally, or use a `blend_mode != -1` sentinel if there's intent to skip blending.


---

### 9. `DescriptorPoolSize = 10` is a fixed pool — no growth, silent descriptor starvation
**File:** `vulkan/util/pipeline.cpp`

The descriptor pool is fixed at 10 sets, recycled round-robin. With the 3-frame kernel ring and multiple channels, under heavy load you could be overwriting a descriptor set that's still in-flight. The fence waits in `image_kernel` are per-slot but the pipeline's descriptor recycling is global and unsynchronised against those fences. If more than 10 concurrent draw calls are in flight across all channels, descriptors get silently stomped. This won't crash immediately (Vulkan may not validate it) but will produce corrupted frames under load.

---

### 10. `device.h` declares `transitionImageLayout` with a `vk::Format` parameter that `device.cpp` does not use
**File:** `vulkan/util/device.h` vs `device.cpp`

```cpp
// device.h declaration:
void transitionImageLayout(const vk::Image&, vk::Format format, ...);

// device.cpp definition:
void transitionImageLayout(const vk::Image&, vk::ImageLayout oldLayout, ...);
//                                           ^ no Format parameter
```

The header declares a signature with `vk::Format format` as the second argument; the `.cpp` definition omits it. This means the function defined in `device.cpp` doesn't match the declaration in `device.h` — this is a **compile error** waiting to happen for any TU that includes `device.h` and tries to call `transitionImageLayout`. The internal calls in `device.cpp` use the local free function (not the declared one), so it doesn't fail *yet*, but this is a latent bug.

---

### 11. `texture::impl` uses `GLsizei` for `width_`, `height_`, `stride_`, `size_`
**File:** `vulkan/util/texture.cpp`

```cpp
GLsizei width_  = 0;
GLsizei height_ = 0;
```

`GLsizei` is an OpenGL type. Using it in the Vulkan implementation is harmless (it's typedef'd to `int`) but is a copy-paste artefact that should be cleaned up — especially since `GLsizei` requires `<GL/glew.h>` or similar include.

niklaspandersson:
Great! I'll have a look at these

src/accelerator/vulkan/image/fragment_shader.frag
```diff
@@ -0,0 +1,578 @@
...
+};
+
+bool is_straight_alpha = (flags & is_straight_alpha_mask) == is_straight_alpha_mask;
+bool has_local_key = (flags & has_local_key_mask) == has_local_key_mask;
+bool has_layer_key  = (flags & has_layer_key_mask) == has_layer_key_mask;
+bool invert = (flags & invert_mask) == invert_mask;
+bool levels = (flags & levels_mask) == levels_mask;
+bool csb = (flags & csb_mask) == csb_mask;
+bool chroma = (flags & chroma_mask) == chroma_mask;
+bool chroma_show_mask = (flags & chroma_show_mask_mask) == chroma_show_mask_mask;
```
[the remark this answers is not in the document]
niklaspandersson: Not sure if this applies for Vulkan. But moving it to main anyway, they're only used there anyway

src/accelerator/vulkan/image/fragment_shader.frag
```diff
@@ -0,0 +1,578 @@
...
+                ); 
+const vec3[3] luma_coefficients = vec3[3](
+                    vec3(0.299, 0.587, 0.114), // Rec. 601
+                    vec3(0.2126, 0.7152, 0.0722),  // Rec. 709
+                    vec3(0.2627, 0.6780, 0.0593)   // Rec. 2020
+                );
+
+mat3 color_matrix = transpose(color_matrices[color_space_index]);
+vec3 luma_coeff = luma_coefficients[color_space_index];
+
```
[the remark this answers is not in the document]
niklaspandersson: Again, not sure if this applies to Vulkan. But moved the color_matrix in to the function where it is used (only a single place), same for luma_coeff

niklaspandersson:
> some other comments I got the other day when I asked claude to compare the review it compared to the opengl implementation. Ive only skim read what it said, not sure if it is all reasonable but it seems more thorough than the default copilot
> 
> Here's a thorough review of PR #1677 (Vulkan accelerator), compared against the OpenGL implementation, focused on correctness and behavioural parity.
> 
> ## Overall Assessment
> The implementation is a solid first pass with good structural fidelity to the OpenGL version. The core rendering pipeline, transform/crop system, and shader logic are closely matched. There are however several real behavioural issues and a few design concerns worth flagging.
> 
> ## 🔴 Bugs / Behavioural Issues
> ### 1. `local_key` texture uses wrong component count (1 vs 4)
> **File:** `vulkan/image/image_mixer.cpp` — `draw(item, ...)` method
> 
> ```c++
> // Vulkan:
> local_key_texture = local_key_texture ? local_key_texture : pass->create_attachment();
> // create_attachment() defaults to components_count=4
> ```
> 
> ```c++
> // OpenGL (image_mixer.cpp:196-198):
> local_key_texture = local_key_texture ? local_key_texture
>                   : ogl_->create_texture(target_texture->width(), target_texture->height(), 1, depth_);
> //                                                                                          ^
> //                                                                            Single channel for a key!
> ```
> 
> The OpenGL version creates a **1-channel** texture for the local key (only alpha/luma matters). The Vulkan version creates a full **4-channel RGBA** attachment. This wastes 4x memory for keys but more critically — the pipeline writes `fragColor.bgra` to the attachment, so when the key texture is read back, the shader reads it as `texture(local_key, ...).r`, which will give the **blue channel** (since the attachment is BGRA-ordered). This will cause silently broken keying. The key texture should be a single-channel texture, or at minimum the write path needs to output to `.r` consistently.

This is a real issue, but the opengl implementation has the same problem. Yes, it creates a single channel target, but it still writes bgra, in effect writing b to the target and discarding gra.
So, same as current implementation ✅, but needs to be adressed in both opengl and vulkan. Already on my list

> ### 2. `renderpass::draw()` — texture array initialised with `params.textures[0]` for missing slots, but there may be zero textures
> **File:** `vulkan/util/renderpass.cpp`
> 
> ```c++
> std::array<vk::ImageView, 7> textures = {params.textures[0]->view(),  // repeated 4x
>                                           params.textures[0]->view(),
>                                           ...
>                                           attachment->view(),
>                                           params.textures[0]->view(),  // local_key default
>                                           params.textures[0]->view()}; // layer_key default
> ```
> 
> If `params.textures` is empty (e.g. a key draw with no texture planes — theoretically guarded upstream but fragile), this crashes. More practically: when `params.local_key` or `params.layer_key` are null, the slots fall back to `params.textures[0]`, but the shader will still sample those slots if the `has_local_key`/`has_layer_key` flags are not set. This is fine behaviour-wise, but it means a valid image view must always exist for those slots. There's no `VK_NULL_HANDLE`-safe dummy texture, which can trigger Vulkan validation errors with descriptorBindingPartiallyBound **only** if the binding is actually accessed. Since the flags gate access in the shader this is borderline — but it makes Vulkan validation noisy and is fragile if the shader is ever changed.

~~I'm aware that is is a bit hacky, but it was the quickest way to get this working **without** validation errors or warnings. I intend to look into bindless or partially binding this to make it cleaner. This is already on my list~~ Fixed in new, separate commit ✅

> ### 3. Frame buffer ring size of 3 and no synchronisation between `image_kernel` vertex buffer and the GPU
> **File:** `vulkan/image/image_kernel.cpp` — `upload_vertex_buffer`
> 
> The vertex buffer is CPU-mapped memory that is written _then_ immediately submitted. With `frame_buffer_size = 3`, there's a fence wait at `create_renderpass` time before reusing a slot. However, the vertex buffer is allocated **per frame slot** but re-used across draw calls within a single frame (since `upload_vertex_buffer` is called once per `renderpass::commit()`). This is fine. But if `create_renderpass` is called again before the previous frame's fence fires and `_device.waitForFences` returns `eTimeout`, a `std::runtime_error` is thrown — which will propagate through the deferred future chain and crash the channel. The OGL version has no equivalent timeout concern since it's serial. **Consider whether a timeout abort is the right recovery path, or whether the channel should be restarted gracefully.**

This might be an oversight on my part, I will have to look into this. I'll put it on my list

> ### 4. `copy_async` (readback) always waits synchronously on the device thread
> **File:** `vulkan/util/device.cpp` — `copy_async(texture)`
> 
> ```c++
> submitSingleTimeCommands([&](vk::CommandBuffer cmd) {
>     // transitions + copyImageToBuffer2
> }, true); // wait=true
> // TODO: return the fence and wait on it in the callers thread instead
> ```
> 
> The OpenGL version uses PBOs and an async readback with the wait deferred to the caller's thread. The Vulkan version **always stalls the device thread** until the copy is complete. This means every frame stalls the entire Vulkan dispatch thread waiting for a GPU→CPU readback. The TODO comment acknowledges this. At broadcast frame rates this will likely be a **significant latency/throughput bottleneck** depending on frame size. Not a crash, but a critical perf issue for production use.
> 
> ### 5. `copy_async` (upload) also always waits synchronously
> **File:** `vulkan/util/device.cpp` — `copy_async(array, ...)`
> 
> ```c++
> submitSingleTimeCommands([&](...) {
>     // transition + copyBufferToImage + transition
> }, true); // wait=true
> ```
> 
> Same issue. OGL uploads textures and gets back a future that can be awaited separately. Here, **every texture upload blocks the dispatch thread**. For multi-plane formats like YCbCr 4:2:0 this means 3 sequential stalling uploads per frame per channel.

~~As the comments suggest, these are issues that I'm aware of. I think we should merge this anyway, so that I can adress these kinds of issues without worrying about drifting out of sync with master. As long as the vulkan accelerator is opt-in, we have plenty of time to fix this~~ Fixed in last commit ✅
 
> ### 6. `renderpass` `commit()` — missing `endRendering()` when `layers_` is non-empty
> **File:** `vulkan/util/renderpass.cpp`
> 
> The loop inside `commit()` calls `beginRendering()` per new attachment and ends up calling `endRendering()` at the very bottom unconditionally. The last rendering pass gets `endRendering()` called at the end of the function, which is correct. But trace through what happens when `layers_` is **empty**: there's a path where `beginRendering` is called to clear, then `cmd_buffer.setViewport/setScissor`, but the loop body never runs, so `endRendering` is called at line ~220 outside the if/else. This is fine. **However**, when `layers_` is non-empty, the final `endRendering()` call at the end of the function body is correct. Confirm this is always reached — it is (after the else block falls through) — so this is actually OK, just hard to follow. Consider restructuring for clarity.

Works as intened ✅ I'll see if this can be refactored or at least drop a comment at a later stage. I'll add it to my list of todos.

> ## 🟡 Potentially Incorrect Behaviour
> ### 7. `blend_mode` comparison in shader uses `>= 0` which always true for `int`
> **File:** `vulkan/image/fragment_shader.frag`
> 
> ```glsl
> if (blend_mode >= 0)
>     color = blend(color);
> ```
> 
> `blend_mode` is declared as `int` in the push constant, and `blend_mode::normal = 0`, so this runs the blend function **on every draw**, including when blend_mode is 0 (normal). The `blend()` function handles this correctly (normal blend is a passthrough when background is considered), but this means `subpassLoad(background)` is **always executed**, requiring the background attachment to always be valid. Compare to how the OGL shader handles this — the OGL shader also always blends, so this is actually consistent. Fine, but the comment about "blend_mode >= 0" is confusing since it's always true — it should just be `blend(color)` unconditionally, or use a `blend_mode != -1` sentinel if there's intent to skip blending.

Mimics the opengl path, which is fint at this stage ✅  But I noticed this too while looking over it. I have this on my list of todos related to this, to at least make this read more natural.

> ### 9. `DescriptorPoolSize = 10` is a fixed pool — no growth, silent descriptor starvation
> **File:** `vulkan/util/pipeline.cpp`
> 
> The descriptor pool is fixed at 10 sets, recycled round-robin. With the 3-frame kernel ring and multiple channels, under heavy load you could be overwriting a descriptor set that's still in-flight. The fence waits in `image_kernel` are per-slot but the pipeline's descriptor recycling is global and unsynchronised against those fences. If more than 10 concurrent draw calls are in flight across all channels, descriptors get silently stomped. This won't crash immediately (Vulkan may not validate it) but will produce corrupted frames under load.

~~This I need to investigate. Adding it to my list~~ Bumped up to 64 ✅

> ### 10. `device.h` declares `transitionImageLayout` with a `vk::Format` parameter that `device.cpp` does not use
> **File:** `vulkan/util/device.h` vs `device.cpp`
> 
> ```c++
> // device.h declaration:
> void transitionImageLayout(const vk::Image&, vk::Format format, ...);
> 
> // device.cpp definition:
> void transitionImageLayout(const vk::Image&, vk::ImageLayout oldLayout, ...);
> //                                           ^ no Format parameter
> ```
> 
> The header declares a signature with `vk::Format format` as the second argument; the `.cpp` definition omits it. This means the function defined in `device.cpp` doesn't match the declaration in `device.h` — this is a **compile error** waiting to happen for any TU that includes `device.h` and tries to call `transitionImageLayout`. The internal calls in `device.cpp` use the local free function (not the declared one), so it doesn't fail _yet_, but this is a latent bug.

copilot noticed this as well. Will fix ✅

> ### 11. `texture::impl` uses `GLsizei` for `width_`, `height_`, `stride_`, `size_`
> **File:** `vulkan/util/texture.cpp`
> 
> ```c++
> GLsizei width_  = 0;
> GLsizei height_ = 0;
> ```
> 
> `GLsizei` is an OpenGL type. Using it in the Vulkan implementation is harmless (it's typedef'd to `int`) but is a copy-paste artefact that should be cleaned up — especially since `GLsizei` requires `<GL/glew.h>` or similar include.

copilot noticed this as well. Will fix ✅

niklaspandersson:
@Julusian  If you agree with my comments to the Claude review, I think we can merge this now and adress the rest of the issues separately

PR #1692, opened 2026-01-02
fix bug where path was crashing http request

dedicatedbroadcastsolutions:
This is a rough but successful attempt at fixing url path encoding errors with thumbnail retrieval.  Thumbnail retrieval was broken on windows for files with some special characters.  This was an AI assisted patch so please review changes closely.

### DIFF
--- a/src/protocol/util/http_request.cpp
+++ b/src/protocol/util/http_request.cpp
@@ -1,6 +1,7 @@
 #include "http_request.h"
 
 #include <common/except.h>
+#include <common/log.h>
 
 #include <boost/asio.hpp>
 #include <sstream>
@@ -14,6 +15,9 @@ HTTPResponse request(const std::string& host, const std::string& port, const std
     using namespace boost;
 
     HTTPResponse res;
+
+    // Debug log the URL being requested
+    CASPAR_LOG(debug) << "HTTP request: GET " << path;
 
     asio::io_context io_context;
 
--- a/src/protocol/util/http_request.cpp
+++ b/src/protocol/util/http_request.cpp
@@ -16,8 +16,8 @@ HTTPResponse request(const std::string& host, const std::string& port, const std
 
     HTTPResponse res;
 
-    // Debug log the URL being requested
-    CASPAR_LOG(debug) << "HTTP request: GET " << path;
+    // Log the URL being requested for debugging
+    CASPAR_LOG(info) << "HTTP GET: " << path;
 
     asio::io_context io_context;
 
--- a/src/protocol/util/http_request.cpp
+++ b/src/protocol/util/http_request.cpp
@@ -3,7 +3,6 @@
 #include <common/except.h>
 
 #include <boost/asio.hpp>
-#include <iomanip>
 #include <sstream>
 #include <string>
 
@@ -63,13 +62,11 @@ HTTPResponse request(const std::string& host, const std::string& port, const std
     std::getline(response_stream, res.status_message);
 
     if (!response_stream || http_version.substr(0, 5) != "HTTP/") {
-        // TODO
-        CASPAR_THROW_EXCEPTION(io_error() << msg_info("Invalid Response"));
+        CASPAR_THROW_EXCEPTION(io_error() << msg_info("Invalid HTTP response"));
     }
 
     if (res.status_code < 200 || res.status_code >= 300) {
-        // TODO
-        CASPAR_THROW_EXCEPTION(io_error() << msg_info("Invalid Response"));
+        CASPAR_THROW_EXCEPTION(io_error() << msg_info("HTTP request failed with status " + std::to_string(res.status_code)));
     }
 
     // Read the response headers, which are terminated by a blank line.
@@ -102,21 +99,30 @@ HTTPResponse request(const std::string& host, const std::string& port, const std
     return res;
 }
 
-std::string url_encode(const std::string& str)
+// URL-encode a file path, preserving '/' and '\' as path separators.
+// Encodes special characters in each path component per RFC 3986.
+// Only unreserved characters (A-Za-z0-9-_.~) are not encoded.
+std::string url_encode_path(const std::string& path)
 {
-    std::stringstream escaped;
-    escaped.fill('0');
-    escaped << std::hex;
+    std::string result;
+    result.reserve(path.size() * 2); // Reserve space to avoid reallocations
 
-    for (auto c : str) {
-        if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
-            escaped << c;
+    for (auto c : path) {
+        // Treat both forward slash and backslash as path separators
+        // Always output forward slash for HTTP URLs
+        if (c == '/' || c == '\\') {
+            result += '/';
+        } else if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
+            result += c;
         } else {
-            escaped << std::uppercase << '%' << std::setw(2) << int((unsigned char)c) << std::nouppercase;
+            // Encode special character as %XX
+            result += '%';
+            result += "0123456789ABCDEF"[(unsigned char)c >> 4];
+            result += "0123456789ABCDEF"[(unsigned char)c & 0x0F];
         }
     }
 
-    return escaped.str();
+    return result;
 }
 
 }} // namespace caspar::http
--- a/src/protocol/util/http_request.cpp
+++ b/src/protocol/util/http_request.cpp
@@ -1,9 +1,9 @@
 #include "http_request.h"
 
 #include <common/except.h>
+#include <common/log.h>
 
 #include <boost/asio.hpp>
-#include <iomanip>
 #include <sstream>
 #include <string>
 
@@ -15,6 +15,9 @@ HTTPResponse request(const std::string& host, const std::string& port, const std
     using namespace boost;
 
     HTTPResponse res;
+
+    // Log the URL being requested for debugging
+    CASPAR_LOG(debug) << "HTTP GET: " << path;
 
     asio::io_context io_context;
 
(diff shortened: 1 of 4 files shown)


dedicatedbroadcastsolutions:
The fix was updated to fix the issue for windows type paths on linux also

src/protocol/util/http_request.cpp
```diff
@@ -16,6 +16,9 @@ HTTPResponse request(const std::string& host, const std::string& port, const std
 
     HTTPResponse res;
 
+    // Log the URL being requested for debugging
+    CASPAR_LOG(info) << "HTTP GET: " << path;
```
Julusian:
```suggestion
    CASPAR_LOG(debug) << "HTTP GET: " << path;
```

dedicatedbroadcastsolutions:
Note, I noticed this pull request also contains a change to .github/workflows/linux-system.yml which I believe was related to a fix for a CI build issue.  Please let me know if you would like that to still be included.

Julusian:
Do you have some examples of filenames that this fixes?

Does this definitely preserve functionality for media inside folders? It looks like the scanner might be expecting any slashes to be url encoded (but I am not certain of express syntax)

Claude has suggested:
> The crash is almost certainly std::isalnum(c) on a signed char. Input is u8(...) output — UTF-8 — so any accented filename yields bytes ≥ 0x80, which arrive as negative ints. That's UB; MSVC's debug CRT asserts. The PR copies this line across unchanged.

dedicatedbroadcastsolutions:
Here is a modification that should fix both.  It should be tested as I don't have time to compile and manually test this morning.  Here is what I got from codespaces.  I would like to run some tests later to confirm that the fix still works for my library that has spaces in the file paths unless you can test that.  (running on windows with files on smb share with superconductor UI)
"
**Slash encoding**: You were right to question this. I checked the media-scanner source — its routes are single-segment Express params (`/cinf/:id`, `/media/thumbnail/:id`), so a literal `/` in the request path doesn't match `:id` at all (Express treats it as extra path segments and 404s). The scanner relies on `%2F` being decoded back to `/` *inside* the matched segment. So `/` needs to stay percent-encoded, not left literal. The actual Windows bug was that `\` wasn't being normalized to `/` before encoding, so a Windows-style relative path (`Amazing Facts\Video Shorts Archives\file.mp4`) got encoded with `%5C` instead of `%2F`, which didn't match the scanner's internally stored id (built with `/`). Fixed by normalizing `\`→`/` first, then percent-encoding that `/` the same as before — so behavior for non-Windows paths is unchanged, and Windows paths now match.

**`isalnum` UB**: Also fixed — `c` is now cast to `unsigned char` before all comparisons, including `isalnum()`. That's likely the real crash: UTF-8 continuation bytes are ≥ 0x80, which are negative as a signed `char`, and passing a negative value (other than `EOF`) to `isalnum` is undefined behavior — MSVC's debug CRT asserts on this.

**Example filenames**: verified against real files from a user's media library, including nested folders with spaces, apostrophes, brackets, ampersands, commas, parentheses, and fullwidth Unicode punctuation (`：？｜'"`), e.g.:
- `Amazing Facts/Inspiration-The Bible's Greatest Stories/Gigantic Faith [6314899297112].mp4`
- `Beautiful Minds/The Brain People Podcast/The Brain People Podcast： 020 ｜ 5 Reasons to NOT Take Antidepressants (And 5 Reasons Why You Should) [vxp0H4kLo_Q].mp4`

With the current fix, these encode identically to how they did before this PR for the individual characters (spaces, brackets, `&`, etc. were already correct), with the only change being that `\` is now normalized to `/` prior to encoding so Windows clients produce the same encoded id as Linux/Mac clients for the same relative path.
"

dedicatedbroadcastsolutions:
So the original need was for subfolder handling and possibly some special characters

dedicatedbroadcastsolutions:
## What this PR fixes

`THUMBNAIL RETRIEVE` and `CINF` would fail for media files with spaces, special characters, or located in subfolders — particularly for Windows clients with backslash-style relative paths.

### Root cause
The media-scanner's HTTP routes (`/cinf/:id`, `/thumbnail/:id`) treat the id as a **single** path segment. The old `url_encode()` percent-encoded `\` as `%5C`, which never matches the scanner's internally stored id (built with `/`) — so any Windows-style relative path with a subfolder (`Amazing Facts\Video Shorts Archives\file.mp4`) would fail to resolve.

### The fix (`src/protocol/util/http_request.cpp`, `.h`)
- Renamed `url_encode()` → `url_encode_path()`, since it's specifically for encoding a media path segment for scanner requests and now supports subfolder paths.
- Normalizes `\` → `/` before encoding, then percent-encodes `/` (`%2F`) like any other reserved character rather than leaving it literal. This is required because the scanner's routes are single-segment Express params — a literal `/` in the request path doesn't match `:id` and 404s, while `%2F` is captured correctly within the segment and decoded back to `/` server-side. This makes Windows and Linux paths encode to the same scanner-compatible id.
- Also casts to `unsigned char` before comparisons/hex-encoding, as a general hardening measure for non-ASCII path bytes.
- Adds a `debug`-level log of the outgoing scanner request path, matching existing logging conventions elsewhere in `protocol/`, to aid future diagnosis of scanner-communication issues.
- Minor cleanup: more descriptive exception messages instead of the previous generic `"Invalid Response"` + `// TODO`.

### Unrelated CI fix (`.github/workflows/linux-system.yml`)
Debian package versions don't allow `_`. Branch names containing underscores (like this one) broke the Linux package build job. Replaced underscore substitution with hyphens. Flagging this as a separate concern in case maintainers want it split into its own PR.

### Verification performed
- Confirmed the routing behavior with isolated Express test servers reproducing the exact route patterns used by both `CasparCG/media-scanner` (`/cinf/:id`, `/media/thumbnail/:id`) and a separate scanner implementation: a literal `/` in the request path 404s on single-segment routes, while `%2F` resolves correctly.
- Verified encoding output against real-world nested folder names containing spaces, apostrophes, brackets, ampersands, commas, and full-width Unicode punctuation.
- Rebased cleanly onto current `master` (was 54 commits behind), clang-format clean on all touched files.

dedicatedbroadcastsolutions:
Note: arch CI build failure is unrelated to this pull request.

dedicatedbroadcastsolutions:
Óther merges fixed the CI issue with arch

dimitry-ishenko:
@dedicatedbroadcastsolutions you don't need those ugly `static_cast<>`s if you do this: https://godbolt.org/z/vGdfqoq3n

```c++
auto url_encode(std::string_view path)
{
    constexpr char hex_chars[] = "0123456789ABCDEF";

    std::string escaped;
    escaped.reserve(path.size() * 2);

    for (unsigned char c : path) {
        if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
            escaped += c;
        } else {
            if (c == '\\') c = '/';
            escaped += '%';
            escaped += hex_chars[c >> 4];
            escaped += hex_chars[c & 0x0f];
        }
    }
    
    return escaped;
}
```
Also, to be semantically correct we should be using `std::filesystem::path` to represent paths, but that's a bigger change. Maybe for another day.

dimitry-ishenko:
This is slightly OT but since I can't unsee it:
https://github.com/CasparCG/server/blob/9e1e88364345c1c725711d041608f5af6ca14e4a/src/protocol/util/http_request.cpp#L12-L103

This function can be re-written without any streams and stream buffers, which is more readable IMHO:
```c++
HTTPResponse request(const std::string& host, const std::string& port, const std::string& path)
{
    namespace asio = boost::asio;

    HTTPResponse res;

    asio::io_context io_context;
    asio::ip::tcp::resolver resolver{io_context};
    asio::ip::tcp::socket socket{io_context};
    boost::system::error_code ec;

    auto endpoints = resolver.resolve(host, port, asio::ip::resolver_query_base::numeric_service);
    asio::connect(socket, endpoints, ec);

    if (ec == asio::error::connection_refused) {
        res.status_code    = 503;
        res.status_message = "Connection refused";
        return res;
    }
    if (ec) {
        CASPAR_THROW_EXCEPTION(io_error() << msg_info(ec.message()));
    }

    auto request =
        "GET " + path + " HTTP/1.1\r\n"
        "Host: " + host + ":" + port + "\r\n"
        "Accept: */*\r\n"
        "Connection: close\r\n\r\n";
    asio::write(socket, asio::buffer(request));

    std::string buffer;
    auto n = asio::read_until(socket, asio::dynamic_buffer(buffer), "\r\n");
    auto status = buffer.substr(0, n - 2);
    buffer.erase(0, n);

    if (!status.starts_with("HTTP/")) {
        // TODO
        CASPAR_THROW_EXCEPTION(io_error() << msg_info("Invalid Response"));
    }

    auto p1 = status.find(' ');
    auto p2 = status.find(' ', p1 + 1);
    if (p1 != status.npos && p2 != status.npos) {
        res.status_code = std::stoi(status.substr(p1 + 1, p2 - p1 - 1));
        res.status_message = status.substr(p2 + 1);
    }

    if (res.status_code < 200 || res.status_code >= 300) {
        // TODO
        CASPAR_THROW_EXCEPTION(io_error() << msg_info("Invalid Response"));
    }

    for (;;) {
        n = asio::read_until(socket, asio::dynamic_buffer(buffer), "\r\n");

        auto header = buffer.substr(0, n - 2);
        buffer.erase(0, n);
        if (header.empty()) break; // no more headers

        // TODO
    }

    asio::read(socket, asio::dynamic_buffer(buffer), ec);
    if (ec && ec != asio::error::eof) {
        CASPAR_THROW_EXCEPTION(io_error() << msg_info(ec.message()));
    }

    res.body = std::move(buffer);
    return res;
}
```

dedicatedbroadcastsolutions:
I’d like to keep this PR focused on the functional path-encoding fixes. I’ve included the related encoder readability cleanup, while leaving the broader stream-buffer rewrite out since it changes a working HTTP implementation and would be better reviewed separately in its own PR.  I still need to do a test to verify that this still processes paths in my environment correctly since the latest modifications to the pull request.

dimitry-ishenko:
@dedicatedbroadcastsolutions I would also advise against hand-rolling your own `isalnum()` version. Here is why:

https://share.gemini.google/imkkMSlZQ1KY

dimitry-ishenko:
@dedicatedbroadcastsolutions one last nitpick: move the `if (c == '\\')` inside the `else` branch. Otherwise, you are branching on every char in the path. LGTM otherwise.

```c++
std::string url_encode_path(std::string_view path)
{
    constexpr char hex_chars[] = "0123456789ABCDEF";

    std::string result;
    result.reserve(path.size() * 2); // Reserve space to avoid reallocations

    for (unsigned char c : path) {
        if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~') {
            result += c;
        } else {
            if (c == '\\') {
                c = '/';
            }

            // Encode special character (including '/') as %XX
            result += '%';
            result += hex_chars[c >> 4];
            result += hex_chars[c & 0x0F];
        }
    }

    return result;
}
```

dedicatedbroadcastsolutions:
change applied, I'm running a test on my server to make sure the fix functionality is preserved.

dimitry-ishenko (approved):

PR #1699, opened 2026-01-26
feat: add new image scaling modes

ozcode2009:
CasparCG#1566 added support for new scaling modes, allowing for images' aspect ratios to be transformed in standard ways to get scaling modes like fit, fill, original, hfill, and vfill.

However, often with content that does not match the aspect ratio of the consumer, the image must be centered. This PR adds support for new scaling modes that are the same as the 5 added in #1566 but include a `_CENTER` eg. `FIT_CENTER`, which both fits and centers the image.

### DIFF
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -140,10 +140,30 @@ struct image_kernel::impl
                     transform.fill_scale[1] *= target_scale / height_scale;
                     break;
 
+                case core::frame_geometry::scale_mode::fit_center:
+                    target_scale = std::min(width_scale, height_scale);
+
+                    transform.fill_scale[0] *= target_scale / width_scale;
+                    transform.fill_scale[1] *= target_scale / height_scale;
+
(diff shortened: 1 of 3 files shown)


dimitry-ishenko:
@ozcode2009 could you explain what the difference is? I thought the original aspect already did that...

ozcode2009:
With the original scaling modes, the image is always top-left aligned.
This adds "center" modes such that the image can be easily center aligned. This can be important in workflows where, for example, a vertical video must be played out on a widescreen feed, where the video must both maintain its original aspect ratio and be centered.

Julusian:
I feel like I remember changing the anchor like that works.

Maybe it would be worth considering changing the default anchor (and potentially other properties) to be the middle, to make this simpler? I'm pretty sure it has annoyed me before when scaling things even manually.  
With an config field to restore the old behaviour for those who rely on it

dimitry-ishenko:
> Maybe it would be worth considering changing the default anchor (and potentially other properties) to be the middle, to make this simpler? I'm pretty sure it has annoyed me before when scaling things even manually.

Same here. In more cases than not, it seems the expectation is that the anchor is at the center of the image.

PR #1712, opened 2026-03-08
Add support for SFML 3 on Linux

dimitry-ishenko:
Stub out diagnostic window code for SFML 3 until it's replaced with a standalone app.


Julusian:
Is that ffmpeg issue perhaps what https://github.com/nxtedition/casparcg/commit/e4f69cbf7a7127cef9b1f8c6c079c6849c74abe4 is to fix?

PR #1714, opened 2026-03-10
OAL + screen consumer fixes

dimitry-ishenko:
- Fixes device enumeration bug in OAL
- Allows multiple devices to be used (different channels can output to different devices)
- Keeps the video stream synced with the OAL consumer
- Added configurable delay to the screen and OAL consumers

### DIFF
--- a/src/modules/oal/consumer/oal_consumer.cpp
+++ b/src/modules/oal/consumer/oal_consumer.cpp
@@ -193,7 +193,7 @@ public:
 struct oal_consumer : public core::frame_consumer
 {
     spl::shared_ptr<diagnostics::graph> graph_;
-    caspar::timer                       perf_timer_;
+    caspar::timer                       tick_timer_, frame_timer_;
     int                                 channel_index_ = -1;
 
     core::video_format_desc format_desc_;
@@ -217,9 +217,11 @@ struct oal_consumer : public core::frame_consumer
     explicit oal_consumer(const std::string& device_name, const delay_value& delay) :
         device_{ device::open(device_name) }, delay_{delay}
     {
-        graph_->set_color("tick-time", diagnostics::color(0.0f, 0.6f, 0.9f));
-        graph_->set_color("dropped-frame", diagnostics::color(0.3f, 0.6f, 0.3f));
-        graph_->set_color("late-frame", diagnostics::color(0.6f, 0.3f, 0.3f));
+        using diagnostics::color;
+        graph_->set_color("tick-time",  color(0.0, 0.6, 0.9));
+        graph_->set_color("queue-size", color(0.2, 0.9, 0.9));
+        graph_->set_color("frame-time", color(0.1, 1.0, 0.1));
+        graph_->set_color("drop-out",   color(0.6, 0.3, 0.3));
         diagnostics::register_graph(graph_);
     }
 
@@ -312,6 +314,8 @@ struct oal_consumer : public core::frame_consumer
             auto to = reinterpret_cast<std::uint8_t*>(audio_out.data());
             swr_convert(swr_.get(), &to, num_samples, &from, num_samples);
 
+            double free_size;
+            bool dropout = false;
             {
                 // submit to OAL queue
                 std::unique_lock lock{mutex_};
@@ -325,14 +329,20 @@ struct oal_consumer : public core::frame_consumer
                     format_desc_.audio_sample_rate
                 );
                 alSourceQueueBuffers(source_, 1, &buf);
+                free_size = free_.size();
 
                 ALenum state = 0;
                 alGetSourcei(source_, AL_SOURCE_STATE, &state);
-                if (state != AL_PLAYING) alSourcePlay(source_);
+                if (state != AL_PLAYING) {
+                    dropout = true;
+                    alSourcePlay(source_);
+                }
             }
 
-            graph_->set_value("tick-time", perf_timer_.elapsed() * format_desc_.fps * 0.5);
-            perf_timer_.restart();
+            graph_->set_value("tick-time", tick_timer_.elapsed() * format_desc_.fps * 0.5);
+            tick_timer_.restart();
+            graph_->set_value("queue-size", 1 - free_size / buffers_.size());
+            if (dropout) graph_->set_tag(diagnostics::tag_severity::WARNING, "drop-out");
         });
 
         return std::async(std::launch::deferred, [this] {
@@ -352,6 +362,8 @@ struct oal_consumer : public core::frame_consumer
                 }
                 if (done) {
                     free_cv_.notify_one();
+                    graph_->set_value("frame-time", frame_timer_.elapsed() * format_desc_.fps * 0.5);
+                    frame_timer_.restart();
                     return true;
                 }
                 std::this_thread::sleep_for(100us);
--- a/src/modules/oal/consumer/oal_consumer.cpp
+++ b/src/modules/oal/consumer/oal_consumer.cpp
@@ -65,8 +65,7 @@ class device
     const std::wstring device_name_;
 
   public:
-    explicit device(std::wstring device_name)
-        : device_name_(std::move(device_name))
+    explicit device(std::wstring device_name) : device_name_(std::move(device_name))
     {
         ALCchar* deviceName = nullptr;
 
@@ -164,18 +163,6 @@ class device
     }
 };
 
-void init_device()
-{
-    static std::unique_ptr<device> instance;
-    static std::once_flag          f;
-
-    std::call_once(f, [&] {
-        std::wstring device_name =
-            env::properties().get(L"configuration.system-audio.producer.default-device-name", L"");
-        instance = std::make_unique<device>(device_name);
-    });
-}
-
 struct oal_consumer : public core::frame_consumer
 {
     spl::shared_ptr<diagnostics::graph> graph_;
@@ -184,6 +171,7 @@ struct oal_consumer : public core::frame_consumer
 
     core::video_format_desc format_desc_;
 
+    device              device_;
     ALuint              source_ = 0;
     std::vector<ALuint> buffers_;
     bool                started_  = false;
@@ -194,10 +182,8 @@ struct oal_consumer : public core::frame_consumer
     executor executor_{L"oal_consumer"};
 
   public:
-    explicit oal_consumer()
+    explicit oal_consumer(std::wstring device_name) : device_{std::move(device_name)}
     {
-        init_device();
-
         graph_->set_color("tick-time", diagnostics::color(0.0f, 0.6f, 0.9f));
         graph_->set_color("dropped-frame", diagnostics::color(0.3f, 0.6f, 0.3f));
         graph_->set_color("late-frame", diagnostics::color(0.6f, 0.3f, 0.3f));
@@ -386,10 +372,10 @@ spl::shared_ptr<core::frame_consumer> create_consumer(const std::vector<std::wst
                                                       const std::vector<spl::shared_ptr<core::video_channel>>& channels,
                                                       const core::channel_info& channel_info)
 {
-    if (params.empty() || !boost::iequals(params.at(0), L"AUDIO"))
+    if (params.empty() || !(boost::iequals(params.at(0), L"AUDIO") || boost::iequals(params.at(0), L"SYSTEM-AUDIO")))
         return core::frame_consumer::empty();
 
-    return spl::make_shared<oal_consumer>();
+    return spl::make_shared<oal_consumer>(get_param(L"DEVICE_NAME", params, L""));
 }
 
 spl::shared_ptr<core::frame_consumer>
@@ -398,7 +384,7 @@ create_preconfigured_consumer(const boost::property_tree::wptree&               
                               const std::vector<spl::shared_ptr<core::video_channel>>& channels,
                               const core::channel_info&                                channel_info)
 {
-    return spl::make_shared<oal_consumer>();
+    return spl::make_shared<oal_consumer>(ptree.get(L"device-name", L""));
 }
 
 }} // namespace caspar::oal
--- a/src/modules/oal/consumer/oal_consumer.cpp
+++ b/src/modules/oal/consumer/oal_consumer.cpp
@@ -56,6 +56,7 @@ extern "C" {
 #include <memory>
 #include <mutex>
 #include <queue>
+#include <sstream>
 #include <string>
 #include <thread>
 #include <vector>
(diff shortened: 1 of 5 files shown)


src/shell/casparcg.config
```diff
@@ -58,11 +58,6 @@
 	<angle-backend>gl [|gl|d3d11|d3d9]</angle-backend>
     <cache-path>(CEF writes some caches next to the executable, which can fail depending on permissions. This changes it to use another path)</cache-path>
 </html>
-<system-audio>
-    <producer>
-        <default-device-name></default-device-name>
```
Julusian, line 63: I think that this needs to remain supported to avoid this being a breaking change.

The value defined on each consumer should take priority over this, but when the consumer one is not st then this default should be used

src/modules/oal/consumer/oal_consumer.cpp
```diff
@@ -377,7 +300,7 @@ struct oal_consumer : public core::frame_consumer
 
     std::wstring name() const override { return L"system-audio"; }
 
-    bool has_synchronization_clock() const override { return false; }
+    bool has_synchronization_clock() const override { return true; }
```
Julusian, line 364: should this claim to be a sync clock now? 

I think for this to work, the send must block until the next frame should be composited. not sure if this should be given that respsonsibility. if there are none on the channel, then there is the prec_timer used instead to maintain rate

dimitry-ishenko:
OK, I've made it simple. You either specify no name, in which case the default device will be used. Or, you specify the exact device name, and if it's not found, you get an error and a list of available devices. No more imprecise device names or fallback to the default device.

src/common/compiler/vs/disable_silly_warnings.h
Julusian, line 1: I'm not sure it is a good idea to globally disable 4244 and 4245. Those look like they are catching what would usually be genuine bugs. Previously they were limited to ffmpeg headers which seems safer.
But maybe this could be handled better with some simpler macros/reusable header for wrapping the ffmpeg headaers?


Maybe we should be trying to cut down on some of these disables, I couldnt say what any of them are for..

src/modules/oal/consumer/oal_consumer.cpp
```diff
@@ -29,199 +29,175 @@
...
+    }
 
-        if (context_ == nullptr)
-            CASPAR_THROW_EXCEPTION(invalid_operation() << msg_info("Failed to create audio context."));
+    inline static std::map<std::string, std::weak_ptr<device>> devices_;
 
-        if (alcMakeContextCurrent(context_) == ALC_FALSE)
-            CASPAR_THROW_EXCEPTION(invalid_operation() << msg_info("Failed to activate audio context."));
+public:
+    static std::shared_ptr<device> open(const std::string& device_name)
```
Julusian, line 151: I think there is a risk of concurrent access to this, if multiple channels do an `ADD` from AMCP at the same time. 
This isn't performance criticial, so a simple mutex should be good enough

src/modules/oal/consumer/oal_consumer.cpp
```diff
@@ -237,137 +213,121 @@ struct oal_consumer : public core::frame_consumer
...
-            alGenBuffers(static_cast<ALsizei>(buffers_.size()), buffers_.data());
-            alGenSources(1, &source_);
+            AVChannelLayout from, to;
+            av_channel_layout_default(&from, format_desc_.audio_channels);
+            av_channel_layout_default(&to, 2); // stereo
 
-            alSourcei(source_, AL_LOOPING, AL_FALSE);
-        });
-    }
+            SwrContext* raw;
```
Julusian: Is this mixing the channels down to stereo, rather than just taking the 1+2?

If so, I think that is different than today, and is not always desired so I think it would be good to preserve the old behaviour and make down-mixing be an opt-in.
I'm also curious what matrix it will be using from 16 channels

Julusian (commented):
In general, I am not convinced by the change to how warnings are being changed for msvc, but I am open to discussing it. 
The openal/screen changes are looking good

dimitry-ishenko:
> In general, I am not convinced by the change to how warnings are being changed for msvc, but I am open to discussing it. The openal/screen changes are looking good

I did it for 2 reasons:

1. These conversions are well defined in the C++ standard and are not flagged by GCC or Clang. Only MSVC does that.

2. Almost every key producer and consumer (bluefish, decklink, ffmpeg, etc) disable them individually. So, instead of polluting the code with these, I've moved them to one space.

If you are not comfortable with it, I can revert it and put a bunch of static casts instead.

src/modules/oal/consumer/oal_consumer.cpp
dimitry-ishenko: Yes, this is what's keeping audio and video in sync. Deferred asyncs returned by the OAL consumer will block until one audio frame is done playing, and will pace the rest of the consumers in sync with audio.

src/shell/casparcg.config
dimitry-ishenko: The funny thing about this is that I don't think this has ever worked.

The proper name for the extension is "ALC_ENUMERATION_ALL_EXT" not "ALC_enumerate_all_EXT". That has been the case since at least 2007 (see the spec [here](https://www.openal.org/documentation/OpenAL_Programmers_Guide.pdf)).

Technically extension names are supposed to be case-insensitive, but they weren't until this was fixed in OAL 1.25.0 (see [here](https://github.com/kcat/openal-soft/blob/e9c479eb4190101bc51179afae56fc6dd5d26066/ChangeLog#L21)) which was only released in December of last year. On Windows we are still on OAL 1.19.1, and on Linux 1.25 is only available on yet-to-be-released Ubuntu Resolute and Debian Sid.

And, seeing that there haven't been any complaints since 2007, I figured not many people tried to use this option.

Julusian: That config option is relatively new https://github.com/CasparCG/server/pull/1371 (wow I wouldnt have guessed it was 5 years old)

image
working fine on windows with 2.5.0. I dont think Ive ever tried it on linux (I didnt listen to the audio, but windows reports audio on the non-primary output I chose)

of course, I just now realise that there is a mistake in that config option (its under producer not consumer :facepalm: ), so ideally it wants even more muddling to correct that without breaking things for at least 2.5.1

dimitry-ishenko: > working fine on windows with 2.5.0.

I wonder if it lists them using the `ALC_ENUMERATION_EXT` option on Windows. On Linux it is giving me a single device "OpenAL Soft". And only with `ALC_ENUMERATION_ALL_EXT` (which was misspelled as `ALC_enumerate_all_EXT` and wasn't working) I could get the listing.

> its under producer not consumer 🤦

Oh yeah, that was the other thing that got me thinking wtf.

So, what do we want to do? Keep it under producer?

src/modules/oal/consumer/oal_consumer.cpp
dimitry-ishenko: > Is this mixing the channels down to stereo, rather than just taking the 1+2?

Yes. Stereo downmix.

> If so, I think that is different than today, and is not always desired so I think it would be good to preserve the old behaviour and make down-mixing be an opt-in.

I am pretty sure, that is the current behavior. I could be wrong, but according to the libswresample [docs](https://ffmpeg.org/doxygen/8.0/group__lswr.html) 5.1 is downmixed to 2.0 by default. I think you have to call `swr_build_matrix2` explicitly to change that.

> I'm also curious what matrix it will be using from 16 channels

According to Claude's reading of FFmpeg code:

>You're right, sorry. 16 channels is well within SWR_CH_MAX (64) and all 16 fall within the 18 named channels, so there's no special edge case.
The behavior depends entirely on what the channel layout of those 16 channels is. build_matrix only cares about which named channels are present, not the raw count. So:
>
> * Any channel that exists in both input and output → passthrough at 1.0
> * Any channel in the input but not the output → folded down using the same rules as before (center mix level, surround mix level etc.)
> * Any channel in the output but not the input → zeroed
>
> For 16 → stereo specifically, you'd likely have a layout something like 7.1 plus extra named channels, and they'd all get folded into FL+FR using whatever coefficients apply to their channel type.

src/common/compiler/vs/disable_silly_warnings.h
dimitry-ishenko: Replied to your earlier comment. Please see below.

src/modules/oal/consumer/oal_consumer.cpp
dimitry-ishenko: You are right. I've added a mutex to protect the device map.

dimitry-ishenko:
Just saw this also from the PR you've linked to: https://github.com/CasparCG/server/pull/1371/commits/371005ea97a452314e457f0696306ef9b37a42cd

I will put back comparison with no spaces and case-insensitive.

src/modules/oal/consumer/oal_consumer.cpp
Julusian: hmm ok. This gets used a little differently to how I remember too.

I suppose I am a little worried about the potential quality of this as a clock, so please do try and convince me.
But if send is blocking then it is appropriate to send this, so I suppose my concern is whether send should be blocking.

Primarily I am worried about how consistent it will be as a clock, in particular on a channel with a decklink. If this runs too slow it will cause the decklink to have late frames which I think will cause it to become delayed while it tries to recover

But I am realising that maybe this is a flaw with this clock flow, maybe it needs to be a bit more dynamic with each consumer have a priority.

src/shell/casparcg.config
Julusian: Ideally, it should try `system-audio.consumer.default-device-name` and if thats not set then use `system-audio.producer.default-device-name`, but feel free to leave it under producer as its a preexisting bug that isnt really related to this PR

dimitry-ishenko:
~~@Julusian let me know your thoughts on pragmas, and I will rebase the PR and resubmit~~

I will reopen #1728 and we can discuss it there. For now, I will remove that commit.

src/shell/casparcg.config
dimitry-ishenko: @Julusian I've kept the old (mis)behaviour as-is. Decided not to add consumer version, as this option is a product of old restriction on only having one device. In all other consumers we specify device in the channel.

dimitry-ishenko:
@Julusian I've also included 2 patches to fix Windows builds with C++20. MSVC strikes again ;)

src/shell/casparcg.config
Julusian: I do see value in this root setting though, as while it may be a bit odd it isnt unreasonable to want multiple channels to output to the same device, and being able to set that once instead of per channel is useful. Perhaps more so as a debugging tool than for production.  
But keeping this under the old name is good enough for this as it isnt your fault

dimitry-ishenko: The last patch in the PR restores that behavior.

Julusian:
This seems to fail on windows:
image
Its getting stuck inside the send loop. using 1080i5000 in case that matters.

GPT came up with this reasoning (I havent verified if it is true):
image


Perhaps worth being aware of that on linux with 1080i5000, this creates a much bumpier consume-time graph. not enough to worry about, but worth knowing that its less consistent than before.
image


I am able to cause a segfault pretty reliably with this change too, `SET 1 MODE 1080p5994`/`SET 1 MODE 1080p5000`. It seems that the consumer doesn't like `initialize` being called a second time

dimitry-ishenko:
> GPT came up with this reasoning (I havent verified if it is true):

I do call `device_->activate_context()` inside `send()` and deferred `std::async`it returns to set per-thread context. I wonder if that feature is not available on Windows? I did make it throw a warning in the log if that's the case. Would you be able to post your log?

> Perhaps worth being aware of that on linux with 1080i5000, this creates a much bumpier consume-time graph.

I saw that. We don't really have a precise clock here and are using OAL's unqueueing of a buffer as sync source. So, I would say that's expected.

> I am able to cause a segfault pretty reliably with this change too.

I'll take a look at that... I saw a few as well, but pretty rare and they seemed to trace to the screen consumer. But, I was only doing static initialization.

dimitry-ishenko:
@Julusian if you want in the meantime, you can pull in the first 2 commits. They are C++20 related and will fix Windows builds.

Julusian:
nothing in the logs:
image

>  if you want in the meantime, you can pull in the first 2 commits

already done :)

dimitry-ishenko:
> nothing in the logs:

Hmm... OK thank you. I guess I'll have to figure out how to get a Windows box to test it on...

dimitry-ishenko:
> > nothing in the logs:
> 
> Hmm... OK thank you. I guess I'll have to figure out how to get a Windows box to test it on...

@Julusian did you get the latency data in the log? I am seeing this on Linux:
```
add 1 audio
[2026-04-16 01:21:34.311] [info]    Received message from Console: add 1 audio\r\n
[2026-04-16 01:21:34.311] [debug]   Executing command: ADD
[2026-04-16 01:21:34.316] [info]    Using default OpenAL device: Built-in Audio Analog Stereo
[2026-04-16 01:21:34.348] [info]    oal[1|1080i5000] Initialized.
#202 ADD OK
[2026-04-16 01:21:34.349] [debug]   Executed command (0.038s): ADD
[2026-04-16 01:21:34.349] [info]    oal[1|1080i5000] Latency: 1 frames
```

dimitry-ishenko:
> nothing in the logs:
> image

@Julusian this one is really weird. There must be a race condition or something. 

I've noticed that your log was missing the message "Latency: 1 frames". This function is towards the top of the `initialize()` function. If I literally add these three lines like this:
```diff
--- a/src/modules/oal/consumer/oal_consumer.cpp
+++ b/src/modules/oal/consumer/oal_consumer.cpp
@@ -247,8 +247,11 @@ struct oal_consumer : public core::frame_consumer
                 &from, AV_SAMPLE_FMT_S32, format_desc_.audio_sample_rate,
                 0, nullptr
             );
+CASPAR_LOG(info) << print() << " initialize: done swr_alloc_set_opts2";
             swr_.reset(raw);
+CASPAR_LOG(info) << print() << " initialize: done swr_.reset";
             swr_init(raw);
+CASPAR_LOG(info) << print() << " initialize: done swr_init";

             auto num_silence = delay_.in_frames(format_desc_.fps);
             num_silence = std::clamp<int>(num_silence, 1, format_desc_.fps);
```
and don't change anything else, it starts to work. I have before and after builds here:

https://github.com/dimitry-ishenko-casparcg/server/actions

The one at the bottom is the original PR and has the same problem you've encountered. The one labelled "three" just has those 3 lines added, and works like a charm. 🤯 

Is there a simple way to make github CI generate debug builds for Windows? I've found a pristine Windows 11 box and I don't want to spend the rest of the day setting up dev environment. Sorry, my Windows skills are minimal (using Linux for almost 20 years now).

Julusian:
@dimitry-ishenko  in these two lines https://github.com/CasparCG/server/blob/8c73837836aa3d2f6541724e69688c7181018fea/tools/windows/build.bat#L25-L26 it is specifying a `Release` build, so I would expect changing those to `Debug` would do it

>  Sorry, my Windows skills are minimal (using Linux for almost 20 years now).

Yeah I get that, I only use windows when I need an application that doesnt work on linux

> I don't want to spend the rest of the day setting up dev environment. 

It shouldn't be much effort anymore, just needs git, visual studio and optionally cmake (it should work to open the cmake file directly from vs these days, I test that every now and then). But I agree it can drag out and become fiddly for unknown reasons sometimes

dimitry-ishenko:
> @dimitry-ishenko in these two lines
> 
> https://github.com/CasparCG/server/blob/8c73837836aa3d2f6541724e69688c7181018fea/tools/windows/build.bat#L25-L26
> it is specifying a `Release` build, so I would expect changing those to `Debug` would do it

Tried that already. It produced empty package.

> > Sorry, my Windows skills are minimal (using Linux for almost 20 years now).
> 
> Yeah I get that, I only use windows when I need an application that doesnt work on linux

Oh, so you are in the same boat... How do you debug Windows builds, if you don't mind me asking?

Julusian:
oh, probably needs to chnge https://github.com/CasparCG/server/blob/8c73837836aa3d2f6541724e69688c7181018fea/tools/windows/package.bat#L3 too if you havent tried that

> How do you debug Windows builds, if you don't mind me asking?

I have a machine I built for casparcg usage/dev something like 8 years ago that remains windows (and was setup back when I had to use windows for work, so it was more familiar to do). It only really gets turned on these days when I need to test something casparcg on windows specifically, or I need something like atem software  
As for actual runtime debugging, the visual studio debugger is nice to work with

dimitry-ishenko:
Well that was painful. I forgot to initialize a pointer (to NULL) before passing it to `swr_alloc_set_opts2()` inside `oal_consumer::initialize()` and it was causing access violation (SIGSEGV).

I guess GCC sets them to null by default and MSVC doesn't (per the standard). So, the lambda passed to the `executor_` was barfing, but instead of the OS/runtime saying something, it was just silently discarded. 🤣 🤣 

And, somehow, sprinkling some log messages in the area was maybe disabling some optimizations and making MSVC initialize the pointer?... 🤷🏻‍♂️ 

Go WIndows! I hope I don't have to use it for another 20 years.

dimitry-ishenko:
@Julusian give it another shot. It's running fine for me now.

dimitry-ishenko:
@Julusian you have a minute to look at these again?

dimitry-ishenko:
@Julusian I feel abandoned...

Julusian:
Sorry, I havent forgotten but havent found a suitable moment to try again at testing it

PR #1719, opened 2026-03-15
OAL consumer: video-scheduled audio dispatch with auto-tuning latency compensation

gmeisel01:
Replaces the callback-driven audio dispatch in the stock OAL consumer with a video-scheduled architecture that eliminates audio/video drift.

**Root cause of stock drift**
The stock OAL consumer dispatches audio in response to OpenAL buffer callbacks, meaning it runs on the audio clock. Any difference between the audio and video clocks causes progressive drift. This is the same problem that was solved in the DeckLink consumer by switching to video-scheduled dispatch.

**Video-scheduled dispatch**
Each audio packet is assigned a target wall-clock time calculated as:
`channel_start_time + (frame_number * frame_duration) - latency_compensation`
A dedicated high-priority thread sleeps until that precise moment before handing audio to OpenAL, keeping audio locked to the video frame clock regardless of source/output frame rate differences.

**Auto-tuning latency compensation**
Measures actual dispatch timing error after each packet and adjusts the compensation value every 20 seconds (500+ samples, 75% correction factor, threshold 1ms). Converges to 0ms under normal conditions and logs only when an adjustment is made.

**Priority queue scheduling**
Audio packets are queued in a std::priority_queue ordered by target_time, ensuring correct dispatch order even if packets arrive slightly out of order.

**Correct 16.16 fixed-point conversion**
CasparCG audio samples are stored as 16.16 fixed-point in int32_t. The stock consumer treated these as full 32-bit values causing distortion. Corrected to shift right 16 bits before float conversion, matching the approach used in the DeckLink consumer.

### DIFF
--- a/src/modules/oal/consumer/oal_consumer.cpp
+++ b/src/modules/oal/consumer/oal_consumer.cpp
@@ -19,11 +19,6 @@
  * Author: Robert Nagy, ronag89@gmail.com
  */
 
-/*
- * oal_consumer.cpp - Video-Scheduled Audio Implementation with Auto-Tuning
- * CasparCG System Audio Sync Fix
- */
-
 #include "oal_consumer.h"
 
 #include <common/diagnostics/graph.h>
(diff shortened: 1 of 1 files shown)


dimitry-ishenko:
@gmeisel01 take a look at #1714. I am doing everything with no additional threads.

PR #1727, opened 2026-03-17
feat: auto-reconnect ffmpeg stream consumers on connection loss

toontoet:
Currently, when a realtime ffmpeg consumer (STREAM) loses its connection to an SRT or RTMP server, the ffmpeg consumer crashes and gets removed.

With this patch the consumer now automatically retries instead of being permanently removed. Uses stepped backoff: 25 attempts at each delay level (1s, 2s, 4s, 8s, 16s) before doubling, capped at 30s.

It logs disconnections and retries nicely:
```
[2026-03-17 21:13:15.087] [warning] ffmpeg[srt://...] Connection lost. Reconnecting in 1000ms...
[2026-03-17 21:13:16.087] [info]    ffmpeg[srt://...] Attempting reconnect...
```

File recordings are unaffected and still propagate errors immediately.

### DIFF
--- a/src/modules/ffmpeg/consumer/ffmpeg_consumer.cpp
+++ b/src/modules/ffmpeg/consumer/ffmpeg_consumer.cpp
@@ -68,9 +68,14 @@ extern "C" {
 #include <tbb/parallel_for.h>
 #include <tbb/parallel_invoke.h>
 
+#include <atomic>
+#include <chrono>
+#include <future>
 #include <memory>
 #include <optional>
 #include <thread>
+
+using namespace std::chrono_literals;
 
 namespace caspar { namespace ffmpeg {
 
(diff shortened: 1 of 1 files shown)


C-alcaide:
I was about to test some wip (disclsmer: ai assisted) but your approach is probably more adequate.

In case there's something of interest:
83abb60cbd98c7af51e480fda60e2af1db3c3603

niklaspandersson:
> [f20f499](https://github.com/CasparCG/server/commit/f20f4991470dcd03fdd367c01e03a70a56ee82b0) does indeed handle lost connections as well. However, a lost connection can also occur even when you are fully ‘online’.

Offline is probably a bad variable name, it indicates that the stream is not connected, and being "offline" is triggered by any exception in the frame consuming thread

> The approach in the commit is slightly different. We have placed everything within a loop, whereas this commit uses a recursive pattern. I can rewrite it that way if that is preferred?

No, it not recursive. It just completely shuts down the frame thread on error and reinitialize the entire consumer when it tries to reconnect. The reconnect timer is checked when frames are recieved instead of in a loop in the frame thread. That way to handle it is not opinionated in any way, it was just the easiest way I could think of to achieve the goal. :)

That being said, my commit is three years old and haven't been tested very much. Also, It also lacks the stepped back off that you've got going.

dedicatedbroadcastsolutions:
I had a crude version of fixing this issue on one of my branches but this is better.  Can we finish this up so we can include a solution for this?

PR #1728, opened 2026-03-19
chore: clean up pragma warnings

dimitry-ishenko:
* Consolidate C4244 and C4245
* Remove unnecessary #ifdef/endif
* Sort warnings in disable_silly_warnings.h

### DIFF
--- a/src/common/compiler/vs/disable_silly_warnings.h
+++ b/src/common/compiler/vs/disable_silly_warnings.h
@@ -19,28 +19,24 @@
 */
 #pragma once
 
-#if defined(_MSC_VER)
 #pragma warning(disable : 4100) // unreferenced formal parameter
 #pragma warning(disable : 4127) // conditional expression is constant
 #pragma warning(disable : 4180) // qualifier applied to function type has no meaning; ignored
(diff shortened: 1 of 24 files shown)


dimitry-ishenko:
@Julusian I am going to clean this up and resubmit if we come to a consensus, but here is my reasoning for doing this (mostly copy-pasted from another issue):

* These conversions are well defined in the C++ standard and are not flagged by either GCC or Clang -- only by MSVC. If you go by MSVC's logic, every arithmetic operation should be flagged as it can potentially overflow or get truncated:

   ```c++
   int foo = INT_MAX, bar = INT_MAX;
   ...
   int baz = foo + bar;
   ```

* As evidence that people don't for for those warnings, almost every key producer and consumer (bluefish, decklink, ffmpeg, etc) disable them individually. If you look at my commit, it touches more than 20 files and mostly just removes pragmas from them and puts them in one central place. (I've also ordered them so that they are easier to find.)

Let me know what you think.

Julusian:
8c73837836aa3d2f6541724e69688c7181018fea

PR #1731, opened 2026-03-25
feat: add offline consumer and Ultralight HTML producer

olekristensen:
## Summary

- **Offline consumer**: renders video+audio to file as fast as the encoder allows, with no real-time synchronisation. Back-pressure via bounded queue. AMCP: `ADD 1 OFFLINE /path.mp4 [ffmpeg args]`
- **Ultralight HTML producer**: synchronous WebKit-based renderer that advances exactly one frame per channel tick. Deterministic at any speed. Custom FileSystem for cross-platform `file://` asset loading.
- **CEF deterministic API**: injects `__caspar_frame`, `__caspar_time`, `__caspar_tick()` per receive() for template compatibility across both engines.
- **Stereo audio**: `-filter:a pan=stereo|c0=c0|c1=c1` downmixes CasparCG's 16-channel output at encode time.

## Blast radius

44 lines modified across 8 existing files. All new code is self-contained or gated behind `ENABLE_ULTRALIGHT`.

| File | Change |
|---|---|
| `ffmpeg/CMakeLists.txt` | +2 lines (add offline consumer files) |
| `ffmpeg/ffmpeg.cpp` | +4 lines (register offline consumer) |
| `modules/CMakeLists.txt` | +4 lines (add ultralight subdir, gated) |
| `CMakeLists.txt` | +2 lines (ENABLE_ULTRALIGHT option) |
| `Bootstrap_Linux.cmake` | +2 lines (find Ultralight SDK) |
| `cg_proxy.cpp` | 1 line changed (`insert` to `insert_or_assign`) |
| `html_producer.cpp` | +24 lines (inject deterministic JS API) |
| `casparcg.config` | +5 lines (document offline element) |

## Performance

| Configuration | Speed |
|---|---|
| Video only (offline) | ~17x real-time |
| Video + Ultralight (offline) | ~2x real-time |
| Video + CEF (offline) | ~17x real-time |
| Precise render (500 frames) | 500 = 500 exact match |

Tested in Docker (linux/amd64, software GL). Native Linux with GPU will be significantly faster.

## Test plan

- [x] Matrix test: 4 configurations (CEF/UL x FFmpeg/Offline) all pass
- [x] Precise render: 500 source frames -> 500 output frames, frame 0 synced
- [x] Audio: stereo AAC via `-filter:a` flag
- [x] Asset loading: PNG, TTF via custom Ultralight FileSystem
- [x] Thread safety: atomics + mutex for AMCP/tick thread boundary
- [x] Code review: 3 critical, 2 high, 5 medium issues found and fixed


olekristensen:
Closing — opened against wrong repo. Will reopen on fork.

niklaspandersson:
Interesting stuff!
I'd like to see two separate PR, one for Offline and a separate one for the Ultralight HTML. I realize Ultralight is required for offline rendering with graphics, but I can see the offline mode having use cases without graphics.

How does the Ultralight HTML compare to CEF when it comes to html/css/javascript standard support?

Also, how much does the Offline consumer differ from the the current ffmpeg consumer? Just glancing at the code gives me the feeling that there is a lot of overlap (duplicated code)

dimitry-ishenko:
Maybe just add an option to the FFmpeg consumer to set `has_synchronization_clock` to true and just consume frames as fast as possible? Or is there something else that keeps it from hogging the CPU at that point?

olekristensen:
Thanks, you are right about the duplicate code.

I made this as an experiment, that's why I kept it separate. But there was a suggestion to merge it into the existing ffmpeg consumer on the forum.

Makes sense to have a seperate ultralight PR. Ultralight is not fully feature complete, but afaik the best option for lockstep rendering. They've got a feature parity issue in their GitHub.

I'll give it a shot and keep it in draft until then.

Julusian:
The forum post: https://casparcgforum.org/t/experiment-offline-rendering-with-casparcg/7525

I gave a couple of suggestions there on the offline consumer part.

I too would like to see this as 2 PRs. The offline consumer part (after my suggestions) I think is very mergable, the ultralight part I think needs more thought on if we want to merge it as it will have a maintenance cost and at this stage I don't know if there is enough demand to justify us taking that on

PR #1734, opened 2026-04-03
Backport OAL fixes for 2.5.1

dimitry-ishenko:


dimitry-ishenko:
@Julusian I've back-ported C++20 and my OAL fixes for the v2.5.x branch. Not sure if there is a way to trigger CI to make sure all builds are good. I will test Linux builds on my box, but I have no way of building for Windows.

PR #1747, opened 2026-05-12
set INSTALL_RPATH for system cef (needed for arch / fedora)

scientress:
The rpath is stripped from the binary during `make install` unless it's set explicitly in `INSTALL_RPATH`.
This is needed on archlinux and probably also fedora.

### DIFF
--- a/src/CMakeModules/CEF_Linux.cmake
+++ b/src/CMakeModules/CEF_Linux.cmake
@@ -48,6 +48,8 @@ if(CEF_USE_FIND_PACKAGE)
         CEF::Library
         "-Wl,-rpath,${_cef_lib_dir}"
     )
+    # cmake install strips the RPATH from the built binary unless we explicity set it in INSTALL_RPATH
+    SET(CMAKE_INSTALL_RPATH _cef_lib_dir)
     # Fedora's CEF headers (145+) use C++20 features (concepts, <=>). Propagate
     # the requirement so any target linking CEF::CEF is compiled with C++20.
     target_compile_features(CEF::CEF INTERFACE cxx_std_20)


dimitry-ishenko:
> This is needed on archlinux and probably also fedora.

Why is this necessary? It works fine on other distros.

scientress:
On arch `libcef.so` is installed to [`/usr/lib/cef/libcef.so`](https://archlinux.org/packages/extra/x86_64/cef/), on fedora it's installed to [`/usr/lib64/cef/libcef.so'](https://packages.fedoraproject.org/pkgs/cef/cef/fedora-rawhide.html#files).
Without an `rpath` set the linked binary will not find the shared object.

Important steps to reproduce:
The binary in the build output folder will have the rpath set automatically, even without the rpath option in [L63](https://github.com/CasparCG/server/blob/71d8edcefdc8f8816fbacdd931035cdd18714179/src/CMakeModules/CEF_Linux.cmake#L63).

But not the binary that is installed with `cmake install`, since cmake strips rpaths.

dimitry-ishenko:
I see, said the blind man... 😉 

So, the reason it works with `CEF_SYSTEM_PACKAGE` is because the library is passed with the full path. IMHO a better solution would be to remove all `-Wl,-rpath` gimmicks and instead enable `CMAKE_INSTALL_RPATH_USE_LINK_PATH` and let CMake  automatically populate `RPATH`.

This is what I did it my Debian/Ubuntu builds: https://github.com/cccp-linux/packages/actions/runs/24059411349/job/70995923251#step:6:2525

scientress:
Oh yeah that is a good idea.

PR #1748, opened 2026-05-19
fix: av_input does not destroy graph_ object before joining worker thread

niklaspandersson:
Fixes a race condition in the destruction of ffmpeg::Input instances.

The graph: object was reset before the worker-thread had time to exit. Potentially leading to nullptr dereference

fix:
remove the line resetting graph_ from the desctructor, since graph_ will be destroyed automatically anyway.


Julusian:
I've also cherry-picked this to 2.5.x. I should probably do a 2.5.1 soon

dimitry-ishenko:
> I've also cherry-picked this to 2.5.x. I should probably do a 2.5.1 soon

🎉 

@Julusian can we please push OAL fixes in there as well?

Issue #1749, opened 2026-05-20
Bug: ArtNet consumer not rasterizing image correctly


dimitry-ishenko:
### Observed Behavior

With this config:
```xml
<?xml version="1.0" encoding="utf-8"?>

<configuration>
    <paths>
        <media-path>media/</media-path>
        <log-path disable="false">log/</log-path>
        <data-path>data/</data-path>
        <template-path>template/</template-path>
    </paths>
    <lock-clear-phrase>secret</lock-clear-phrase>
    <video-modes>
        <video-mode>
            <id>144x72p60</id>
            <width>144</width>
            <height>72</height>
            <time-scale>60000</time-scale>
            <duration>1000</duration>
            <cadence>800</cadence>
        </video-mode>
    </video-modes>
    <channels>
        <channel>
            <!-- <video-mode>144x72p60</video-mode> -->
            <video-mode>144x72p60</video-mode>
            <consumers>
                <screen />

                <artnet>
                    <universe>0</universe>
                    <host>10.42.0.21</host>
                    <port>6454</port>
                    <refresh-rate>30</refresh-rate>
                    <fixtures>
                        <!-- Row 0 -->
                        <fixture>
                            <type>RGBW</type>
                            <start-address>1</start-address>
                            <fixture-count>18</fixture-count>
                            <fixture-channels>4</fixture-channels>
                            <x>72</x>
                            <y>36</y>
                            <width>144</width>
                            <height>72</height>
                            <rotation>0</rotation>
                        </fixture>
                    </fixtures>
                </artnet>

            </consumers>
        </channel>
    </channels>
    <controllers>
        <tcp>
            <port>5250</port>
            <protocol>AMCP</protocol>
        </tcp>
    </controllers>
    <amcp>
        <media-server>
            <host>localhost</host>
            <port>8000</port>
        </media-server>
    </amcp>
</configuration>
```
If I play this image:

<img width="144" height="72" alt="Image" src="https://github.com/user-attachments/assets/e845ea18-6787-405a-9ef3-eaca4c4983b7" />

I see a row of red pixels followed by green pixels. And, the server generates the following ArtNet packets:

```
22:44:24.186051 eth0  In  IP laptop.39323 > FPP.6454: UDP, length 530
	0x0000:  4500 022e 854d 4000 4011 9e81 0a2a 0088  E....M@.@....*..
	0x0010:  0a2a 0015 999b 1936 021a 4fff 4172 742d  .*.....6..O.Art-
[...]

PR #1751, opened 2026-05-21
Pixel consumer

dimitry-ishenko:
Sample config for testing:

```xml
<configuration>
    <paths>
        <media-path>media/</media-path>
        <log-path disable="false">log/</log-path>
        <data-path>data/</data-path>
        <template-path>template/</template-path>
    </paths>

    <video-modes>
        <video-mode>
            <id>matrix</id>
            <width>18</width>
            <height>16</height>
            <time-scale>60000</time-scale>
            <duration>1000</duration>
            <cadence>800</cadence>
        </video-mode>
    </video-modes>

    <channels>
        <channel>
            <video-mode>matrix</video-mode>
            <consumers>
                <screen />

                <pixel>
                    <protocol>ArtNet</protocol>
                    <host>10.42.0.21</host>
                    <port>6454</port>

                    <universe>0</universe>
                    <start-address>1</start-address>
                    <type>RGBW</type>
                    <coef>
                        <r>1.0</r>
                        <g>0.55</g>
                        <b>0.154</b>
                        <w>0.21</w>
                    </coef>
                    <gamma>2.2</gamma>
                </pixel>

            </consumers>
        </channel>
    </channels>

    <controllers>
        <tcp>
            <port>5250</port>
            <protocol>AMCP</protocol>
        </tcp>
    </controllers>
</configuration>
```

### DIFF
--- a/src/common/endian.h
+++ b/src/common/endian.h
@@ -18,55 +18,84 @@
  *
  * Author: Helge Norberg, helge.norberg@svt.se
  * Author: Robert Nagy, ronag89@gmail.com
+ * Author: Dimitry Ishenko, dimitry (dot) ishenko (at) (gee) mail (dot) com
  */
 
 #pragma once
 
-#include <cstdint>
+#include <concepts>
+#include <cstddef> // std::size_t
 #include <type_traits>
+
+#if defined(_MSC_VER)
+#include <stdlib.h> // _byteswap_...
+#endif
+
+template <typename T, std::size_t N>
+concept integral_of_size = std::integral<T> && (sizeof(T) == N);
 
 namespace caspar {
 
-template <typename T>
-typename std::enable_if<sizeof(T) == sizeof(std::uint8_t), T>::type swap_byte_order(const T& value)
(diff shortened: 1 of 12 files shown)


dimitry-ishenko:
@Julusian this is a follow-up of the discussion in #1749.

I've figured it's probably better to keep this as a separate consumer. 

I've decoupled the ArtNet sink and made it easy to add other types of sinks in the future.

This adds a few interesting possibilities. First, we can add other output protocols, such as sACN (E1.31). Or, we can add a pipe as a sink and output raw pixel data for processing by external tools. This makes CasparCG act as a compositing engine. And so on.

dimitry-ishenko:
Here is an example of the sync-check video on an 18x16 RGBW pixel matrix:

https://github.com/user-attachments/assets/691337eb-d1b7-4e8a-89e3-743c8eae6ea2

The original video is here:

https://github.com/user-attachments/assets/b4fd1758-de88-457e-bcbd-1c5d3cd59aea

PR #1753, opened 2026-06-05
feat: add MIXER CLEARALL command

ePirat:
This allows to easily clear all mixer transforms
on all channels and layers.

### DIFF
--- a/src/protocol/amcp/AMCPCommandsImpl.cpp
+++ b/src/protocol/amcp/AMCPCommandsImpl.cpp
@@ -1402,6 +1402,15 @@ std::wstring mixer_clear_command(command_context& ctx)
     return L"202 MIXER OK\r\n";
 }
 
+std::wstring mixer_clear_all_command(command_context& ctx)
+{
+    for (auto& ch : *ctx.channels) {
+        ch.stage->clear_transforms();
+    }
+
+    return L"202 MIXER OK\r\n";
+}
+
 std::wstring channel_grid_command(command_context& ctx)
 {
     int   index = 1;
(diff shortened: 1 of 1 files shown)


dimitry-ishenko:
Without getting into the technical merits of this command, IMHO `MIXER CLEAR ALL` is be better.

ePirat:
> Without getting into the technical merits of this command, IMHO `MIXER CLEAR ALL` is be better.

I would agree, however it seems not that trivial to implement (or maybe someone more familiar with the code base can give me a hint) as MIXER CLEAR is already an existing channel-command.

dimitry-ishenko:
Maybe something like this?...

```c++
std::wstring mixer_clear_command(command_context& ctx)
{
    if ctx.parameters.size() > 1 && ctx.parameters[1] == L"all") {
        for (auto& ch : *ctx.channels)
            ch.stage->clear_transforms();
    } else {
        int layer = ctx.layer_id;
        if (layer == -1)
            ctx.channel.stage->clear_transforms();
        else
            ctx.channel.stage->clear_transforms(layer);
    }
    return L"202 MIXER OK\r\n";
}
```

ePirat:
@dimitry-ishenko That doesn't work because MIXER CLEAR is a channel command, so the channel id parsing (and failing thereof when it is not a valid integer but ALL) happens before it even reaches the `mixer_clear_command` handler function.

dimitry-ishenko:
> @dimitry-ishenko That doesn't work because MIXER CLEAR is a channel command, so the channel id parsing (and failing thereof when it is not a valid integer but ALL) happens before it even reaches the `mixer_clear_command` handler function.

You can also do what INFO does - register MIXER CLEAR as a global command. Then one can do:
```
MIXER 1-10 CLEAR - clear layer
MIXER 1 CLEAR    - clear channel
MIXER CLEAR      - clear everything
```

But honestly, I don't even know if this command is even worth it. You can get a list of channels using the INFO command I believe, and from there issue MIXER CLEAR for each channel.

PR #1754, opened 2026-06-09
MXL Producer & Consumer

mint-dewit:
Implementing https://github.com/dmf-mxl/mxl

This is in quite an early state but I wanted to open the PR for potential discussion

As of opening this PR this should be able to play back MXL video and audio flows and write MXL video and audio flow to disk
**Not** supported are flows with Alpha Channel and Interlaced flows. And also none of the fabrics stuff, but that's not officially released either.

One discussion point I'd like to bring up is how to link against the MXL SDK. I am currently statically linking (and probably need to fix the path) which means we have to build to MXL SDK for target platforms. The alternative is dynamically linking but that has the downside of having to include a few libraries that the SDK needs. I don't really consider myself qualified to have an opinion here.


dimitry-ishenko:
Nice work @mint-dewit. A couple of critiques if I may:

1. Would it be possible to not pull in yet another boost library? There are several header-only libraries that require no runtime dependencies. The most popular one is [nhlohmann/json](https://github.com/nlohmann/json).

2. IMHO `v210` should be its own component, as you are basically copying what @niklaspandersson did for decklink.

3. `boost::regex` has a replacement in the Standard Library. For example (untested):

```c++
std::wregex re{L"^mxl://([^/]+)?(/[^?]+)\\?id=([^&]+)(?:&id=([^&]+))?"};

std::wcmatch match; 
if (std::regex_search(name_or_url, match, re)) {
    std::wstring authority = match[1].str();
    std::wstring domain    = match[2].str();
    std::wstring id1       = match[3].str();
    std::wstring id2       = match[4].str();
}
```

4. You are including `boost/thread.hpp` but I didn't see it being used anywhere. If it is, there is also a replacement in the Standard Library.

Issue #1549, opened 2024-06-25
Bug: [Short description of the bug]


leftearsons:
### Observed Behavior

Starting from somewhere around July 2023, CasparCG (v2.3.x-2022-04-25 onwards, with current master versions included as well) will crash unexpectedly from time to time without writing any error log messages at all!   The occurrence frequency cannot be predicted and could span from one day up to one month, but we still have some servers that could operate for longer without crashing.   
The only trace that is visible are some information and error logs in the Windows event viewer.   
I have noticed that the issue is irrelevant of whether flash or html graphics are in use, as well as not necessarily related to the video source (would take place both with SDI and HLS video sources).    
Based on our statistics though, it seems to be more frequent when loading streaming video content (HLS specifically).   

I am adding the related logs from the event viewer for your reference:

```
ERROR LOG:

Faulting application name: casparcg.exe, version: 2.4.0.0, time stamp: 0x661ffb91
Faulting module name: ntdll.dll, version: 10.0.19041.2130, time stamp: 0xb5ced1c6
Exception code: 0xc0000374
Fault offset: 0x00000000000ff6a9
Faulting process id: 0x1a98
Faulting application start time: 0x01daadd1fa3f3c3b
Faulting application path: C:\Users\Admin\Documents\Video Service\graphics\casparcg.exe
Faulting module path: C:\WINDOWS\SYSTEM32\ntdll.dll
Report Id: 0dd7e7bd-29c1-4cc5-8982-397b89622534
Faulting package full name: 
Faulting package-relative application ID: 
```

```
INFORMATION LOG:

Fault bucket 1919053832705669865, type 4
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: casparcg.exe
P2: 2.4.0.0
P3: 661ffb91
P4: StackHash_30e5
P5: 10.0.19041.2130
P6: b5ced1c6
P7: c0000374
P8: PCH_7B_FROM_ntdll+0x000000000009DC14
P9: 
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER1311.tmp.dmp
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER16CB.tmp.WERInternalMetadata.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER16DB.tmp.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER16D9.tmp.csv
[...]

PR #1755, opened 2026-06-18
fix(ffmpeg/core): harden producer teardown against heap corruption under heavy looping/CLEAR load

C-alcaide:
**Disclamer: This work has been 'done' with AI, I have lightly tested to make sure it doesn't break anything obvious, but #1755 and #1756 are just drafts for now that require discussion to see if this could be the right approach or not. This is trying to solve a series of unexpected crashes (nothing shows in the logs)  related to ntdll/heap corruption/access violation, sometimes with no new commands or interactions at that point, the looping of 5 channels was the only active work. We have counted 26 hard crashes in the last 1,5 years (using 2.4.3 stable since last september). Since I haven't seen any work specifically addressing similar issues I used Opus 4.8 to offer the fixes**

## Problem
Under sustained looping playback combined with frequent CLEAR/REMOVE (and a DeckLink output), the server can intermittently crash with heap corruption in `ntdll`. The allocator is tbbmalloc with `tbbmalloc_proxy`, which overrides the global `malloc`/`free`; this makes cross-DLL frees during producer teardown a heap-corruption vector, and the single-threaded destroyer plus an unbounded backlog made the failure window worse under load.

This PR is **defensive hardening** of the teardown path. It does not change the decode pipeline behaviour. (A separate, stacked PR addresses the underlying per-loop churn that triggers it.)

## Changes
1. **Route producer teardown through a dedicated executor**
   `~ffmpeg_producer` no longer destroys the producer inline; it hands the `unique_ptr` to a teardown executor so frees happen on a controlled thread.

2. **Use a small bounded teardown pool instead of a single thread**
   A single destroyer thread serialised all teardowns and caused head-of-line blocking when one producer was slow to release. Replaced with a fixed 3-worker round-robin pool (`ffmpeg_producer_destroyer_0..2`).

3. **Bound the producer-destroyer backlog with a synchronous fallback**
   The previous `CASPAR_VERIFY(size() < 8)` could trip (or grow unbounded) under bursty teardown. Replaced with an explicit backlog limit: when the queue is at capacity, the producer is destroyed synchronously on the calling thread instead of being enqueued, bounding memory and avoiding the hard assert.

## Risk
Low. No change to decoding/looping output; only the lifetime/threading of producer destruction. Validated with a Release build (`casparcg.exe` links clean) and a smoke test of the assembled runtime.

### DIFF
--- a/src/core/producer/frame_producer_registry.cpp
+++ b/src/core/producer/frame_producer_registry.cpp
@@ -86,7 +86,12 @@ class destroy_producer_proxy : public frame_producer
         if (!destroyer)
             return;
 
-        CASPAR_VERIFY(destroyer->size() < 8);
+        // Throwing from a destructor (as CASPAR_VERIFY would) while many layers are being
+        // cleared at once is unsafe and can itself abort the process. Apply backpressure via
+        // logging instead of asserting.
+        if (destroyer->size() >= 8) {
+            CASPAR_LOG(warning) << L"Producer destroyer backlog: " << destroyer->size();
+        }
 
         auto producer = new spl::shared_ptr<frame_producer>(std::move(producer_));
 
--- a/src/core/producer/frame_producer_registry.cpp
+++ b/src/core/producer/frame_producer_registry.cpp
@@ -86,11 +86,26 @@ class destroy_producer_proxy : public frame_producer
         if (!destroyer)
             return;
 
-        // Throwing from a destructor (as CASPAR_VERIFY would) while many layers are being
-        // cleared at once is unsafe and can itself abort the process. Apply backpressure via
-        // logging instead of asserting.
-        if (destroyer->size() >= 8) {
-            CASPAR_LOG(warning) << L"Producer destroyer backlog: " << destroyer->size();
+        // Backpressure for the asynchronous destroyer queue.
+        //
+        // The queue used to be guarded by CASPAR_VERIFY(size < 8), but throwing from a
+        // destructor while many layers are being cleared at once is unsafe and can itself
+        // abort the process. Simply dropping the guard would let the backlog (and the memory
+        // of every not-yet-destroyed producer) grow without bound if destruction cannot keep
+        // up. Instead, once the backlog passes a threshold, destroy this producer
+        // synchronously on the current thread. That bounds the queue and provides real
+        // backpressure without ever throwing.
+        constexpr std::size_t destroyer_backlog_limit = 8;
+
+        if (destroyer->size() >= destroyer_backlog_limit) {
+            CASPAR_LOG(warning) << L"Producer destroyer backlog (" << destroyer->size()
+                                << L") reached limit; destroying synchronously to apply backpressure.";
+            try {
+                producer_.reset();
+            } catch (...) {
+                CASPAR_LOG_CURRENT_EXCEPTION();
+            }
+            return;
         }
 
         auto producer = new spl::shared_ptr<frame_producer>(std::move(producer_));
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -181,7 +181,10 @@ class Decoder
 
         thread = boost::thread([this]() {
             try {
-                while (!thread.interruption_requested()) {
+                // NOTE: Check interruption via the current-thread helper rather than reading the
+                // `thread` member, which is still being assigned by the constructor when this
+                // lambda starts running (data race on the partially-constructed member).
+                while (!boost::this_thread::interruption_requested()) {
                     auto av_frame = alloc_frame();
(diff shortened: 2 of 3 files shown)


dimitry-ishenko:
Sorry, I didn't read too deep into your PR, but are you saying we need a thread pool to destroy ffmpeg producers?

C-alcaide:
> Sorry, I didn't read too deep into your PR, but are you saying we need a thread pool to destroy ffmpeg producers?

Need is a strong word haha.  Here's what happened: in the most recent instance right before the crash the last log lines showed 3 or 4 ffmpeg producers being destroyed at the same time, as well as some screen consumers. WIth that log as a reference Claude might have chased the wrong assumption that the 'mass teardown' specifically caused the crash. 
It's hard to actually affirm that, but I didn't want to leave it out so we can discuss it and see if it aligns with what other people have seen. I think most times the crashes occur while just looping, so the other PR might be the actual fix we need.

In any case the first suggested change was a single serialized thread, but we might have encountered some blocking behaviour on edge cases so the thread pool was intended to mitigate that (not very likely) chance.

dimitry-ishenko:
I took a closer look at the code and here are my two dinars:

1. In your PR you are changing order of destruction in `AVProducer::~Impl()`. Have you checked if that alone fixes the problem? If it doesn't, shouldn't it be removed from the PR?

2. Currently, `~ffmpeg_producer` does destruction via a detached thread. But, from what I understand, every producer gets wrapped in an instance of `destroy_producer_proxy` upon creation, which also does destruction on another thread. So, the detached thread in `~ffmpeg_producer` is really for the birds...

   And in your PR you are adding yet another destroyer on top?... Sorry, but I am getting a bit lost.

3. Since `ffmpeg_producer` spins off a throw-away thread to destroy its resources, and then itself gets destroyed quicker by the proxy, is it possible that's causing access violation and a crash? I don't know if there is any interaction between them, so I am just shooting in the dark here.
 
   But in either case, that ffmpeg producer destructor should be removed, and the implicit destructor will destroy the resources inside the proxy on another thread. I would try that instead and monitor the queue size in the proxy to see if it ever reaches above 8.

C-alcaide:
Thanks a lot for your input, I don't have the knowledge to verify the code itself, so if what AI suggested is just plain wrong we're relying on the actual programmers to catch it...

> Have you checked if that alone fixes the problem?

Actually confirming if the fix works may take a long while, I'm trying but this bug is not easily reproducible at all. As the report #1549 said, the frequency changes a lot.

Tomorrow I will try to narrow down the possible causes, see if the older logs throw more light into this.

C-alcaide:
Some good news: I was able to get the windows dumps and asked Claude to analize them more in-depth.
Here's the report that it created:

https://github.com/C-alcaide/server/blob/CasparVPV/docs/REPORT_CasparCG_dumps.md

_The main take-away is that having 500 threads growing over time into 600 is probably not intended behaviour and it causes the crash.
So it seems that after the fix we can check how many threads are being created while looping for some hours and we can hopefully confirm that it worked_
**Edit: this hasn't been verified, since there's no real sign of  a leak in the threads**

dimitry-ishenko:
I wonder if that ffmpeg producer destructor is causing that, as it just spins up and detaches a thread for each instance:

https://github.com/CasparCG/server/blob/d603ee91fa568ca38199048d38005b886eba75ba/src/modules/ffmpeg/producer/ffmpeg_producer.cpp#L95-L104

Could you just delete that function and see what happens? (The default-generated dtor will delete the `producer_` variable synchronously.)

If my understanding of how producers get wrapped in the `destroy_producer_proxy` instance, then all destruction will still happen on another thread but sequentially. And if the destruction queue gets larger than 8, it should start triggering warnings in the log here:

https://github.com/CasparCG/server/blob/d603ee91fa568ca38199048d38005b886eba75ba/src/core/producer/frame_producer_registry.cpp#L89

C-alcaide:
> Could you just delete that function and see what happens?

I will try to test this next in a few days.

Today I tried to create synthetic tests (with local files) in case I could see signs of a leak in the threads (as the report mentions) but so far there's no signs that threads nor handles are leaking or not getting freed properly, which means that this is just a rare race event, maybe more related to the teardown path than the looping (it's still just speculation).

That also means that the changes proposed in the PRs wouldn't actually fix the issue, just reduce the frequency of the 'buggy' path being run.

Next week I will try to find common denominators in the days the crashes happened, but so far it's looking like a needle in a haystack.

C-alcaide:
Turns out there was a leak ( #1757 ) and handles were duplicated without actual threads, but it mostly applies to receiving streams with bad conectivity (constant reconnections, bad frames, etc.) so I'm not sure it helps my case with local files. I'll have to keep digging. That investigation took 2 full days of synthetic tests and discarding theories that AI was suggesting, hopefully it solves  #1549  at least.

PR #1756, opened 2026-06-18
fix(ffmpeg): reuse decoders across loop wrap instead of recreating worker threads + codec contexts

C-alcaide:
> **Stacked on #1755.** This branch currently contains #1755's three commits plus the single commit reviewed here. Please review/merge after #1755; once it merges I'll rebase and this PR will reduce to just the decoder-reuse commit.

**Disclamer: This work has been 'done' with AI, I have lightly tested to make sure it doesn't break anything obvious, but #1755 and #1756 are just drafts for now that require discussion to see if this could be the right approach or not. This is trying to solve a series of unexpected crashes (nothing shows in the logs)  related to ntdll/heap corruption/access violation, sometimes with no new commands or interactions at that point, the looping of 5 channels was the only active work. We have counted 26 hard crashes in the last 1,5 years (using 2.4.3 stable since last september). Since I haven't seen any work specifically addressing similar issues I used Opus 4.8 to offer the fixes**

`I decided to separate the work into two different PRs since this one is much more invasive than the other commits, and changes a more fundamental part of the core.`

## Problem
`AVProducer::Impl::seek_internal()` runs on every loop wrap (and every seek) and calls `decoders_.clear()`, which destroys **every** `Decoder` — joining its worker `boost::thread` and freeing its `AVCodecContext` — and then `reset()` recreates them all (new thread + `avcodec_alloc_context3`/`avcodec_open2`).

For a looping clip this means a continuous create/join of decoder threads and open/close of codec contexts every loop iteration. With `tbbmalloc_proxy` overriding global `malloc`/`free`, that per-loop allocation churn is the root cause behind the idle-loop heap corruption (the teardown hardening in #1755 narrows the window but does not remove the churn).

## Change
Reuse decoders across a loop wrap by flushing them in place instead of destroying and recreating them:

- Add a thread-safe `Decoder::flush()`. The producer thread raises a flush request and blocks on an ack; the **owning worker thread** performs the actual flush at a safe sync point (`do_flush()`): clears the input/output queues, calls `avcodec_flush_buffers(ctx)`, and resets `next_pts`/`eof`. The codec context is therefore never touched from the producer thread — closing the data race that makes naive in-place flushing unsafe.
- The worker loop now checks the flush request at the top of the loop and in every blocking `condition_variable` predicate (one input wait, two output waits), discarding any in-flight frame so the flush is serviced promptly.
- `seek_internal()` now flushes the existing decoders instead of clearing the map. `reset()` is unchanged: it still rebuilds the (cheap) filter graphs and re-binds them to the surviving decoders.

A flushed decoder is behaviourally identical to a freshly constructed one (empty queues, flushed codec buffers, `next_pts = AV_NOPTS_VALUE`, `eof = false`), so output is unchanged — only the per-loop heap churn is removed.

## Risk
Moderate (concurrency change), mitigated by keeping all `ctx` access on the worker thread and using a request/ack handshake. Scope is intentionally limited to **decoders**; filter graphs are still rebuilt each loop.

## Testing
- Built `Release`; `av_producer.cpp` / `ffmpeg.lib` / `casparcg.exe` compile and link clean.
- Looping playback: clean loop seam, A/V sync stable.
- Primary success signal: **flat RSS over a long loop** (no per-loop alloc/free).
- Spot-checked user `CALL ... SEEK` accuracy and play-once EOF.

### DIFF
--- a/src/core/producer/frame_producer_registry.cpp
+++ b/src/core/producer/frame_producer_registry.cpp
@@ -86,7 +86,12 @@ class destroy_producer_proxy : public frame_producer
         if (!destroyer)
             return;
 
-        CASPAR_VERIFY(destroyer->size() < 8);
+        // Throwing from a destructor (as CASPAR_VERIFY would) while many layers are being
+        // cleared at once is unsafe and can itself abort the process. Apply backpressure via
+        // logging instead of asserting.
+        if (destroyer->size() >= 8) {
+            CASPAR_LOG(warning) << L"Producer destroyer backlog: " << destroyer->size();
+        }
 
         auto producer = new spl::shared_ptr<frame_producer>(std::move(producer_));
 
--- a/src/core/producer/frame_producer_registry.cpp
+++ b/src/core/producer/frame_producer_registry.cpp
@@ -86,11 +86,26 @@ class destroy_producer_proxy : public frame_producer
         if (!destroyer)
             return;
 
-        // Throwing from a destructor (as CASPAR_VERIFY would) while many layers are being
-        // cleared at once is unsafe and can itself abort the process. Apply backpressure via
-        // logging instead of asserting.
-        if (destroyer->size() >= 8) {
-            CASPAR_LOG(warning) << L"Producer destroyer backlog: " << destroyer->size();
+        // Backpressure for the asynchronous destroyer queue.
+        //
+        // The queue used to be guarded by CASPAR_VERIFY(size < 8), but throwing from a
+        // destructor while many layers are being cleared at once is unsafe and can itself
+        // abort the process. Simply dropping the guard would let the backlog (and the memory
+        // of every not-yet-destroyed producer) grow without bound if destruction cannot keep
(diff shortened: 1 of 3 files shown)


dimitry-ishenko:
> `AVProducer::Impl::seek_internal()` runs on every loop wrap (and every seek) and calls `decoders_.clear()`, which destroys **every** `Decoder` — joining its worker `boost::thread` and freeing its `AVCodecContext` — and then `reset()` recreates them all (new thread + `avcodec_alloc_context3`/`avcodec_open2`).

🤯 🤯 🤯

C-alcaide:
I found out that this 'reuse decoders' strategy was tried and discarded in 2018 since it caused broken loops and innacurate seeking 0efbcf341 

But a lot of the inner architecture and the eof logic has changed greatly since then, so if we don't see any of those issues with this PR maybe we can consider going back to this strategy.

Maybe @ronag can shed some light into this.

PR #1761, opened 2026-06-28
fix: escape backslashes and control chars in html cg update payload

firecraftgaming:
update() builds a JS update("\<data\>") call but only escaped ", not \. So any backslash in the payload (like the \" in JSON) broke the JS string and threw SyntaxError: missing ) after argument list.

### DIFF
--- a/src/modules/html/producer/html_cg_proxy.cpp
+++ b/src/modules/html/producer/html_cg_proxy.cpp
@@ -60,8 +60,12 @@ void html_cg_proxy::next(int layer) { producer_->call({L"next()"}); }
 
 void html_cg_proxy::update(int layer, const std::wstring& data)
 {
-    auto escaped =
-        boost::algorithm::replace_all_copy(boost::algorithm::trim_copy_if(data, boost::is_any_of(" \"")), "\"", "\\\"");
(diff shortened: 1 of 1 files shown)


dimitry-ishenko:
Maybe some day we will be able to ditch boost, speed up compilation by 10X and won't need to use pre-compiled headers.

If you feel the same way, you can use this instead:

```c++
    std::wstring_view view = data;
    auto start = view.find_first_not_of(L" \"");
    auto end = view.find_last_not_of(L" \"");

    std::wstring escaped;
    if (start != view.npos)
    {
        view = view.substr(start, end - start + 1);
        escaped.reserve(view.size() * 1.1);

        for (auto c : view)
            switch (c) {
                case L'\\': escaped += L"\\\\"; break;
                case L'"' : escaped += L"\\\""; break;
                case L'\n': escaped += L"\\n";  break;
                case L'\r': escaped += L"\\r";  break;
                case L'\t': escaped += L"\\t";  break;
                default:    escaped += c;
            }
    }
```

Or, better yet, maybe add a couple of 2/3 liners in `common/string.h` so that other parts of the code can use them as well. 

Just my 2 dinars' worth.

Julusian:
@dimitry-ishenko feel free to do a PR for that, but im going to merge this as is for now just so that this is fixed even if not optimally

PR #1765, opened 2026-07-07
Add primary color grading: WHITEBALANCE, LIFT/MIDTONE/GAIN, HUESHIFT, TONEBALANCE, SPLITTONE, CDL

C-alcaide:
## Summary

Adds the standard primary color-grading toolset to the OpenGL mixer as per-layer, fully tweenable AMCP `MIXER` commands:

- **`MIXER WHITEBALANCE`** — temperature / tint.
- **`MIXER LIFT` / `MIXER MIDTONE` / `MIXER GAIN`** — DaVinci-style 3-way (shadows / midtones / highlights) per-channel corrector.
- **`MIXER HUESHIFT`** — global hue rotation.
- **`MIXER TONEBALANCE`** — luminance-masked shadow / highlight balance.
- **`MIXER SPLITTONE`** — independent shadow / highlight tint with a crossover.
- **`MIXER CDL`** — ASC CDL (Slope / Offset / Power + saturation), the industry-standard primary grade exchange format.

Every effect is off by default and costs nothing when neutral (the kernel only sets a single `*_enable = false` uniform). All operate in the working color space before the legacy levels/CSB stage and preserve HDR headroom (no clamping).

## Commits

Split into reviewable commits, one tool per commit:

1. `mixer: add MIXER WHITEBALANCE command`
2. `mixer: add MIXER LIFT/MIDTONE/GAIN 3-way color corrector`
3. `mixer: add MIXER HUESHIFT command`
4. `mixer: add MIXER TONEBALANCE command`
5. `mixer: add MIXER SPLITTONE command`
6. `mixer: add MIXER CDL (ASC CDL) command`

## AMCP syntax

```
MIXER <ch-layer> WHITEBALANCE <temperature> <tint> [duration] [tween]
MIXER <ch-layer> LIFT    <r> <g> <b> [duration] [tween]
MIXER <ch-layer> MIDTONE <r> <g> <b> [duration] [tween]
MIXER <ch-layer> GAIN    <r> <g> <b> [duration] [tween]
MIXER <ch-layer> HUESHIFT <degrees> [duration] [tween]
MIXER <ch-layer> TONEBALANCE <shadows> <highlights> [duration] [tween]
MIXER <ch-layer> SPLITTONE <sr> <sg> <sb> <hr> <hg> <hb> [balance] [duration] [tween]
MIXER <ch-layer> SPLITTONE RESET
MIXER <ch-layer> CDL <sR sG sB> <oR oG oB> <pR pG pB> [saturation] [duration] [tween]
MIXER <ch-layer> CDL RESET
```

Sending any command with no parameters queries the current value. Neutral defaults: `WHITEBALANCE 0 0`, `LIFT 0 0 0`, `MIDTONE 1 1 1`, `GAIN 1 1 1`, `HUESHIFT 0`, `TONEBALANCE 0 0`, `SPLITTONE 0 0 0 0 0 0 0.5`.

### Ranges

- `temperature` / `tint`: −1..+1 (cool..warm / magenta..green)
- `hue` degrees: −180..+180
- `shadows` / `highlights`: −1..+1
- split `balance`: 0..1 crossover (default 0.5)

### Examples

```
MIXER 1-10 WHITEBALANCE 0.3 -0.1        # warm up, slight magenta
MIXER 1-10 LIFT 0.02 0.0 -0.02          # cool shadows
MIXER 1-10 GAIN 1.1 1.0 0.95            # warm highlights
MIXER 1-10 HUESHIFT 15                  # rotate hue +15°
MIXER 1-10 TONEBALANCE -0.2 0.15        # crush shadows, lift highlights
MIXER 1-10 SPLITTONE 0.0 0.05 0.1 0.1 0.05 0.0 0.5   # teal shadows, orange highlights
MIXER 1-10 WHITEBALANCE 0.4 0 25 easeinoutsine       # tween over 25 frames
MIXER 1-10 CDL 1.1 1.0 0.9  0.0 0.0 0.02  1.0 1.0 1.0  0.9 (warm slope, slight blue offset, desaturate)
```

## Implementation notes

- Each parameter is added to `image_transform`, wired through `image_transform::tween`, `operator==`, and the transform-combine step, so grades animate and stack like existing mixer values.
- Shader work stays in the existing OGL image pipeline (`shader.frag` grade functions + `image_kernel.cpp` uniform setup). `rgb2hsv`/`hsv2rgb` already existed upstream and are reused for the hue shift.
- Per-channel wheels/split-tone upload RGB swapped to the shader's internal BGR working order.
- Independent of #1758 (ACES color management) and #1762 (image effects) — builds on clean `master`.

## Testing

Builds clean on Windows (VS2026, Release). Wiki docs to follow separately.

### DIFF
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -302,10 +302,9 @@ struct image_kernel::impl
                 std::abs(gain[1] - 1.0) > epsilon || std::abs(gain[2] - 1.0) > epsilon;
             if (lmg_active) {
                 shader_->set("lmg_enable", true);
-                // Upload swapped to BGR order so .r affects displayed Blue, .b affects displayed Red.
-                shader_->set("lmg_lift", lift[2], lift[1], lift[0]);
-                shader_->set("lmg_midtone", midtone[2], midtone[1], midtone[0]);
-                shader_->set("lmg_gain", gain[2], gain[1], gain[0]);
+                shader_->set("lmg_lift", lift[0], lift[1], lift[2]);
+                shader_->set("lmg_midtone", midtone[0], midtone[1], midtone[2]);
+                shader_->set("lmg_gain", gain[0], gain[1], gain[2]);
             } else {
                 shader_->set("lmg_enable", false);
             }
@@ -338,9 +337,8 @@ struct image_kernel::impl
                 std::abs(hc[0]) > epsilon || std::abs(hc[1]) > epsilon || std::abs(hc[2]) > epsilon;
             if (split_active) {
                 shader_->set("split_tone_enable", true);
-                // Upload swapped to BGR order so .r affects displayed Blue, .b affects displayed Red.
-                shader_->set("split_shadow_color", sc[2], sc[1], sc[0]);
-                shader_->set("split_highlight_color", hc[2], hc[1], hc[0]);
+                shader_->set("split_shadow_color", sc[0], sc[1], sc[2]);
+                shader_->set("split_highlight_color", hc[0], hc[1], hc[2]);
                 shader_->set("split_balance", static_cast<float>(transforms.image_transform.split_balance));
             } else {
                 shader_->set("split_tone_enable", false);
@@ -360,10 +358,9 @@ struct image_kernel::impl
                 std::abs(cs - 1.0) > epsilon;
             if (cdl_active) {
                 shader_->set("cdl_enable", true);
-                // Upload swapped to BGR working order (.r=Blue, .b=Red).
-                shader_->set("cdl_slope", s[2], s[1], s[0]);
-                shader_->set("cdl_offset", o[2], o[1], o[0]);
-                shader_->set("cdl_power", p[2], p[1], p[0]);
+                shader_->set("cdl_slope", s[0], s[1], s[2]);
+                shader_->set("cdl_offset", o[0], o[1], o[2]);
+                shader_->set("cdl_power", p[0], p[1], p[2]);
                 shader_->set("cdl_saturation", static_cast<float>(cs));
             } else {
                 shader_->set("cdl_enable", false);
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -290,6 +290,27 @@ struct image_kernel::impl
             shader_->set("white_balance", false);
         }
 
+        // Lift / Midtone / Gain (per-channel 3-way colour corrector)
+        {
+            const auto& lift    = transforms.image_transform.lift;
+            const auto& midtone = transforms.image_transform.midtone;
+            const auto& gain    = transforms.image_transform.gain;
+            bool        lmg_active =
+                std::abs(lift[0]) > epsilon || std::abs(lift[1]) > epsilon || std::abs(lift[2]) > epsilon ||
+                std::abs(midtone[0] - 1.0) > epsilon || std::abs(midtone[1] - 1.0) > epsilon ||
+                std::abs(midtone[2] - 1.0) > epsilon || std::abs(gain[0] - 1.0) > epsilon ||
+                std::abs(gain[1] - 1.0) > epsilon || std::abs(gain[2] - 1.0) > epsilon;
+            if (lmg_active) {
+                shader_->set("lmg_enable", true);
+                // Upload swapped to BGR order so .r affects displayed Blue, .b affects displayed Red.
+                shader_->set("lmg_lift", lift[2], lift[1], lift[0]);
+                shader_->set("lmg_midtone", midtone[2], midtone[1], midtone[0]);
+                shader_->set("lmg_gain", gain[2], gain[1], gain[0]);
+            } else {
+                shader_->set("lmg_enable", false);
+            }
+        }
+
         // Setup drawing area
 
         GL(glViewport(0, 0, params.background->width(), params.background->height()));
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -347,6 +347,29 @@ struct image_kernel::impl
             }
         }
 
+        // ASC CDL (Slope/Offset/Power)
+        {
+            const auto& s  = transforms.image_transform.cdl_slope;
+            const auto& o  = transforms.image_transform.cdl_offset;
+            const auto& p  = transforms.image_transform.cdl_power;
+            double      cs = transforms.image_transform.cdl_saturation;
+            bool        cdl_active =
+                std::abs(s[0] - 1.0) > epsilon || std::abs(s[1] - 1.0) > epsilon || std::abs(s[2] - 1.0) > epsilon ||
+                std::abs(o[0]) > epsilon || std::abs(o[1]) > epsilon || std::abs(o[2]) > epsilon ||
+                std::abs(p[0] - 1.0) > epsilon || std::abs(p[1] - 1.0) > epsilon || std::abs(p[2] - 1.0) > epsilon ||
+                std::abs(cs - 1.0) > epsilon;
+            if (cdl_active) {
+                shader_->set("cdl_enable", true);
+                // Upload swapped to BGR working order (.r=Blue, .b=Red).
+                shader_->set("cdl_slope", s[2], s[1], s[0]);
+                shader_->set("cdl_offset", o[2], o[1], o[0]);
+                shader_->set("cdl_power", p[2], p[1], p[0]);
+                shader_->set("cdl_saturation", static_cast<float>(cs));
+            } else {
+                shader_->set("cdl_enable", false);
+            }
+        }
+
         // Setup drawing area
 
         GL(glViewport(0, 0, params.background->width(), params.background->height()));
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -319,6 +319,16 @@ struct image_kernel::impl
             shader_->set("hue_shift_enable", false);
         }
 
+        // Tonal balance (shadows / highlights)
+        if (std::abs(transforms.image_transform.shadows) > epsilon ||
+            std::abs(transforms.image_transform.highlights) > epsilon) {
+            shader_->set("tonebalance_enable", true);
+            shader_->set("tb_shadows", static_cast<float>(transforms.image_transform.shadows));
+            shader_->set("tb_highlights", static_cast<float>(transforms.image_transform.highlights));
+        } else {
+            shader_->set("tonebalance_enable", false);
+        }
+
         // Setup drawing area
 
         GL(glViewport(0, 0, params.background->width(), params.background->height()));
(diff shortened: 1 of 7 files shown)


Julusian:
@niklaspandersson regarding the vulkan mixer, are you happy to port any ogl mixer changes to the vulkan one, or should that be a soft requirement for any prs?

C-alcaide:
> @niklaspandersson regarding the vulkan mixer, are you happy to port any ogl mixer changes to the vulkan one, or should that be a soft requirement for any prs?

Sorry to jump in.
I have the equivalent vulkan mixer shaders 'ready' (need more thorough testing probably) in another branch, We also discussed this same topic and decided to start with the ogl and port everything once the vk mixer is mature enough.

src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -290,6 +290,27 @@ struct image_kernel::impl
...
+            const auto& midtone = transforms.image_transform.midtone;
+            const auto& gain    = transforms.image_transform.gain;
+            bool        lmg_active =
+                std::abs(lift[0]) > epsilon || std::abs(lift[1]) > epsilon || std::abs(lift[2]) > epsilon ||
+                std::abs(midtone[0] - 1.0) > epsilon || std::abs(midtone[1] - 1.0) > epsilon ||
+                std::abs(midtone[2] - 1.0) > epsilon || std::abs(gain[0] - 1.0) > epsilon ||
+                std::abs(gain[1] - 1.0) > epsilon || std::abs(gain[2] - 1.0) > epsilon;
+            if (lmg_active) {
+                shader_->set("lmg_enable", true);
+                // Upload swapped to BGR order so .r affects displayed Blue, .b affects displayed Red.
```
niklaspandersson: Why do you need to swap the order?

src/accelerator/ogl/image/shader.frag
```diff
@@ -539,6 +545,16 @@ vec3 apply_white_balance(vec3 c, float temp, float tint_val)
     return c;
 }
 
+// ---- Lift / Midtone / Gain (3-way colour corrector) ----
+// out = pow(max(c * gain + lift, 0.0), 1.0 / midtone).
+// No upper clamp so HDR values pass through; negatives clamped for pow() safety.
+vec3 apply_lmg(vec3 c, vec3 lift, vec3 midtone, vec3 gain)
+{
+    c = max(c * gain + lift, vec3(0.0));
+    c = pow(c, max(vec3(0.01), 1.0 / midtone));
```
niklaspandersson: why the max(0.01 ...) in the exponential? Isn't it the midtone-value that should be protected from zero?

src/accelerator/ogl/util/transforms.cpp
```diff
@@ -81,6 +81,9 @@ void apply_transform_colour_values(core::image_transform& self, const core::imag
         self.midtone[i] *= other.midtone[i];
         self.gain[i]    *= other.gain[i];
     }
+
+    // Hue shift: additive
+    self.hue_shift += other.hue_shift;
```
niklaspandersson: Should this maybe be enforced to be in the [-180, 180] range?

src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -329,6 +329,24 @@ struct image_kernel::impl
...
+        // Split toning (independent shadow / highlight tint)
+        {
+            const auto& sc = transforms.image_transform.split_shadow_color;
+            const auto& hc = transforms.image_transform.split_highlight_color;
+            bool        split_active =
+                std::abs(sc[0]) > epsilon || std::abs(sc[1]) > epsilon || std::abs(sc[2]) > epsilon ||
+                std::abs(hc[0]) > epsilon || std::abs(hc[1]) > epsilon || std::abs(hc[2]) > epsilon;
+            if (split_active) {
+                shader_->set("split_tone_enable", true);
+                // Upload swapped to BGR order so .r affects displayed Blue, .b affects displayed Red.
```
niklaspandersson: why swapped?

src/accelerator/ogl/util/transforms.cpp
```diff
@@ -88,6 +88,14 @@ void apply_transform_colour_values(core::image_transform& self, const core::imag
     // Tone balance: additive
     self.shadows    += other.shadows;
     self.highlights += other.highlights;
+
+    // Split toning: additive colours, balance override if non-default
+    for (int i = 0; i < 3; ++i) {
+        self.split_shadow_color[i]    += other.split_shadow_color[i];
+        self.split_highlight_color[i] += other.split_highlight_color[i];
+    }
+    if (other.split_balance != 0.5)
```
niklaspandersson: I don't know how this particular color correction work, but why is split_balance just overridden when all the other float parameters are either multiplicative or addative?

src/accelerator/ogl/image/image_kernel.cpp
```diff
@@ -347,6 +347,29 @@ struct image_kernel::impl
...
+            const auto& p  = transforms.image_transform.cdl_power;
+            double      cs = transforms.image_transform.cdl_saturation;
+            bool        cdl_active =
+                std::abs(s[0] - 1.0) > epsilon || std::abs(s[1] - 1.0) > epsilon || std::abs(s[2] - 1.0) > epsilon ||
+                std::abs(o[0]) > epsilon || std::abs(o[1]) > epsilon || std::abs(o[2]) > epsilon ||
+                std::abs(p[0] - 1.0) > epsilon || std::abs(p[1] - 1.0) > epsilon || std::abs(p[2] - 1.0) > epsilon ||
+                std::abs(cs - 1.0) > epsilon;
+            if (cdl_active) {
+                shader_->set("cdl_enable", true);
+                // Upload swapped to BGR working order (.r=Blue, .b=Red).
```
niklaspandersson: why swapped?

niklaspandersson:
> > @niklaspandersson regarding the vulkan mixer, are you happy to port any ogl mixer changes to the vulkan one, or should that be a soft requirement for any prs?
> 
> Sorry to jump in. I have the equivalent vulkan mixer shaders 'ready' (need more thorough testing probably) in another branch, We also discussed this same topic and decided to start with the ogl and port everything once the vk mixer is mature enough.

Yea. Like @C-alcaide said, in this case I think it is covered. In the general case, I'm fine with porting these kinds of changes to the vulkan accelerator. At least as long as it is in the "experimental" phase.

niklaspandersson:
Looks good.

I have one general question / concern regarding valid values of all of these parameters. I think we should validate that provided values in the MIXER commands are in the legal range for each command. Also, when combining transforms, especially additively, I think we need to validate that the result is in the correct range

src/accelerator/ogl/image/image_kernel.cpp
C-alcaide: OpenGL fragment shader works in BGR, but maybe it's better to do the swap in the main function instead, I think that's how it's done in other places as well, so it should be more consistent,

src/accelerator/ogl/image/shader.frag
C-alcaide: yeah good catch

src/accelerator/ogl/util/transforms.cpp
C-alcaide: Yeah I wasn't sure either, Claude said:

> "The honest answer: split_balance is a crossover position in luma, not a strength. Adding two 0.5s gives 1.0 (everything is shadow); multiplying gives 0.25. Neither means anything. And the shader has exactly one split_balance uniform and one colour pair, so two stacked split-tone stages cannot be represented at all — any composition rule is lossy. Adding the colours and keeping a single crossover is the least-wrong approximation. Upstream's precedent for shape parameters like this is equally arbitrary: chroma key's target_hue/hue_width/softness all use std::max 
> 
> (transforms.cpp:60-64).
> 
> But the implementation is genuinely fragile and I'd change it:
> 
> if (other.split_balance != 0.5)
> Float equality against a sentinel. After a tween, split_balance can be 0.49999999, which reads as "explicitly set". Better rule: take other's balance only when other actually has split toning active (a non-zero shadow or highlight colour) — that's the condition the shader's enable check already uses — and document that stacking two split tones collapses to one crossover."

I guess this is a bit more coherent behaviour.

C-alcaide:
> Looks good.
> 
> I have one general question / concern regarding valid values of all of these parameters. I think we should validate that provided values in the MIXER commands are in the legal range for each command. Also, when combining transforms, especially additively, I think we need to validate that the result is in the correct range

Do you mean validate as in putting 'guards' internally in the server? Legal range will have to take into account colorspace and sdr/hdr then. 
Or do you mean just measuring the results with an external harness to make sure the math is solid?

niklaspandersson:
> > Looks good.
> 
> > 
> 
> > I have one general question / concern regarding valid values of all of these parameters. I think we should validate that provided values in the MIXER commands are in the legal range for each command. Also, when combining transforms, especially additively, I think we need to validate that the result is in the correct range
> 
> 
> 
> Do you mean validate as in putting 'guards' internally in the server? Legal range will have to take into account colorspace and sdr/hdr then. 
> 
> Or do you mean just measuring the results with an external harness to make sure the math is solid?

I just mean checking against the range in which they are defined in the amcp command executor, and after combining (checking that the value is -1 <= x <= 1 for instance, or -180 to 180)

src/accelerator/ogl/util/shader.cpp
```diff
@@ -132,7 +132,7 @@ struct shader::impl
         GL(glUniform3f(get_uniform_location(name.c_str()),
                        static_cast<float>(value0),
                        static_cast<float>(value1),
-                       static_cast<float>(value1)));
```
niklaspandersson, line 135: Great find! I guess not a lot of people use the CSB

src/accelerator/ogl/image/shader.frag
```diff
@@ -519,13 +554,132 @@ vec4 get_rgba_color()
...
+// ---- Tonal Balance (Shadows / Highlights separation) ----
+// Soft luminance-based masks drive additive offsets.  No clamping so HDR
+// headroom is preserved for downstream tonemapping.
+vec3 apply_tone_balance(vec3 c, float shadows, float highlights)
+{
+    // dot(c.bgr, ...) — c is BGR here, so an unswizzled dot applies red's Rec.709
+    // coefficient to blue and blue's to red. Greys are unaffected by the exchange,
+    // which is why only saturated colour was ever wrong. ContrastSaturationBrightness
+    // above already does this by hand via luma_coeff.bgr.
+    float lum         = dot(c.bgr, vec3(0.2126, 0.7152, 0.0722));
```
niklaspandersson: These coefficients, to get the luma from rgb, you have the same vec3 hardcoded in three places: apply_tone_balance, apply_split_tone and apply_cdl. Also, Those coefficients are already passed to the shader in the luma_coeff uniform (that you so convieniently fixed as well). Using the uniform gets the added benefit of having the correct coefficients based on the channels gammut (bt709, bt2020 etc)

niklaspandersson:
This looks good to me now. 

@C-alcaide, what are your thought on the new luma_coeff commit? With your use-cases, would the height < 700 => bt601 heuristic be a problem compared to always using bt709 coefficients?

Also, your ACES PR will allow the user to override that gammut, right?

C-alcaide:
Yes I was already preparing a PR about the 700 lines logic.

It makes sense that if the channel is set to PAL or NTSC it could still apply that (mostly so we don't break current functionality), but it should never apply it for custom resolution channels.
 And also manual override with the new command is always an option.

PR #1770, opened 2026-08-02
fix(decklink): free model name string returned by GetModelName()

dedicatedbroadcastsolutions:
Updated: (Note, this is an AI assisted modification)

### What
`get_model_name()`, `get_mode_name()` and `version()` in
`src/modules/decklink/util/util.h` each obtain a `String` from a DeckLink SDK
`Get*()` call (`GetModelName`/`GetName`/`GetString`) and pass it to
`to_string()`. On non-Windows platforms these calls return a heap-allocated
C string that the caller owns and must free, but `to_string()` only
converted it, leaking it on every call.

### Why
On Windows, `String` is a `BSTR` and `to_string()` already takes ownership
of it via `bstr_t(bstr_string, false)`, whose destructor frees the BSTR.
An earlier version of this PR freed the string again at the `get_model_name()`
call site via `SysFreeString()`, which would double-free on Windows
(as pointed out in review) since `to_string()` already owns and frees it.

### Change
Rework the non-Windows `to_string()` overload in `decklink_api.h` to take
ownership of its input and `free()` it after conversion, mirroring the
existing Windows behaviour. This fixes the leak uniformly for all three
call sites, with no risk of double-freeing on Windows. No behavioral/output
change - this is purely a memory-leak fix.

### DIFF
--- a/src/modules/decklink/decklink_api.h
+++ b/src/modules/decklink/decklink_api.h
@@ -112,6 +112,7 @@ T* get_raw(const CComPtr<T>& ptr)
 #include "linux_interop/DeckLinkAPIConfiguration.h"
 #include "linux_interop/DeckLinkAPIConfiguration_v10_11.h"
 #include "linux_interop/DeckLinkAPI_v10_11.h"
+#include <cstdlib>
 #include <memory>
 #include <typeinfo>
 
@@ -124,7 +125,13 @@ using BOOL     = bool;
 using UINT32   = uint32_t;
 using LONGLONG = int64_t;
 
-static std::wstring to_string(String utf16_string) { return u16(utf16_string); }
+// Takes ownership of utf16_string, matching the BSTR to_string() overload above, which also frees its input.
+static std::wstring to_string(String utf16_string)
+{
+    auto result = u16(utf16_string);
+    free((void*)utf16_string);
+    return result;
+}
 
 static void com_initialize() {}
 


src/modules/decklink/util/util.h
```diff
@@ -248,8 +248,15 @@ template <typename T>
 static std::wstring get_model_name(const T& device)
 {
     String pModelName;
-    if (SUCCEEDED(device->GetModelName(&pModelName)))
-        return to_string(pModelName);
+    if (SUCCEEDED(device->GetModelName(&pModelName))) {
+        std::wstring result = to_string(pModelName);
+#if defined(_MSC_VER)
+        ::SysFreeString(pModelName);
```
Julusian: I dont think this is safe. The [docs ](https://learn.microsoft.com/en-us/cpp/cpp/bstr-t-bstr-t?view=msvc-170) say that the way we implement `to_string` on windows it is taking ownership of the string

So perhaps the fix should be to rework the implementation of `to_string` on linux to have the same behaviour. There are other uses of `to_string` that I wonder if they will have the same issue

dedicatedbroadcastsolutions: Good point, I pushed some changes to attempt to implement that.

dedicatedbroadcastsolutions:
changed implementation to fix to_string for linux instead of only one particular usage.

PR #1771, opened 2026-08-09
fix(decklink): replace removed av_opt_set_int_list macro with av_opt_set_bin

dedicatedbroadcastsolutions:
## Problem
The Linux CI build (`archlinux:latest` matrix job) fails with:
```
modules/decklink/producer/decklink_producer.cpp:220:16: error:
  'av_opt_set_int_list' was not declared in this scope; did you mean 'av_opt_set_int'?
```
`av_opt_set_int_list` was a convenience macro in FFmpeg's `libavutil/opt.h` that has been removed in newer FFmpeg releases (Arch's rolling `archlinux:latest` image currently ships FFmpeg 8, where it's gone). This isn't related to any CasparCG code change — it broke because Arch's `ffmpeg` package was updated between CI runs; this same job passed as recently as 2026-08-04.

## Fix
The macro only ever expanded to a call to the stable, non-deprecated `av_opt_set_bin()`:
```c
#define av_opt_set_int_list(obj, name, val, term, flags) \
    (av_int_list_length(val, term) > INT_MAX / sizeof(*(val)) ? \
     AVERROR(EINVAL) : \
     av_opt_set_bin(obj, name, (const uint8_t *)(val), \
                    av_int_list_length(val, term) * sizeof(*(val)), flags))
```
Since the three call sites here use fixed-size local arrays with a known terminator, this replaces each `av_opt_set_int_list(...)` call with the equivalent direct `av_opt_set_bin(...)` call, computing the byte size at compile time via `sizeof(array) - sizeof(*array)` (excluding the terminator element, matching the macro's behavior). This works unchanged across FFmpeg 6.1 through 8.x+, since `av_opt_set_bin` itself was never removed.

## Note
An alternative, more future-proof fix using the newer `av_opt_set_array()` API (for FFmpeg versions that have it) already exists as part of #1637's much larger closed-captions refactor, but that PR is a large feature change. This PR is a minimal, standalone fix scoped only to unblocking the CI build.

## Testing
- Verified against FFmpeg 6.1 headers (`libavutil-dev`/`libavfilter-dev` on Ubuntu) that `av_opt_set_bin` has the same signature/behavior as used by the removed macro.
- clang-format clean.

### DIFF
--- a/src/modules/decklink/producer/decklink_producer.cpp
+++ b/src/modules/decklink/producer/decklink_producer.cpp
@@ -216,16 +216,29 @@ struct Filter
             FF(avfilter_graph_create_filter(
                 &sink, avfilter_get_by_name("buffersink"), "out", nullptr, nullptr, graph.get()));
 
+            // av_opt_set_int_list() was removed in newer FFmpeg; av_opt_set_bin() is the stable underlying call
             AVPixelFormat pix_fmts[] = {pix_fmt, AV_PIX_FMT_NONE};
-            FF(av_opt_set_int_list(sink, "pix_fmts", pix_fmts, -1, AV_OPT_SEARCH_CHILDREN));
+            FF(av_opt_set_bin(sink,
+                              "pix_fmts",
+                              reinterpret_cast<const uint8_t*>(pix_fmts),
+                              sizeof(pix_fmts) - sizeof(*pix_fmts),
+                              AV_OPT_SEARCH_CHILDREN));
         } else if (type == AVMEDIA_TYPE_AUDIO) {
             FF(avfilter_graph_create_filter(
                 &sink, avfilter_get_by_name("abuffersink"), "out", nullptr, nullptr, graph.get()));
 
             AVSampleFormat sample_fmts[]  = {AV_SAMPLE_FMT_S32, AV_SAMPLE_FMT_NONE};
             int            sample_rates[] = {format_desc.audio_sample_rate, 0};
-            FF(av_opt_set_int_list(sink, "sample_fmts", sample_fmts, -1, AV_OPT_SEARCH_CHILDREN));
-            FF(av_opt_set_int_list(sink, "sample_rates", sample_rates, 0, AV_OPT_SEARCH_CHILDREN));
+            FF(av_opt_set_bin(sink,
+                              "sample_fmts",
+                              reinterpret_cast<const uint8_t*>(sample_fmts),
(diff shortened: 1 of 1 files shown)


dedicatedbroadcastsolutions:
This is a CI fix extracted from Pull request 1637 and is not really needed if that pull request is merged.

Julusian:
While this fixes the build error, I think it is highlighting a logic error with ffmpeg 8.0+

for the ffmpeg producer, we are already using a different code path for ffmepg 8+, as the properties to be set have changed, and that change will resolve this build error too:

https://github.com/CasparCG/server/blob/685c2317e2593a92a0c2b64a108cd178eb761b6a/src/modules/ffmpeg/producer/av_producer.cpp#L583-L620

dedicatedbroadcastsolutions:
This PR is intentionally a minimal, temporary fix — it mirrors the existing inline `#if LIBAVUTIL_VERSION_MAJOR >= 60` pattern already used in `av_producer.cpp`/`ffmpeg_consumer.cpp`, just to unblock the currently-broken CI build with the smallest possible change.

#1637 has a better, more maintainable long-term fix for this: it introduces a shared `av_opt_array_ref` helper (in `av_util.h`/`.cpp`) that replaces this whole block in `decklink_producer.cpp` with `create_buffersink()`/`create_abuffersink()` calls, avoiding the duplicated version-gating logic entirely for this file. (Note: it doesn't yet consolidate the same pattern in `av_producer.cpp`/`ffmpeg_consumer.cpp` — those still use the inline style — so full deduplication across all three call sites is still a follow-up either way.)

Once #1637 merges, it will simply replace this block, so there's no conflict in landing this first. Happy to close this in favor of #1637 whenever that's ready to merge — until then, this fixes for the new arch test without waiting on review of the larger feature.  (please do still review it)

src/modules/decklink/producer/decklink_producer.cpp
```diff
@@ -217,15 +217,43 @@ struct Filter
                 &sink, avfilter_get_by_name("buffersink"), "out", nullptr, nullptr, graph.get()));
 
             AVPixelFormat pix_fmts[] = {pix_fmt, AV_PIX_FMT_NONE};
+            // FFmpeg 8 removed av_opt_set_int_list() and renamed pix_fmts -> pixel_formats with a typed-array API
```
Julusian:
```suggestion
```

Issue #1364, opened 2021-02-03
CasparCG in Docker does not accept UTF-8 in AMCP commands


toresbe:
### Expected behaviour

A file containing non-ASCII characters can be played.

### Current behaviour

CasparCG fails silently, logging a decode error
```
[2021-02-03 12:17:33.317] [info]    async_event_server[:5250] Accepted connection from 172.17.0.1 (2 connections).
[2021-02-03 12:17:33.318] [info]    Received message from 172.17.0.1: PLAY 1-50 "626646/broadcast/Livet og d?den.dv" "-filter:a aformat=sample_rates=48000" MIX 50 1 LINEAR R?\r\nSEEK 87141
[2021-02-03 12:17:33.318] [error]   Exception: libs/log/src/code_conversion.cpp(130): Throw in function (unknown)
[2021-02-03 12:17:33.318] [error]   Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::log::v2s_mt_posix::conversion_error> >
[2021-02-03 12:17:33.318] [error]   std::exception::what: Could not convert character encoding
[2021-02-03 12:17:33.318] [error]   
[2021-02-03 12:17:33.318] [error]    0# caspar::log::get_stack_trace[abi:cxx11]() in bin/casparcg
[2021-02-03 12:17:33.318] [error]    1# caspar::IO::connection::handle_read(boost::system::error_code const&, unsigned long) in bin/casparcg
[2021-02-03 12:17:33.318] [error]    2# boost::asio::detail::reactive_socket_recv_op<boost::asio::mutable_buffers_1, std::_Bind<void (caspar::IO::connection::*(caspar::spl::shared_ptr<caspar::IO::connection>, std::_Placeholder<1>, std::_Placeholder<2>))(boost::system::error_code const&, unsigned long)> >::do_complete(void*, boost::asio::detail::scheduler_operation*, boost::system::error_code const&, unsigned long) in bin/casparcg
[2021-02-03 12:17:33.318] [error]    3# 0x0000560D3F56D069 in bin/casparcg
[2021-02-03 12:17:33.318] [error]    4# 0x00007F7D845106DF in /usr/lib/x86_64-linux-gnu/libstdc++.so.6
[2021-02-03 12:17:33.318] [error]    5# 0x00007F7D7D25D6DB in lib/libpthread.so.0
[...]

Issue #1018, opened 2018-06-06
AMCP filenames are case sensitive for some (æøåÆØÅ) characters 


Julusian:
### Expected Behaviour
`PLAY 1-10 test-øVELSE-2` and `PLAY 1-10 test-ØVELSE-2` should both be able to play the same file

### Current Behaviour
Only one of them manages to play the file. The other fails with file not found.

Media-scanner correctly converts the case of the filenames, which means that CLS can report a filename which will not play due to it now being the 'wrong' casing.

### Steps to Reproduce 
* Rename a media file to test-ØVELSE-2
* Try to run the play command with different casing of the filename, in particular the Ø

### Environment
* CasparCG Server version: 2.2.0 8dcbc80
* Operating system: Win 10

dotarmin:
Hi @Julusian!

I'm not sure if I labeled this correctly as a bug but I'm doing so anyway. The file should play neither less what case is used. In my world it should already work because Windows it-self is not case sensitive as Unix but it should be fixed to work smoothly.

/Armin

rrebuffo:
Is there any news on this? Current builds still carry this bug. Lowercase filenames with special characters won’t play. Images play just fine in any case.
Could it be some bug in the boost filesystem library?

ronag:
It's probably a to lower/upper conversion which is based on ascii.

Julusian:
Digging into this, I have found https://github.com/CasparCG/server/blob/master/src/shell/main.cpp#L73-L74 which suggests that the locale we use internally is being forced to a bad value. This line suggests that we will never handle unicode correctly, as we are ignoring what the os wants.
The `sscanf` issue it refers to might be fixed upstream, as since that commit ffmpeg has made their own locale agnostic `sscanf` implementation https://github.com/FFmpeg/FFmpeg/commits/master/libavutil/avsscanf.c. But this is not being used everywhere so reverting this `sscanf` fix could reveal bugs inside of ffmpeg.

But by forcing the locale used when calling `iequals`, this has been fixed

Issue #1043, opened 2018-07-16
Exception when try display error on national language


drakmor:
May'be help - https://stackoverflow.com/questions/3760731/exceptions-with-unicode-what

### Expected Behaviour
Get normal text of current error ("can't connect")

### Current Behaviour
Boost exception and no valid error message

### Steps to Reproduce 
0. Set Russian system locale
1. Close "scanner.exe"
2. Run "casparcg.exe"
3. Send "CLS" command

### Environment
* CasparCG Server version: 2.2.0 e113f8ca
* Operating system: Windows 10 LTSB (Russian)

### Attachments
```
[2018-07-16 19:45:17.127] [info]    Received message from 127.0.0.1: CLS\r\n
[2018-07-16 19:45:19.130] [error]   Exception: libs\log\src\code_conversion.cpp(130): Throw in function (unknown)
[2018-07-16 19:45:19.130] [error]   Dynamic exception type: class boost::exception_detail::clone_impl<struct boost::exception_detail::error_info_injector<class boost::log::v2s_mt_nt6::conversion_error> >
[2018-07-16 19:45:19.130] [error]   std::exception::what: Could not convert character encoding
[2018-07-16 19:45:19.130] [error]
[2018-07-16 19:45:19.130] [error]    0# 0x00007FF7528E934E in casparcg
[2018-07-16 19:45:19.130] [error]    1# 0x00007FF7528E8C7F in casparcg
[2018-07-16 19:45:19.130] [error]    2# 0x00007FF752DE5A67 in casparcg
[2018-07-16 19:45:19.130] [error]    3# _C_specific_handler in VCRUNTIME140
[2018-07-16 19:45:19.130] [error]    4# _TypeMatch in VCRUNTIME140
[2018-07-16 19:45:19.130] [error]    5# RtlCaptureContext in ntdll
[2018-07-16 19:45:19.130] [error]    6# 0x00007FF752A8D7BB in casparcg
[2018-07-16 19:45:19.130] [error]    7# 0x00007FF752951548 in casparcg
[2018-07-16 19:45:19.130] [error]    8# 0x00007FF7529BBD9F in casparcg
[2018-07-16 19:45:19.130] [error]    9# 0x00007FF7529B9010 in casparcg
[2018-07-16 19:45:19.130] [error]   10# 0x00007FF7528E4069 in casparcg
[2018-07-16 19:45:19.130] [error]   11# o__strtoui64 in ucrtbase
[...]

PR #1772, opened 2026-08-09
fix(core): force UTF-8-capable global locale instead of trusting the environment

dedicatedbroadcastsolutions:
## Problem
`setup_global_locale()` derives the process-wide codepage (codecvt) facet from the deployment environment (`boost::locale::generator gen("")`). If that environment doesn't have a UTF-8 locale configured — common on minimal Docker images, or misconfigured setups — the resulting locale's codecvt facet can't represent non-ASCII characters.

The AMCP protocol mandates UTF-8 for all communication ("All communication is presumed to be encoded in UTF-8" — [AMCP-Protocol.md](https://github.com/CasparCG/help/wiki/AMCP-Protocol)), and this is already hardcoded elsewhere (`AMCPProtocolStrategy.cpp` decodes incoming bytes with the literal `"UTF-8"`, not a configurable/environment-derived charset). But the *global* locale used for logging (and, historically, for filename case-insensitive matching) was still silently dependent on whatever the environment provides.

This caused Boost.Log to throw `conversion_error` the moment a non-ASCII character (e.g. in a filename) was logged, aborting command processing **before** it ever reached the AMCP parser — see #1364, where a `PLAY` command with a non-ASCII filename failed silently for exactly this reason. #1018 is a related report of the same environment-locale fragility, there affecting filename case-insensitivity (`Ø`/`ø`), fixed separately by constructing a local `std::locale("")` in `common/filesystem.cpp` rather than relying on the global one.

#1707 already stopped the resulting *crash* by sanitizing logged output with `replace_nonprintable_copy()` — but that's a mitigation: it still substitutes `?` for any character the environment's locale can't represent, and only for the specific call sites it touched. This PR fixes the actual root cause.

## Fix
Force the codepage facet to always support UTF-8, via `std::codecvt_utf8<wchar_t>`, instead of trusting the environment:

```cpp
std::locale::global(std::locale(std::locale::classic(), new std::codecvt_utf8<wchar_t>));
```

This requires no OS locale data or ICU, so it behaves identically on any platform/container regardless of configuration — unlike naming a specific locale (e.g. `"en_US.UTF-8"`), which isn't guaranteed to be installed.

### Why this is safe
Checked every other consumer of the global locale in the codebase:
- `common/filesystem.cpp` (the #1018 fix) and `shell/linux_specific.cpp` (console I/O) construct their own separate `std::locale("")` — unaffected.
- `AMCPCommandsImpl.cpp` explicitly imbues a specific file stream with `"en_US.ISO8859-1"` for a legacy file format — unaffected (local imbue, not global).
- `vanc_op47_strategy.cpp`/`vanc_scte104_strategy.cpp` use the global locale's `ctype` facet, not `codepage` — `setup_global_locale()` never touched `ctype` before or after this change, so unaffected.
- Boost.Log's sink formatting is the one real consumer of the global codepage facet — exactly what needed fixing.

## Verification
Reproduced the exact failure mechanism outside the full build: used the same `std::codecvt<wchar_t,char>` facet Boost.Log's `code_conversion.cpp` relies on internally, under the current code vs. this fix, across several environments:

```
BEFORE, LC_ALL=C (minimal Docker-like env):        codecvt_base::error   (-> the thrown conversion_error)
BEFORE, LC_ALL=C.UTF-8:                             OK
AFTER,  LC_ALL=C:                                   OK - "Livet og døden.dv" (correct, full fidelity)
AFTER,  completely empty environment (env -i):      OK - "Livet og døden.dv"
```

Confirms this resolves the issue even in the worst-case environment, with no regression when a proper locale is already configured, and with full character fidelity (not just `?` substitution).

### DIFF
--- a/.github/workflows/linux-system.yml
+++ b/.github/workflows/linux-system.yml
@@ -48,13 +48,7 @@ jobs:
                       container: debian:sid # Debian Testing
                       apt_prefix: ""
                       upload_artifacts: false # Because of CEF, this is not built correctly
-                    - os: ubuntu-latest
-                      family: debian
-                      distribution: trixie
-                      arch: amd64
-                      container: debian:trixie
-                      apt_prefix: ""
-                      upload_artifacts: false # Because of CEF, this is not built correctly
+                      experimental: true
                     - os: ubuntu-latest
                       family: archlinux
                       distribution: latest
@@ -62,9 +56,11 @@ jobs:
                       container: archlinux:latest
                       apt_prefix: ""
                       upload_artifacts: false # Not supported
+                      experimental: true
 
         runs-on: ${{ matrix.os }}
         container: ${{ matrix.container }}
+        continue-on-error: ${{ matrix.experimental || false }} # Stop failures of smoke tests being loud
         defaults:
             run:
                 shell: bash
--- a/README.md
+++ b/README.md
@@ -56,8 +56,7 @@ Getting Started
 Documentation
 -------------
 
-The most up-to-date documentation is always available at
-https://github.com/CasparCG/help/wiki
+The most up-to-date documentation is always available at https://casparcg.com/docs/wiki
 
 Ask questions in the forum: https://casparcgforum.org/
 
--- a/src/CMakeModules/Bootstrap_Linux.cmake
+++ b/src/CMakeModules/Bootstrap_Linux.cmake
@@ -87,6 +87,14 @@ if (ENABLE_HTML)
                 "<SOURCE_DIR>/Release/libcef.so"
                 "<BINARY_DIR>/libcef_dll_wrapper/libcef_dll_wrapper.a"
         )
+        # CEF needs the resources next to libcef.so
+        ExternalProject_Add_Step(cef copy_resources
+                COMMAND ${CMAKE_COMMAND} -E copy_directory
+                "<SOURCE_DIR>/Resources"
+                "<SOURCE_DIR>/Release"
+                DEPENDEES build
+                DEPENDERS install
+        )
         ExternalProject_Get_Property(cef SOURCE_DIR)
         ExternalProject_Get_Property(cef BINARY_DIR)
 
--- a/src/CMakeModules/Bootstrap_Windows.cmake
+++ b/src/CMakeModules/Bootstrap_Windows.cmake
@@ -91,8 +91,8 @@ add_definitions( -DBOOST_LOCALE_HIDE_AUTO_PTR )
 # FFMPEG
 casparcg_add_external_project(ffmpeg-lib)
 ExternalProject_Add(ffmpeg-lib
-	URL ${CASPARCG_DOWNLOAD_MIRROR}/ffmpeg/ffmpeg-7.0.2-full_build-shared.7z
-	URL_HASH MD5=c5127aeed36a9a86dd3b84346be182f8
+	URL ${CASPARCG_DOWNLOAD_MIRROR}/ffmpeg/ffmpeg-8.1.2-full_build-shared.7z
+	URL_HASH SHA256=cba748035c21ce1431d0823c7a3a711f38616f89f87a265dceddf9b7f6749d2d
 	DOWNLOAD_DIR ${CASPARCG_DOWNLOAD_CACHE}
 	CONFIGURE_COMMAND ""
 	BUILD_COMMAND ""
@@ -102,14 +102,13 @@ ExternalProject_Get_Property(ffmpeg-lib SOURCE_DIR)
 set(FFMPEG_INCLUDE_PATH "${SOURCE_DIR}/include")
 set(FFMPEG_BIN_PATH "${SOURCE_DIR}/bin")
 link_directories("${SOURCE_DIR}/lib")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avcodec-61.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avdevice-61.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avfilter-10.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avformat-61.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avutil-59.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/postproc-58.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/swresample-5.dll")
-casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/swscale-8.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avcodec-62.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avdevice-62.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avfilter-11.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avformat-62.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/avutil-60.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/swresample-6.dll")
+casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/swscale-9.dll")
 # for scanner:
 casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/ffmpeg.exe")
 casparcg_add_runtime_dependency("${FFMPEG_BIN_PATH}/ffprobe.exe")
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -192,8 +192,17 @@ struct image_kernel::impl
         const auto is_hd       = params.pix_desc.planes.at(0).height > 700;
         const auto color_space = is_hd ? params.pix_desc.color_space : core::color_space::bt601;
 
+        // Row order is R, G, B; columns are Y, Cb, Cr. Each row follows from the
+        // luma coefficients: Cr->R is 2(1-Kr), Cb->B is 2(1-Kb), and the two green terms
+        // are -2(1-Kb)Kb/Kg and -2(1-Kr)Kr/Kg.
+        //
+        // The BT.601 Cr->G term was -0.509 and should be -0.714136:
+        // -2(1-0.299)(0.299)/0.587. BT.709 and BT.2020 were both already correct, which
+        // is what made this hard to see -- only SD material was affected, and only in
+        // green. Measured on flat patches decoded as BT.601, green was out by up to 20
+        // LSB on saturated colour while red and blue were exact.
         const float color_matrices[3][9] = {
-            {1.0, 0.0, 1.402, 1.0, -0.344, -0.509, 1.0, 1.772, 0.0},                          // bt.601
+            {1.0, 0.0, 1.402, 1.0, -0.344136, -0.714136, 1.0, 1.772, 0.0},                    // bt.601
             {1.0, 0.0, 1.5748, 1.0, -0.1873, -0.4681, 1.0, 1.8556, 0.0},                      // bt.709
             {1.0, 0.0, 1.4746, 1.0, -0.16455312684366, -0.57135312684366, 1.0, 1.8814, 0.0}}; // bt.2020
         const auto color_matrix = color_matrices[static_cast<int>(color_space)];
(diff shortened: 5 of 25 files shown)


dedicatedbroadcastsolutions:
ACMP spec states UTF-8## Issues addressed

- Fixes #1364 — root cause of the reported crash/silent `PLAY` failure on non-ASCII filenames in Docker: `setup_global_locale()` trusted the deployment environment for a UTF-8-capable locale. #1707 already stopped the crash via output sanitization; this fixes the actual root cause so the command executes correctly rather than just avoiding the crash.
- Related to #1018 (already closed, fixed separately) — same root cause (environment-dependent locale), there affecting Unicode case-insensitive filename matching. This hardens that same code path (`common/filesystem.cpp`, `common/os/linux/filesystem.cpp`) so it no longer depends on the deployment environment having a real locale installed.
- Also fixes an unreported, more severe instance of the same bug found while investigating the above: `shell/linux_specific.cpp`'s startup console I/O setup used the same unguarded `std::locale("")`, which runs unconditionally during server startup — a misconfigured environment could have prevented the server from starting at all.

dedicatedbroadcastsolutions:
I still need to test this in an actual build.  If anyone does that please comment so we can confirm if this is ready to mark for review.

Julusian:
I know I have battled character encoding in the past, so am not confident in any change which touches it.

I got claude to do a dive and give a summary. I havent verified all of it, but it seems sane/reasonable

----
 
## Blocking
 
### 1. `std::codecvt_utf8<wchar_t>` is the wrong facet on Windows
 
`std::codecvt_utf8` converts UTF-8 ↔ UCS-2/UCS-4, not UTF-16. With 16-bit `wchar_t` on Windows, anything outside the BMP arrives as a surrogate pair and is rejected as an invalid code point, so emoji still throw `conversion_error` — the exact #1707 case this PR describes as an inadequate mitigation.
 
It isn't a regression (the ANSI codepage facet fails there too), but it does mean the "full fidelity, not `?` substitution" claim only holds on Linux, which is also the only platform in the verification table. The standard-library pairing for 16-bit `wchar_t` is `std::codecvt_utf8_utf16`, so a portable version of this fix would need a conditional on `wchar_t` width.
 
### 2. `<codecvt>` is deprecated and removed in C++26
 
Deprecated in C++17 by P0618, removed in C++26 by P2871R3 (which also takes out `wstring_convert`/`wbuffer_convert`). We build at `cxx_std_20` (`src/CMakeModules/CasparCG_Util.cmake:36`, `src/shell/CMakeLists.txt:37`), so it compiles today but warns — MSVC needs `_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING`. Not somewhere we want new code.
 
### Both are solved by using Boost.Locale, which is already a dependency
 
```cpp
std::locale::global(std::locale(std::locale::classic(),
                                new boost::locale::utf8_codecvt<wchar_t>));
```
 
`boost::locale::utf8_codecvt<CharType>` is documented as converting UTF-8 to UTF-16 **and** UTF-32 via `wchar_t`, so the surrogate case is handled correctly. Like the proposed fix it needs no OS locale data and no ICU, so the environment-independence argument is unchanged. `boost::locale::util::create_utf8_codecvt(base, char_facet_t::wchar_f)` is the factory equivalent if you prefer. Same one-line blast radius, same audit, neither objection.
 
## Scoping
 
### #1707 can't be reverted on the back of this
 
Worth stating explicitly so nobody assumes otherwise. On Windows the sanitization stays load-bearing for non-BMP input, and on Linux it still covers call sites that reach the file sink, where `my_formatter` runs with `print_all_characters = true`.
 
### "Root cause" should be narrowed to logging
 
`common/filesystem.cpp`, `common/os/linux/filesystem.cpp` and `ffmpeg_producer.cpp` each construct their own `std::locale("")` for case-insensitive matching, and those stay environment-dependent after this change. That isn't a fair thing to ask you to fix in the same PR: a codecvt can be synthesized from nothing, which is exactly why this approach works, but case folding needs real Unicode tables from OS locale data or ICU. There's no equivalent one-liner. Just please reword the claim — this fixes the environment dependence *of logging*, not of everything.
 
Related, and worth settling before this lands: 9795c780 (the #1018 fix) carries a note that its local `std::locale("")` "might be unnecessary once main.cpp's locale setup has been removed." After this PR the global is `classic()` plus a UTF-8 codecvt, so its `ctype` facet is still the ASCII-only classic one and those call sites can never drop the local locale. If the long-term intent was to delete `setup_global_locale()` entirely, this moves in the opposite direction and cements the workaround.
 
### Adjacent bug, if you want it in scope
 
`log.h::replace_nonprintable` constructs `std::locale("")` **per logged message** — a per-line allocation in the log path, and an unguarded throw if `LANG` names a locale that isn't installed (common on slim images that set `LANG=en_US.UTF-8` without generating it). Same class of environment fragility this PR is about, and a cheap fix if you want to fold it in.
 
## Verified — no need to re-argue
 
- The file sink genuinely is the only consumer of the global codepage facet. `add_cout_sink()` uses `wtext_ostream_backend` + `wformatting_ostream`, wide all the way into `std::wcout`, which carries its own locale from `linux_specific.cpp`. `add_file_sink()` is the narrow one and the synchronous one, which is why the throw surfaces on the caller's thread in `connection::handle_read`.
- `classic()` as the base is faithful. `boost::locale::generator` already defaults to `std::locale::classic()`, and only the `codepage` category was enabled, so ctype/numeric behaviour is unchanged. The VANC `use_facet<ctype<wchar_t>>` sites and the `setlocale(LC_ALL, "C")` FFmpeg workaround are genuinely unaffected.
- Ordering is correct: `setup_global_locale()` at `main.cpp:219` precedes `add_cout_sink()` (236) and `add_file_sink()` (254), so both sinks capture the new locale at construction.
- One point in favour that the writeup doesn't make: the second `setup_global_locale()` call exists because CEF resets the global locale. Making that reset environment-independent is a real robustness win on its own.
- 
## Re-test matrix
 
The verification so far tested the codecvt facet standalone, which validates the mechanism but not the server. Could we get these on a real build, on **both Windows and Linux** — `wchar_t` width is the whole issue.
 
| # | Test | Why |
|---|---|---|
| 1 | `CG UPDATE` with non-BMP emoji (😀, 🇳🇱, 👨‍👩‍👧‍👦) on **Windows** | The surrogate-pair failure above |
| 2 | Same on Linux under `LC_ALL=C` and `env -i` | Should render fully, not `?` |
| 3 | BMP non-ASCII on a Russian-locale Windows box (#1043) | The other historical Windows path |
| 4 | Check the **log file**, not just the console | `add_file_sink` is the narrow sink with `print_all_characters = true` — that's where fidelity actually changes. Console goes through `wcout`'s own locale and should not move at all |
| 5 | #1364 end to end: minimal Docker image, `PLAY` a file with `ø` in the name | Confirm the command reaches the parser instead of failing silently |
| 6 | #1018 in that same container: `PLAY` with `Ø` vs `ø` | Expected to still fail there, since `std::locale("")` resolves to `C` with no case mappings past ASCII. Documents the boundary rather than assuming it. Works fine under `C.UTF-8`, which uses glibc's i18n `LC_CTYPE` |
| 7 | FFmpeg decimal separator on a `de_DE`/`fr_FR` host: `-filter:a pan=...` and float `MIXER` values | The `setlocale(LC_ALL, "C")` interaction from 3d946dc0 / 3ae143dd — historically the most fragile thing in this function |
| 8 | Decklink VANC: op47 teletext and SCTE-104 | They use `use_facet<ctype<wchar_t>>(std::locale())`. The audit says ctype is untouched and I agree, but it's a cheap smoke test |
| 9 | `DATA RETRIEVE` on a legacy Latin-1 file with no BOM | `read_latin1_file`'s `en_US.ISO8859-1` imbue (5ebb4be8) — local, should be unaffected |
| 10 | HTML producer running, check CEF console messages log correctly | Exercises the second `setup_global_locale()` call after CEF resets the global. Note the sinks already exist by then, so that reset cannot re-imbue them |
| 11 | Malformed or truncated UTF-8 from a client | The `conv::skip` path in `strategy_adapters.cpp` |

dedicatedbroadcastsolutions:
I'll take a look and process that report later.  This one may take a while to finish and sit for a bit as I need to get some toolchains set up to properly test it.  This was something that got built on the side when I was working on my url fix and I decided it needed to be a separate project.  This one may not get finished before the next release.

Note, I assume we want to keep the acmp requirement for UTF-8 intact.

dedicatedbroadcastsolutions:
Revision of description.

## Problem
`setup_global_locale()` derives the process-wide codepage (codecvt) facet from the deployment environment (`boost::locale::generator gen("")`). If that environment doesn't have a UTF-8 locale configured — common on minimal Docker images, or misconfigured setups — the resulting locale's codecvt facet can't represent non-ASCII characters.

The AMCP protocol mandates UTF-8 for all communication ("All communication is presumed to be encoded in UTF-8" — [AMCP-Protocol.md](https://github.com/CasparCG/help/wiki/AMCP-Protocol)), and this is already hardcoded elsewhere (`AMCPProtocolStrategy.cpp` decodes incoming bytes with the literal `"UTF-8"`, not a configurable/environment-derived charset). But the *global* locale used for logging was still silently dependent on whatever the environment provides.

This caused Boost.Log to throw `conversion_error` the moment a non-ASCII character (e.g. in a filename) was logged, aborting command processing **before** it ever reached the AMCP parser — see #1364, where a `PLAY` command with a non-ASCII filename failed silently for exactly this reason.

**Scope note:** this PR fixes the environment-dependence *of logging* specifically. It does not fix the separate, harder problem of case-insensitive filename matching (`common/filesystem.cpp`, `common/os/linux/filesystem.cpp`) remaining environment-dependent (#1018) — that needs real Unicode case-folding tables from OS locale data or ICU, which a synthesized codecvt can't provide. This PR does harden those call sites (and the Linux console I/O setup in `shell/linux_specific.cpp`, and `log::replace_nonprintable`) to prefer `C.UTF-8` and fall back gracefully instead of throwing on an unconfigured environment, but they still can't case-fold non-ASCII the same way a real locale would.

#1707 already stopped the resulting *crash* by sanitizing logged output with `replace_nonprintable_copy()` — that mitigation stays load-bearing (e.g. for non-BMP input on Windows) and is unaffected by this PR; it still substitutes `?` for anything the environment's locale can't represent, and only for the specific call sites it touched.

## Fix
Force the codepage facet to always support UTF-8 via `boost::locale::utf8_codecvt<wchar_t>` (already a project dependency), instead of trusting the environment:

```cpp
std::locale::global(std::locale(std::locale::classic(), new boost::locale::utf8_codecvt<wchar_t>));****

dedicatedbroadcastsolutions:
Checks someone needs to perform:

CG UPDATE with non-BMP emoji (😀, 🇵🇱, 👨‍👩‍👧‍👦) on Windows — the surrogate-pair failure case
Same on Linux under LC_ALL=C and env -i — should render fully, not ?
BMP non-ASCII on a Russian-locale Windows box (#1043)
Check the log file, not just the console — add_file_sink is the narrow sink with print_all_characters = true, where fidelity actually changes; console goes through wcout's own locale and should not move at all
#1364 end to end: minimal Docker image, PLAY a file with ø in the name
#1018 in that same container: PLAY with Ø vs ø — expected to still fail there (documents the boundary), works fine under C.UTF-8
FFmpeg decimal separator on a de_DE/fr_FR host: -filter:a pan=... and float MIXER values
Decklink VANC: op47 teletext and SCTE-104 (uses ctype, should be unaffected — cheap smoke test)
DATA RETRIEVE on a legacy Latin-1 file with no BOM — should be unaffected
HTML producer running, check CEF console messages log correctly — exercises the second setup_global_locale() call after CEF resets the global
Malformed or truncated UTF-8 from a client — the conv::skip path in strategy_adapters.cpp

dedicatedbroadcastsolutions:
AI analysis

## Follow-up on #1018 case-folding (re: scope note in the revised description above)

One correction to the scope note: this PR didn't ignore #1018 — commit `086a14b61` ("make case-insensitive file matching independent of deployment locale") already hardens `common/filesystem.cpp`/`common/os/linux/filesystem.cpp` to prefer `C.UTF-8` instead of blindly trusting the environment, same as `b8bcf5436` does for the console I/O setup in `shell/linux_specific.cpp`. So it was already attempted here, alongside the logging fix — just not fully closed on every platform.

Digging into how far that gets us: on **glibc-based Linux** (Debian/Ubuntu/Arch — all our CI targets), `C.UTF-8` actually resolves to a locale with full Unicode-aware `LC_CTYPE` case mapping, even though its collation stays plain ASCII order. So `Ø`/`ø` matching should already work there without any environment configuration — matching what @Julusian noted in the review ("works fine under `C.UTF-8`").

It's still not fully closed on:
- **Windows** — `"C.UTF-8"` isn't a recognized locale name in the MSVC CRT, so it falls through to the environment locale, then to `classic()` (ASCII-only) if that isn't configured either.
- **musl-based Linux** (e.g. Alpine) — musl doesn't support `C.UTF-8` or real locale data at all, same ASCII-only fallback.

Unlike the logging fix (`boost::locale::utf8_codecvt` needs no locale data at all and is fully environment-independent by construction), case-folding fundamentally needs real Unicode case tables from *somewhere* — there's no equivalent "just synthesize it" trick. Options to fully close the remaining gap:
1. **Link ICU directly** — bundles complete Unicode tables, works identically regardless of OS locale config. Adds a new build dependency though, and it's not clear from the current build scripts whether Boost.Locale is already linked against ICU here (Linux pulls Boost from distro packages, Windows from a prebuilt zip on CasparCG's mirror — neither confirms it).
2. **Hand-roll a small, portable Unicode case-folding table** covering Latin-1 Supplement + Latin Extended-A (covers `Ø/ø`, `ß`, and effectively all Western European filenames), with no new dependencies, identical behavior on every platform.
3. **Leave it as the documented boundary** it already is on Windows/musl — no crash either way, just no case-insensitive match for non-ASCII filenames there without a configured locale.

Given this is a separate, harder problem than the logging crash this PR primarily fixes, I'd suggest tracking full closure as follow-up work rather than blocking this PR on it — happy to pick up option 2 in a follow-up PR if there's interest.

PR #1775, opened 2026-08-14
fix(mixer): the sub-720 BT.601 rule overrode tagged metadata instead of filling in for it

C-alcaide:
Both mixers resolved the YCbCr decode matrix as

```cpp
is_hd ? pix_desc.color_space : color_space::bt601
```

which applies the SD convention as an **override** rather than a fallback. A source that
explicitly declared BT.709 at 960x540 or 1024x640 was matrixed as BT.601 on any channel,
including 2160p. Nothing downstream can undo it: the matrix is applied at texture-fetch
time, before any colour management.

## Why it could not simply be written as a fallback

`color_space` had no way to say *not declared* — it defaulted to `bt709`, so a file tagged
BT.709 and a file that declared nothing were the same value.

Following review, that state is now an enum value rather than a separate boolean.
`unknown` is added last (the value is a matrix index in both mixer shaders), and one
function resolves it:

```cpp
inline color_space decode_color_space(const pixel_format_desc& desc)
{
    if (desc.color_space != color_space::unknown) {
        return desc.color_space;
    }
    return desc.planes.at(0).height > 700 ? color_space::bt709 : color_space::bt601;
}
```

A declared space is honoured at any size; an undeclared one gets BT.601 below 720 lines and
BT.709 above, as before. Both kernels become a single call, so the two backends cannot
diverge on the convention.

Producers that read metadata report `unknown` when there is none — ffmpeg from
`AVCOL_SPC_UNSPECIFIED`, the DeckLink producer when the card carries no colorspace flag. A
producer that never set a colour space also says `unknown`, and resolves to exactly what the
`bt709` default resolved to before, so untouched producers are unchanged.

## Two things that follow from putting the state in the enum

**The DeckLink producer had the same defect.** With a separate boolean it was never set
there, so an SD-SDI feed the card tags Rec2020 was still forced to BT.601. Distinguishing
"card said Rec709" from "card said nothing" is three lines in the same function, and it is
required rather than optional: without it, untagged SD-SDI would silently move from BT.601
to BT.709.

**The destination cannot enter into the decision**, because `decode_color_space` takes only
the source descriptor. What matrix a file was encoded with is a property of the file; keying
off the channel would trade one guess for another and would read untagged BT.601 material as
BT.709 on any non-broadcast raster — a regression for exactly the SD content the convention
exists to serve. Where the guess is wrong and the file cannot be re-tagged, the answer is an
explicit override, not a second heuristic.

## Measurements

Each case builds four saturated, asymmetric patches, encodes them to YCbCr with a known
matrix, and tags the file with an independently chosen colorspace (or none). Both candidate
decodes are then predicted in closed form and the capture is matched against each, so a case
reports **which matrix was used** rather than passing a tolerance — a wrong tolerance cannot
turn one answer into the other, because the two predictions are 10–20 LSB apart on these
patches. Greys are near-invariant between the two matrices, which is why the patches are
asymmetric.

| clip | channel | decoded as | worst residual |
| :--- | :--- | :--- | ---: |
| tagged BT.709, 640x480 | 1080p | **bt709** | 1.66 LSB |
| tagged SMPTE170M, 640x480 | 1080p | bt601 | 0.41 LSB |
| untagged, truly BT.601 | 1080p | bt601 | 0.41 LSB |
| untagged, truly BT.709 | 1080p | bt601 | 1.23 LSB |
| untagged, truly BT.709 | 1280x720 custom | bt601 | 1.23 LSB |
| untagged, truly BT.601 | 1280x720 custom | bt601 | 0.41 LSB |

The first row is the defect: it decoded as `bt601` before this change.

The fourth row is the convention's known cost — untagged BT.709 content below 720 lines is
read as BT.601 — and is unchanged by this PR.

The last two rows are a regression guard rather than a feature: they assert that the
channel raster does not influence the decode.

**6/6 on OpenGL and 6/6 on Vulkan**, with residuals equal to two decimals between the two
backends.

### DIFF
--- a/src/accelerator/ogl/image/image_kernel.cpp
+++ b/src/accelerator/ogl/image/image_kernel.cpp
@@ -189,8 +189,7 @@ struct image_kernel::impl
             params.layer_key->bind(static_cast<int>(texture_id::layer_key));
         }
 
-        const auto is_hd       = params.pix_desc.planes.at(0).height > 700;
-        const auto color_space = is_hd ? params.pix_desc.color_space : core::color_space::bt601;
+        const auto color_space = core::decode_color_space(params.pix_desc);
 
         // Row order is R, G, B; columns are Y, Cb, Cr. Each row follows from the
         // luma coefficients: Cr->R is 2(1-Kr), Cb->B is 2(1-Kb), and the two green terms
--- a/src/accelerator/vulkan/image/image_kernel.cpp
+++ b/src/accelerator/vulkan/image/image_kernel.cpp
@@ -290,8 +290,7 @@ struct image_kernel::impl
             uniforms.precision_factor[n] = get_precision_factor(params.textures[n]->depth());
         }
 
-        const auto is_hd           = params.pix_desc.planes.at(0).height > 700;
-        const auto color_space     = is_hd ? params.pix_desc.color_space : core::color_space::bt601;
+        const auto color_space     = core::decode_color_space(params.pix_desc);
         uniforms.color_space_index = static_cast<uint32_t>(color_space);
 
         if (params.pix_desc.is_straight_alpha) {
--- a/src/core/frame/pixel_format.h
+++ b/src/core/frame/pixel_format.h
@@ -51,6 +51,10 @@ enum class color_space
     bt601,
     bt709,
     bt2020,
+
+    /// The source did not say; `decode_color_space` below resolves it. Must stay last,
+    /// because the value is a matrix index in both mixer shaders.
+    unknown,
 };
 
 struct pixel_format_desc final
@@ -79,7 +83,7 @@ struct pixel_format_desc final
 
     pixel_format_desc() = default;
 
-    explicit pixel_format_desc(pixel_format format, core::color_space color_space = core::color_space::bt709)
+    explicit pixel_format_desc(pixel_format format, core::color_space color_space = core::color_space::unknown)
         : format(format)
         , color_space(color_space)
     {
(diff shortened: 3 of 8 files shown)


Julusian:
Instead of a separate color_space_specified boolean that needs to be passed around, what about adding an 'unknown' or 'auto' value to the enum? 

Also please tell claude to cut down the verbosity f comments. It loves to write walls of comments to explain simple things obvious in the code or to explain what it changed and why, which just adds unnecessary noise and doesnt help the next reader

C-alcaide:
> Also please tell claude to cut down the verbosity f comments. It loves to write walls of comments to explain simple things obvious in the code or to explain what it changed and why, which just adds unnecessary noise and doesnt help the next reader

Yeah something is going on with the comments, it keeps repeating things everywhere for no reason. I'll look into it, I changed model and provider recently so it could be misconfigured in the .md instructions.

C-alcaide:
> Instead of a separate color_space_specified boolean that needs to be passed around, what about adding an 'unknown' or 'auto' value to the enum? 

That's a more appropiate solution for sure.

 Should be fine now.

Julusian:
yeah thats better

I think there are 2 cases which are missed here currently:
1) I believe the image producer can produce yuv frames, and is currently ignoring the `av_frame->colorspace` reported of those frames. Not 100% caused by this, but could cause some visual regressions as it loses the auto 601 rule. (and fixes others which shouldnt have had that rule)

2) `decode_color_space` should probably be treating any 'color_space::unknown' rgb frames as bt709 instead of the auto logic. The mixer expects all rgb to be srgb which uses bt709 primaries, so any logic that relies on the produced color_space should ideally follow that and not drop to 601. Again not really caused by this, so this could be a follow up

C-alcaide:
About item 2, what do you think about this alternative it suggested?

'instead of "treat unknown RGB as bt709", the cleaner statement is the sub-720 rule only applies to YCbCr formats — pixel_format_desc already carries format, so the check needs no plumbing, and framing it that way makes bt709-for-RGB fall out as a consequence rather than a special case.'

Julusian:
Yeah sure

C-alcaide:
About item 1:
I think the only way to fix a regression with stills I just saw after measuring was to ask for the color space for those cases when it's actually present. 

It's true that it's not a lot of cases (601 tagged Webp that I found) but at least it would cover similar ones I haven't tried as well.

Julusian:
As in you think more needs to be done on this, or your last commit solved it?

I'm happy with this now, unless you are aware of more potential issues

C-alcaide:
Last commit should do the trick.
Thank you

PR #1791, opened 2026-09-11
Add new channels library and use it in `ffmpeg::Decoder`

programmerjake:
This is split out from #1637.

This adds a channels API loosely inspired by [Rust's mpsc channels](https://doc.rust-lang.org/std/sync/mpsc/), though it only supports a single sender and receiver per channel, unlike the Rust channels.

Changing `ffmpeg::Decoder` over to use channels is needed because #1637 depends on it and because I think it makes it much easier to reason about. (#1637 does a bunch of redirecting channels which can't easily be done with the old mutex/cond-var implementation of `Decoder`).

### DIFF
--- a/src/common/channel.h
+++ b/src/common/channel.h
@@ -1,0 +1,330 @@
+/*
+ * Copyright (c) 2025 Jacob Lifshay <programmerjake@gmail.com>
+ *
+ * This file is part of CasparCG (www.casparcg.com).
+ *
+ * CasparCG is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * CasparCG is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with CasparCG. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+// use boost's types to support thread interruption
+#include <boost/thread/condition_variable.hpp>
+#include <boost/thread/mutex.hpp>
+
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <queue>
+#include <string>
+
+#ifdef CASPARCG_DEBUG_CHANNEL
+#include <sstream>
+
+#include "common/log.h"
+#endif
+
+namespace caspar {
+template <typename T>
+class Sender;
+
+template <typename T>
+class Receiver;
+
+struct channel_capacity_is_hint_t final
+{
+};
+
+inline constexpr channel_capacity_is_hint_t channel_capacity_is_hint{};
+
+template <typename T>
+std::pair<Sender<T>, Receiver<T>> channel(std::size_t capacity, std::string debug_name = {});
+
+template <typename T>
+std::pair<Sender<T>, Receiver<T>>
+channel(std::size_t capacity, channel_capacity_is_hint_t, std::string debug_name = {});
+
+template <typename T>
+class Receiver final
+{
+    template <typename>
+    friend class Sender;
+
+    template <typename T2>
+    friend std::pair<Sender<T2>, Receiver<T2>> channel(std::size_t capacity, std::string debug_name);
+
+    template <typename T2>
+    friend std::pair<Sender<T2>, Receiver<T2>>
+    channel(std::size_t capacity, channel_capacity_is_hint_t, std::string debug_name);
+
+  public:
+    Receiver() noexcept
+        : state_()
+    {
+    }
+    Receiver(const Receiver&)                = delete;
+    Receiver(Receiver&&) noexcept            = default;
+    Receiver& operator=(const Receiver&)     = delete;
+    Receiver& operator=(Receiver&&) noexcept = default;
+
+    bool closed() const
+    {
+        if (!state_)
+            return true;
+        auto lock = boost::unique_lock(state_->lock_);
+        return state_->queue_.empty() && state_->sender_destroyed_;
+    }
+    bool empty() const
+    {
+        if (!state_)
+            return true;
+        auto lock = boost::unique_lock(state_->lock_);
+        return state_->queue_.empty();
+    }
+    std::size_t capacity() const
+    {
+        if (!state_)
+            return 0;
+        return state_->capacity_;
+    }
+
+    struct TryReceiveResult final
+    {
+        std::optional<T> value;
+        bool             closed;
+    };
+    TryReceiveResult try_receive()
+    {
+        if (!state_)
+            return {std::nullopt, true};
+        auto lock = boost::unique_lock(state_->lock_);
+        return state_->try_receive_locked();
+    }
+    /// not named `receive` since that would imply throwing an exception when closed
+    std::optional<T> try_receive_blocking()
+    {
+        if (!state_)
+            return std::nullopt;
+        auto lock = boost::unique_lock(state_->lock_);
+
+        TryReceiveResult try_receive_result{std::nullopt, true};
+        state_->cond_.wait(lock, [&]() {
+            try_receive_result = state_->try_receive_locked();
+            return try_receive_result.value || try_receive_result.closed;
+        });
+        return std::move(try_receive_result.value);
+    }
+
+  private:
+    struct TrySendResult final
+    {
+        std::optional<T> unsent;
+        bool             closed;
+    };
+    struct State final
+    {
+        // use boost's types to support thread interruption
+        boost::mutex              lock_;
+        boost::condition_variable cond_;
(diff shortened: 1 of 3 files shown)


src/common/channel.h
```diff
@@ -0,0 +1,330 @@
...
+    return {Sender<T>(std::unique_ptr<State, typename Sender<T>::Deleter>(state)),
+            Receiver<T>(std::unique_ptr<State, typename Receiver<T>::Deleter>(state))};
+}
+
+template <typename T>
+std::pair<Sender<T>, Receiver<T>> channel(std::size_t capacity, channel_capacity_is_hint_t, std::string debug_name)
+{
+    using State = typename Receiver<T>::State;
+    auto state  = new State(capacity, true, std::move(debug_name));
+    return {Sender<T>(std::unique_ptr<State, typename Sender<T>::Deleter>(state)),
```
niklaspandersson, line 327: Please add a comment describing the Deleter. Taking two unique_ptrs to the same object always raises eyebrows.

src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -231,17 +231,14 @@ class Decoder
...
-                            output.push(std::move(av_frame));
+                        if (output_sender.try_send_blocking(av_frame)) {
+                            break; // no receiver -- we're done
                         }
                     }
                 }
             } catch (boost::thread_interrupted&) {
                 // Do nothing...
             } catch (...) {
-                eof = true;
```
niklaspandersson, line 244: As far as I can tell (with the help of Claude) this "EOF on error" functionality is not preserved in the refactor, leaving producers that raises any kind of decode error open indefinetely without making any progress.

Did you test this with known corrupt sources? Also, it might also be a good idea to test how looping / seeking is handled if you haven't already.

niklaspandersson:
I like the abstraction and improvement to readability that this brings.

There is an experimental channel implementation in boost. Did you try that before rolling your own?

Also, I took the help of Claude to review this. The comments I posted are the ones that I think needs adressing before merging, but I'll post the entire review here for completeness. 

# Claude review

## Summary

The channel library is a clean piece of work and the `Decoder` refactor is *almost* behaviour-preserving — but not quite. There's one real regression on the decode-error path, and it stems from a design point worth fixing properly rather than patching.

---

## 1. Blocking: the decode-error path no longer signals EOF

Before:

```cpp
} catch (...) {
    eof = true;                      // <-- removed
    CASPAR_LOG_CURRENT_EXCEPTION();
}
```

That `eof = true` made `pop()` start returning `alloc_frame()` — an empty frame with `data[0] == nullptr`, which is exactly the sentinel `AVProducer::schedule()` looks for at lines 1154/1164 to call `av_buffersrc_close()` and erase the source. So a decode error ended the clip cleanly (and looped, if configured).

After the change, nothing is pushed on that path, and `output.try_receive().closed` can only become true once `output_sender` is destroyed. My simulation of both paths:

```
=== A. normal AVERROR_EOF path ===
    pop() #0 -> real frame                                     want_packet()=1
    pop() #1 -> EOF SENTINEL (data[0]==null -> buffersrc_close) want_packet()=1
    pop() #2 -> nullptr  (schedule() does `continue`)           want_packet()=1

=== B. decode-error path (catch(...)) ===
    [decode thread] exception logged, thread exits
    pop() #0 -> real frame                                     want_packet()=1
    pop() #1 -> nullptr  (schedule() does `continue`)           want_packet()=1
    pop() #2 -> nullptr  (schedule() does `continue`)           want_packet()=1
```

Path A survives only because the EOF branch explicitly pushes a sentinel before `break` — and `schedule()` needs it just once. Path B produces no sentinel ever, so `schedule()` hits `if (!frame) continue;` forever: the source is never closed, never erased, and the filter starves. A corrupt stream now stalls where it used to terminate.

`FF()` / `FF_RET(ret, "avcodec_receive_frame")` throw on ordinary decode errors, so this isn't a theoretical path.

## 2. Root cause: `closed()` tracks callable teardown, not loop exit

I assumed Boost destroys a thread's callable when the thread function returns. It doesn't — it destroys it at `join()`/detach:

```
AFTER fn returned, BEFORE join: live=1
  Probe dtor  live=0
AFTER join:                     live=0
```

Since `~Decoder()` is what joins, `output_sender` and `input_receiver` stay alive for the entire life of the `Decoder`. Consequences:

- The `closed` branch of `pop()` is effectively dead code in this integration.
- `want_packet()` keeps returning `true` after the decode thread has finished (old code returned `false` the moment `eof` was set). It's bounded — `full()` trips at 2 — so it's not a leak, but `AVProducer::want_packet()` keeps advertising interest in a dead decoder and `schedule()` keeps handing it packets nobody will read.

Rather than just pushing a sentinel in `catch (...)`, I'd suggest giving `Sender`/`Receiver` an explicit `close()` and calling it before each `break` and in the catch. That restores the promptness of the old `eof` flag, keeps the intent local to the decode loop, and stops correctness from depending on Boost's teardown timing.

## 3. Worth fixing while you're in here

**`avcodec_flush_buffers()` before the `break` is now dead code.** It existed so the loop could continue; it can't now. Remove it, or the next reader will assume the decoder is reusable.

**The send/check order reads backwards:**

```cpp
FF(avcodec_send_packet(ctx.get(), packet_opt.value_or(nullptr).get()));
if (!packet_opt) break;
```

On a closed channel this sends the drain NULL and then breaks without draining, discarding whatever is still buffered. It's unreachable today (`~Decoder` joins before `input` is destroyed), but it reads like a bug — either `break` first or comment why.

Worth a comment for the opposite reason too: the real drain signal from `schedule()` (`p.second.push(nullptr)`, line 1120) arrives as an **engaged** optional holding a null `shared_ptr`. I verified this — `has_value=1, ptr=0` — so it correctly sends NULL to avcodec and does *not* break. That's right, but `if (!packet_opt)` sitting right there looks like it should catch that case.

**`push()`'s non-blocking-ness hangs on `channel_capacity_is_hint`, ~90 lines away.** Verified: with the hint, 100 000 sends into a capacity-2 channel took 3 ms and never blocked, while `full()` still trips at 2 — exactly reproducing the old `push()`/`want_packet()` pair. But `schedule()` runs on the producer's executor thread, and line 1125 literally says `// TODO (fix): limit it->second.input.size()?`. If someone acts on that TODO and drops the hint, `push()` becomes blocking and deadlocks the producer. Please put a comment on `push()` saying the hint is load-bearing.

**Gratuitous copies:** `push()` takes `packet` by value then passes it as an lvalue; likewise `output_sender.try_send_blocking(av_frame)` in the per-frame path. The original used `std::move`. One atomic refcount each — trivial, but a needless diff.

**Release builds pay for debug names.** Both `boost::format(...)` strings are built unconditionally per `Decoder`, then discarded when `CASPARCG_DEBUG_CHANNEL` is off — and decoders are recreated on every seek/loop via `decoders_.clear()`. Guard the call sites, or have `channel()` take a callable.

## 4. channel.h nits

- The `cond_.wait` predicates in `try_send_blocking` / `try_receive_blocking` perform the actual transfer as a side effect, and call `notify_all()` from inside `wait`. It's correct (I tested it) because the operations are idempotent on failure — but predicates are invoked an unspecified number of times including spuriously, so this is the single thing most likely to get broken later. It deserves a comment.
- `full()` doesn't mean "a send would block" on a hint channel. Worth documenting; it's the one name that misleads.
- `debug_name` is unused when the macro is off → `-Wunused-parameter` under `-Wextra` (I hit it). MSVC's C4100 is suppressed via `disable_silly_warnings.h` so `/WX` won't break, but `[[maybe_unused]]` is free.
- `if (is_sender)` → `if constexpr`.
- `channel()` does `new State` then builds two `unique_ptr`s from the same raw pointer. Safe today since everything on that path is `noexcept`; one added throwing operation makes it a double-free.
- `TrySendResult` is declared in `Receiver`'s private section and re-exported publicly from `Sender` — it belongs in `Sender`.
- Missing newline at end of file; `#include <utility>` for `std::pair`.
- `[=, this]` is fine (the project sets `CXX_STANDARD 20`), but the `=` captures nothing the body uses — `[this, input_receiver = ..., output_sender = ...]` is clearer.

## 5. What's good

- Using Boost's mutex/condvar for interruptibility is the right call, and I confirmed `try_receive_blocking` is a clean interruption point.
- The dual-ownership `unique_ptr` + `Deleter` scheme is sound. I verified that a sender blocked in `try_send_blocking` is woken when the receiver dies and gets its value handed back rather than dropped.
- The thread now exits at EOF instead of parking forever in `input_cond.wait` — a genuine improvement over the old code.
- Deleted copy ops, and returning unsent values instead of silently dropping them, are both the right instincts.

Given this is infrastructure split out of #1637 and meant to grow, a small unit test for close semantics, the hint behaviour, and blocked-sender wakeup would pay for itself fast — I wrote roughly that harness to review this, and it caught the issue in §1.

One question for the author: was looping/seeking playback tested, and was a deliberately corrupt file tried? §1 only shows up in the second case.

**Sources:** [[CasparCG/server PR #1791](https://github.com/CasparCG/server/pull/1791)](https://github.com/CasparCG/server/pull/1791)

programmerjake:
force pushed for the changes from #1792, nothing fixed here yet.

PR #1792, opened 2026-09-11
Add new APIs for creating `[a]buffersink`s and switch to using them

programmerjake:
The existing code does a bunch of creating buffersinks in different spots and that makes it easy for them to become inconsistent, this PR creates one common API that can be easily used by the rest of the project.

This is split out from #1637

### DIFF
--- a/src/modules/ffmpeg/producer/av_producer.cpp
+++ b/src/modules/ffmpeg/producer/av_producer.cpp
@@ -45,7 +45,7 @@ extern "C" {
 #include <deque>
 #include <iomanip>
 #include <memory>
-#include <queue>
+#include <optional>
 #include <sstream>
 #include <string>
 #include <thread>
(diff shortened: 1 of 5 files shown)


src/modules/ffmpeg/producer/av_producer.cpp
```diff
@@ -45,7 +45,7 @@ extern "C" {
 #include <deque>
 #include <iomanip>
 #include <memory>
-#include <queue>
```
niklaspandersson, line 48: std::queue is still used in this file

src/modules/ffmpeg/util/av_util.h
```diff
@@ -48,4 +58,77 @@ std::map<std::string, std::string> to_map(AVDictionary** dict);
...
+    constexpr av_opt_array_ref(const T* data, std::size_t size) noexcept
+        : data_(size == 0 ? nullptr : data)
+        , size_(size)
+    {
+    }
+    constexpr av_opt_array_ref(std::initializer_list<T> data) noexcept
+        : av_opt_array_ref(data.begin(), data.size())
+    {
+    }
+    template <typename = std::enable_if<std::is_integral_v<T> || std::is_enum_v<T>>>
```
niklaspandersson: The std::enable_if condition doesn't work. i think you mean std::enable_if_t.  Overload currently works anyway because of regular overload resolution. 
Same below

src/modules/ffmpeg/util/av_util.cpp
```diff
@@ -414,4 +417,163 @@ uint64_t get_channel_layout_mask_for_channels(int channel_count)
     return channel_layout;
 }
 
+#if LIBAVFILTER_VERSION_INT < AV_VERSION_INT(10, 6, 0)
```
niklaspandersson, line 440: I'm not too fond of these C-style defines. I think they're difficult to read, they mangle source code lines numbers, and make it impossible to set breakpoints in them.

Claude helped me mock a cpp-alternative using a template function and traits which I think is much easier to read but still adhere to the same "philosophy":

[av_util_attributes.cpp](https://github.com/user-attachments/files/32464520/av_util_attributes.cpp)

niklaspandersson:
Good abstractions. Everything that that makes the code easier to read and reason about and also reduced code duplication is a win in my book.

looks good to me, besides the comments above.

dimitry-ishenko:
IMHO `std::optional` is not necessary, since `av_opt_array_ref` already provides empty state.

programmerjake:
> IMHO `std::optional` is not necessary, since `av_opt_array_ref` already provides empty state.

I disagree, empty is not the same thing as default (what you get if you never set some property). `nullopt` allows you to retain the default state.

src/modules/ffmpeg/producer/av_producer.cpp
programmerjake: ah, I missed that. I removed it because `clangd` said it didn't provide any symbols (probably because `<queue>` was already included by a previous header that is included).

src/modules/ffmpeg/util/av_util.cpp
programmerjake: yeah, I will rewrite it to use templates, though I will be writing my own since I do not want to submit changes written using AI.

dimitry-ishenko:
> > IMHO `std::optional` is not necessary, since `av_opt_array_ref` already provides empty state.
> 
> I disagree, empty is not the same thing as default (what you get if you never set some property). `nullopt` allows you to retain the default state.

@programmerjake You are right of course. In general sense they mean two different things. But, in the context of `av_opt_set_array()` they do the same thing. From the `av_opt_set_array()` documentation:

> **nb_elems**	number of array elements to modify;

So, passing an empty array means this function is a no-op.

I am not saying you should pass empty arrays to it -- that would be a waste of electrons. I am saying `std::optional` is unnecessary, and in fact is a bit misleading as it makes it seem that these two calls have different results:

```c++
auto x = create_buffersink(foo, bar, std::nullopt);
auto y = create_buffersink(foo, bar, av_opt_array_ref<AVPixelFormat>{});
```
when the only difference is that the latter makes an unnecessary call to `av_opt_set_array()` which does nothing.

Instead, I would suggest to just pass straight arrays and check:

```c++
if (pixel_formats.size()) {
    set_pixel_formats(retval, *pixel_formats);
}
```

You could also add a default value, so that if someone wants "defaults" they can just omit it, eg:
```c++
AVFilterContext*
create_buffersink(AVFilterGraph*, const char*, av_opt_array_ref<AVPixelFormat> = {});
```

src/modules/ffmpeg/util/av_util.h
programmerjake: Fixed.

src/modules/ffmpeg/util/av_util.cpp
programmerjake: Rewritten to use templates.

programmerjake:
> Instead, I would suggest to just pass straight arrays and check

changed to use empty arrays to mean don't change anything, except on old ffmpeg where we set the `all_channel_counts` flag when passed an empty array.

src/modules/ffmpeg/producer/av_producer.cpp
programmerjake: Readded the `<queue>` include, turns out I removed it because later changes in #1637 remove all uses of `std::queue`, but the removed include accidentally ended up here when rebasing.

dimitry-ishenko:
I took a closer look at it. There is a much simpler way of doing it:
1. `av_opt_array_ref` is essentially an `std::span` with a few convenience constructors.
2. There no need for macros or tags.

Below is more or less complete sketch. Notice, there are no macros or tags anywhere. Just a simple linear code that's easy read for lazy humans. The only "inconvenient" part about `std::span` is that you can't pass initializer lists directly to functions. You either need to do this:
```c++
sink = create_buffersink(graph.get(), "out", std::array{pix_fmt});
```
or this:
```c++
const AVPixelFormat pix_fmts[] = {AV_PIX_FMT_RGB24,
                                  ...
                                  AV_PIX_FMT_GBRAP16};
sink = create_buffersink(graph.get(), "out", pix_fmts);
```
Unfortunately, we will have to wait until c++26 to be able to pass `{pix_fmt}` directly to `std::span`.

---

**av_util.h:**
```c++
std::span<const AVPixelFormat> get_supported_pixel_formats(const AVCodecContext*, const AVCodec*);
void set_pixel_formats(AVFilterContext*, std::span<const AVPixelFormat>);

std::span<const AVSampleFormat> get_supported_sample_formats(const AVCodecContext*, const AVCodec*);
void set_sample_formats(AVFilterContext*, std::span<const AVSampleFormat>);

std::span<const int> get_supported_sample_rates(const AVCodecContext*, const AVCodec*);
void set_sample_rates(AVFilterContext*, std::span<const int>);

std::span<const AVChannelLayout> get_supported_channel_layouts(const AVCodecContext*, const AVCodec*);
void set_channel_layouts(AVFilterContext*, std::span<const AVChannelLayout>);
AVChannelLayout get_channel_layout_default(int nb_channels);

AVFilterContext* create_buffersink(AVFilterGraph*, const char* name, std::span<const AVPixelFormat> = {});

AVFilterContext* create_abuffersink(AVFilterGraph*, const char* name,
    std::span<const AVSampleFormat> = {}, std::span<const int> = {}, std::span<const AVChannelLayout> = {}
);
```

**av_util.cpp:**
```c++
#if LIBAVFILTER_VERSION_INT >= AV_VERSION_INT(10, 6, 0)

namespace {

template<typename T>
auto get_supported(const AVCodecContext* avctx, const AVCodec* codec, AVCodecConfig config)
{
    const void* data = nullptr;
    int size = 0;
    FF(avcodec_get_supported_config(avctx, codec, config, 0, &data, &size));
    return std::span{static_cast<const T*>(data), static_cast<std::size_t>(size)};
}

template<typename T>
void set_array(AVFilterContext* target, const char* name, AVOptionType type, std::span<const T> values)
{
    FF(av_opt_set_array(target, name, AV_OPT_SEARCH_CHILDREN | AV_OPT_ARRAY_REPLACE,
                        0, static_cast<unsigned>(values.size()), type, values.data()));
}

}

std::span<const AVPixelFormat> get_supported_pixel_formats(const AVCodecContext* avctx, const AVCodec* codec)
{
    return get_supported<const AVPixelFormat>(avctx, codec, AV_CODEC_CONFIG_PIX_FORMAT);
}
void set_pixel_formats(AVFilterContext* target, std::span<const AVPixelFormat> values)
{
    set_array(target, "pixel_formats", AV_OPT_TYPE_PIXEL_FMT, values);
}

std::span<const AVSampleFormat> get_supported_sample_formats(const AVCodecContext* avctx, const AVCodec* codec)
{
    return get_supported<AVSampleFormat>(avctx, codec, AV_CODEC_CONFIG_SAMPLE_FORMAT);
}
void set_sample_formats(AVFilterContext* target, std::span<const AVSampleFormat> values)
{
    set_array(target, "sample_formats", AV_OPT_TYPE_SAMPLE_FMT, values);
}

std::span<const int> get_supported_sample_rates(const AVCodecContext* avctx, const AVCodec* codec)
{
    return get_supported<int>(avctx, codec, AV_CODEC_CONFIG_SAMPLE_RATE);
}
void set_sample_rates(AVFilterContext* target, std::span<const int> values)
{
    set_array(target, "samplerates", AV_OPT_TYPE_INT, values);
}

std::span<const AVChannelLayout> get_supported_channel_layouts(const AVCodecContext* avctx, const AVCodec* codec)
{
    return get_supported<AVChannelLayout>(avctx, codec, AV_CODEC_CONFIG_CHANNEL_LAYOUT);
}
void set_channel_layouts(AVFilterContext* target, std::span<const AVChannelLayout> values)
{
    set_array(target, "channel_layouts", AV_OPT_TYPE_CHLAYOUT, values);
}

#else

namespace {

const AVCodec* resolve_codec(const AVCodecContext* avctx, const AVCodec* codec, AVMediaType expected,
    const std::source_location& location = std::source_location::current())
{
    if (!codec)
        codec = avctx->codec;
    if (codec->type != expected)
        FF_RET(AVERROR(EINVAL), location.function_name());
    return codec;
}

template <typename T>
void set_bin(AVFilterContext* target, const char* name, std::span<const T> values)
{
    FF(av_opt_set_bin(target, name, reinterpret_cast<const uint8_t*>(values.data()),
                      static_cast<int>(values.size_bytes()), AV_OPT_SEARCH_CHILDREN));
}

template <typename T>
constexpr std::span<const T> span_until(const T* data, T terminator) noexcept
{
    std::size_t size = 0;
    while (data[size] != terminator) ++size;
    return { data, size };
}

template <typename T, std::predicate<const T&> Fn>
constexpr std::span<const T> span_until_if(const T* data, Fn is_terminator)
{
    if (!data) return {};
    std::size_t size = 0;
    while (!is_terminator(data[size])) ++size;
    return { data, size };
}

}

std::span<const AVPixelFormat> get_supported_pixel_formats(const AVCodecContext* avctx, const AVCodec* codec)
{
    codec = resolve_codec(avctx, codec, AVMEDIA_TYPE_VIDEO);
    return span_until(codec->pix_fmts, AV_PIX_FMT_NONE);
}
void set_pixel_formats(AVFilterContext* target, std::span<const AVPixelFormat> values)
{
    set_bin(target, "pix_fmts", values);
}

std::span<const AVSampleFormat> get_supported_sample_formats(const AVCodecContext* avctx, const AVCodec* codec)
{
    codec = resolve_codec(avctx, codec, AVMEDIA_TYPE_AUDIO);
    return span_until(codec->sample_fmts, AV_SAMPLE_FMT_NONE);
}
void set_sample_formats(AVFilterContext* target, std::span<const AVSampleFormat> values)
{
    set_bin(target, "sample_fmts", values);
}

std::span<const int> get_supported_sample_rates(const AVCodecContext* avctx, const AVCodec* codec)
{
    codec = resolve_codec(avctx, codec, AVMEDIA_TYPE_AUDIO);
    return span_until(codec->supported_samplerates, 0);
}
void set_sample_rates(AVFilterContext* target, std::span<const int> values)
{
    set_bin(target, "sample_rates", values);
}

std::span<const AVChannelLayout> get_supported_channel_layouts(const AVCodecContext* avctx, const AVCodec* codec)
{
    codec = resolve_codec(avctx, codec, AVMEDIA_TYPE_AUDIO);
    return span_until_if(codec->ch_layouts, [](auto&& l) { return 0 == av_channel_layout_check(&l); });
}
void set_channel_layouts(AVFilterContext* avctx, std::span<const AVChannelLayout> values)
{
    // TODO
}

#endif

AVChannelLayout get_channel_layout_default(int nb_channels)
{
    AVChannelLayout retval{};
    av_channel_layout_default(&retval, nb_channels);
    return retval;
}

AVFilterContext*
create_buffersink(AVFilterGraph* graph, const char* name, std::span<const AVPixelFormat> pixel_formats)
{
    AVFilterContext* retval = FFMEM(avfilter_graph_alloc_filter(graph, avfilter_get_by_name("buffersink"), name));
    if (!pixel_formats.empty()) {
        set_pixel_formats(retval, pixel_formats);
    }
    FF(avfilter_init_str(retval, nullptr));
    return retval;
}

AVFilterContext*
create_abuffersink(AVFilterGraph* graph, const char* name, std::span<const AVSampleFormat> sample_formats,
    std::span<const int> sample_rates, std::span<const AVChannelLayout> channel_layouts)
{
    AVFilterContext* retval = FFMEM(avfilter_graph_alloc_filter(graph, avfilter_get_by_name("abuffersink"), name));
    if (!sample_formats.empty()) {
        set_sample_formats(retval, sample_formats);
    }
    if (!sample_rates.empty()) {
        set_sample_rates(retval, sample_rates);
    }
    if (!channel_layouts.empty()) {
        set_channel_layouts(retval, channel_layouts);
    }
#if LIBAVFILTER_VERSION_INT < AV_VERSION_INT(10, 6, 0)
    else {
        FF(av_opt_set_int(retval, "all_channel_counts", 1, AV_OPT_SEARCH_CHILDREN));
    }
#endif
    FF(avfilter_init_str(retval, nullptr));
    return retval;
}
```

dimitry-ishenko:
@programmerjake you can pull from here if you want:

https://github.com/dimitry-ishenko-casparcg/server/tree/buffersink-api

It compiles fine, but I haven't tested anything.

PR #1795, opened 2026-09-11
fix(ffmpeg): guard all_channel_counts option removed in FFmpeg 8

### LINKED ISSUE #1794: FFmpeg 8+: audio producer crashes for every media file with an audio track (all_channel_counts option removed)
## Summary

On FFmpeg 8+ (confirmed with FFmpeg 9.0.1), `casparcg-server-git` (built against system FFmpeg, tested on Arch Linux, revision `13cf17e6a`) crashes its audio producer for **every** media file that contains an audio track — video+audio files and audio-only files alike. Video-only and image files are unaffected.

## Root cause

`av_producer.cpp` sets the abuffersink option `all_channel_counts` unconditionally:

https://github.com/CasparCG/server/blob/13cf17e6a/src/modules/ffmpeg/producer/av_producer.cpp#L595

The surrounding code already has a version guard (`LIBAVUTIL_VERSION_MAJOR >= 60`) for the `sample_formats`/`samplerates` options, added because FFmpeg 8 changed that part of the `AVFilterContext` options API — but the `all_channel_counts` call itself was left outside/before that guard.

`all_channel_counts` was removed from `abuffersink` entirely in FFmpeg 8. Confirmed directly against FFmpeg's own source (`release/9.0` branch): `abuffersink_options[]` in [`libavfilter/buffersink.c`](https://github.com/FFmpeg/FFmpeg/blob/release/9.0/libavfilter/buffersink.c) only lists `sample_formats`, `samplerates`, `channel_layouts` — no `all_channel_counts`.

## Observed error

```
[error]   Exception: .../modules/ffmpeg/producer/av_producer.cpp(595): Throw in function caspar::ffmpeg::Filter::Filter(...)
[error]   Dynamic exception type: boost::wrapexcept<caspar::ffmpeg::ffmpeg_error_t>
[error]   [boost::errinfo_api_function_*] = av_opt_set_int(sink, "all_channel_counts", 1, (1 << 0))
```

`PLAY` returns `202 PLAY OK` (the AMCP ack happens before the producer is actually constructed), but the layer then throws immediately afterward and stays silent/empty.

## Reproduction

1. Build/run casparcg-server against FFmpeg 8+ (e.g. Arch Linux's current `ffmpeg` package, 9.0.1).
2. Any media file with an audio track, e.g. a plain WAV:
   ```
   ffmpeg -f lavfi -i "sine=frequency=1000:duration=10" -ac 2 -ar 48000 -c:a pcm_s16le test.wav
   ```
3. `PLAY 1-1 test` (or any video file that has an audio stream).
4. Server log shows the exception above immediately after `PLAY OK`; the layer never produces audio (or any output, for the audio-only case).

## Fix

PR incoming: version-guard the `all_channel_counts` call the same way the adjacent `sample_formats`/`samplerates` calls already are, so it's only set on FFmpeg < 8 where the option still exists.

## Environment

- CasparCG Server: `2.6.0 Dev`, revision `13cf17e6a` (`casparcg-server-git` r4222.13cf17e6a, Arch Linux AUR)
- FFmpeg: 9.0.1 (Arch Linux `extra/ffmpeg`)
- OS: Arch Linux, x86_64

florianguenther:
Fixes #1794

FFmpeg 8 removed the `abuffersink` AVOption `all_channel_counts` (confirmed against FFmpeg's `release/9.0` branch — `abuffersink_options[]` in `libavfilter/buffersink.c` only lists `sample_formats`, `samplerates`, `channel_layouts`).

`av_producer.cpp` already version-gates the adjacent `sample_formats`/`samplerates` option calls for this exact FFmpeg 8 API change (`LIBAVUTIL_VERSION_MAJOR >= 60`), but the `all_channel_counts` call right above them was left unguarded. On FFmpeg 8+ that call now fails ("Option not found"), which throws inside `Filter::Filter()` and aborts the audio producer for **every** media file with an audio track (video+audio and audio-only alike) — see #1794 for repro, log output and environment details.

## Fix

Wrap the `all_channel_counts` call in the same version guard, so it's only set on the older FFmpeg API where the option still exists:

```cpp
#if LIBAVUTIL_VERSION_MAJOR < 60 // "all_channel_counts" was removed from abuffersink in FFmpeg 8
            FF(av_opt_set_int(sink, "all_channel_counts", 1, AV_OPT_SEARCH_CHILDREN));
#endif
```

## Testing

Built and tested locally against FFmpeg 9.0.1 (Arch Linux): reproduced the crash before the fix (audio-only WAV and video+audio files both threw immediately after `PLAY OK`), confirmed clean playback with audio after applying this patch and rebuilding.


Julusian:
Are you sure that 8 is the threshold for this? Both main & 2.5.x (soon to be 2.5.1) are running 8.1.2 on windows and there havent been issues with this, and running locally on fedora (also 8.1.2) it is also happy.

I can't say whether that property is doing anything though, but it doesnt appear to be a problem to set it.

So perhaps this should be retargeted to only apply when ffmpeg >=9?

PR #1800, opened 2026-09-18
feat(ffmpeg): decode-alpha-mode, so straight-alpha media can composite correctly

C-alcaide:
`make_frame`'s `is_straight_alpha` decides whether the shader runs `col.rgb *= col.a` before the blend, which is premultiplied source-over. The ffmpeg producer never set it, so straight-alpha media has its colour added to what is behind it — an opaque fill where the clip is transparent. `image_producer.cpp:60` already passes `true`, so the two producers disagree about identical content.

Nothing in ProRes 4444 or QuickTime Animation declares the mode, and both populations are real: After Effects and CasparCG's own recordings are premultiplied, Premiere and Media Encoder are straight. Detection isn't safe either — a dark graphic satisfies `rgb <= alpha` whichever convention it is. So it's declared, in the order **caller > config > default**:

```xml
<ffmpeg><producer>
    <decode-alpha-mode>default [default|straight|premultiplied]</decode-alpha-mode>
```

```
PLAY 1-1 "clip" STRAIGHT
PLAY 1-1 "clip" PREMULTIPLIED
```

**The default is premultiplied — what the server does today — so omitting the line changes nothing.** An unrecognised config value warns and uses premultiplied.

## Verified

Straight-alpha ProRes 4444, left half opaque colour, right half white at alpha 0, over an opaque `(32, 96, 160)` background. The right half is the verdict — 223 codes apart on blue, so the arms can't be confused. One fresh server per arm.

| arm | transparent half | |
| :--- | :--- | :--- |
| no config line | `[255 255 255]` | as today |
| `default` | `[255 255 255]` | as today |
| `premultiplied` | `[255 255 255]` | as today |
| `bananas` (bad value) | `[255 255 255]` | as today, with a warning |
| `straight` | `[ 32  96 160]` | the background |
| `premultiplied` + `STRAIGHT` | `[ 32  96 160]` | keyword wins |
| `straight` + `PREMULTIPLIED` | `[255 255 255]` | keyword wins |

7/7.

Repro fixture, if useful:

```sh
ffmpeg -f lavfi -i "color=c=0xC83CF0:s=1920x1080:r=25:d=2,drawbox=x=960:y=0:w=960:h=1080:c=white:t=fill,format=gbrp" \
       -f lavfi -i "color=c=white:s=1920x1080:r=25:d=2,drawbox=x=960:y=0:w=960:h=1080:c=black:t=fill,format=gray" \
       -filter_complex "[0][1]alphamerge[v]" -map "[v]" \
       -c:v prores_ks -profile:v 4444 -pix_fmt yuva444p10le straight.mov
```

## Notes

`spl::make_shared` forwards at most twelve arguments and the constructor now takes thirteen, hence `spl::make_shared_ptr(std::make_shared<>)` rather than another overload in `common/memory.h`. Happy to change it.

Scoped to the ffmpeg producer. Two things deliberately left out: FFmpeg 8's `AVFrame.alpha_mode`, which would let PNG/EXR/JXL/Matroska declare their own mode instead of falling to the default; and the ffmpeg **consumer**, which writes the mixer's premultiplied output as-is while the IMAGE consumer straightens — so a qtrle recording played back needs `PREMULTIPLIED`. Both are separate changes and I'd rather agree the shape here first.


Julusian:
For some reason I thought I had already done this, but perhaps that was just why the image producer has straight alpha enforced like that..

> The default is premultiplied — what the server does today — so omitting the line changes nothing. An unrecognised config value warns and uses premultiplied.

I feel that it would be best if the default was straight alpha, with this config flag existing as the opt-out to restore previous behaviour.  
Intepreting as straight alpha is technically correct, and if we start respecting the `alpha_mode` property on frames, then this could end up being inconsistent anyway.



>  Two things deliberately left out: FFmpeg 8's AVFrame.alpha_mode

This would be good to tackle at the same time. Both respecting that and this default will mess with how files are interpreted so it would be better to do this all in one go rather than smaller unclear steps.

Especially as I am not 100% sure how these should all interact. The AMCP parameters should win, then the ffmpeg flag. But should the config default override ffmpeg or be the fallback for unspecified?  
Maybe the config should have values to ignore what ffmpeg claims, and values to stand in for unspecified?

> and the ffmpeg consumer

I am happy for this to be a follow up. Question is whether it is good enough to simply set `alpha_mode` or do we need to tell it to straighten? It sounds like the requirements will vary per codec, so this will need some thought and testing.

C-alcaide:
Yeah I think we should try to get some other people input into this to consider how it can affect them.

The reason I scratched a previous idea of just changing it 'all' to straight is that doing that means the actual premultiplied files coming from After Effects will be decoded as straight and render darker than they should on areas with partial alpha (0.5 alpha is the most visible case, it loses around 1/4 brightness). So basically I thought such a breaking change might not be welcome or worth it.
If we look around: Resolume also defaults to premultiplied and allows changing it by file.
I think Disguise goes the other way and defaults to straight, but also lets you change it.

There's probably good arguments to go either way, and since there's no way to reliably know if it's straight or premultiplied for ProRes and QTRLE (the most common ones I assume) we do have to choose a default and leave the other opt-in. I thought maintaining current default was the best option at least for now.

I do agree that for the (not very common) cases that the files actually expose their alpha mode, the config line and flag on command should probably be ignored, since it's forcing an incorrect decoding for no good reason.

 For now I will add ffmpeg reading the flag for the codecs that expose it.

C-alcaide:
I checked which formats actually declare anything, since that decides how much the two changes really interact.

| file | producer | declaration |
| :--- | :--- | :--- |
| `.png` still | image | n/a — image producer already forces straight |
| png inside a container | ffmpeg | straight |
| `.exr` | ffmpeg | premultiplied |
| `.jxl` | ffmpeg | straight (a real per-file flag) |
| `.webm` VP9+alpha | ffmpeg | **stream metadata**; the frame says `unspecified` |
| `.mov` ProRes 4444 / qtrle | ffmpeg | nothing |

Only `pngdec.c`, `exr.c` and `libjxldec.c` set `AVFrame.alpha_mode` in FFmpeg 8, and the first two just state their format's fixed convention — only jpegxl reads a per-file flag.

So respecting it is correct but nearly inert: it changes png-in-a-container and jxl, and nothing at all for ProRes 4444, QuickTime RLE, Hap or NotchLC — which is all the content the default actually governs. I think that mostly answers the interaction worry: the overlap is three rare formats, not something pervasive.

Two things I hit implementing it:

**WebM goes a different way.** `matroskadec.c` copies AlphaMode to *stream metadata*, not to `AVFrame.alpha_mode`, so reading the frame property does nothing for VP9+alpha. That's probably the most worthwhile declaration to honour — unambiguous, and an actual delivery format — but it needs its own code.

**The filter graph drops it.** Reading `frame->alpha_mode` where we build the frame silently does nothing, because the buffersrc link is configured without it and the sink then hands back `unspecified`. ffmpeg says so itself:

```
filter context - ... alpha: unspecified, incoming frame - ... alpha: straight
```

Making it work needs the decoder's value latched pre-filter and set on `AVBufferSrcParameters::alpha_mode`. There's no `av_buffersrc_parameters_set` call today, so it's new plumbing rather than a few lines — which is the main reason I'd originally kept it separate.

On precedence, concretely:

```
<decode-alpha-mode>default|straight|premultiplied|force-straight|force-premultiplied</decode-alpha-mode>
```

The first three say what to assume when the file is silent, and a declaring file still wins over them; `force-*` ignores the declaration too; `PLAY ... STRAIGHT|PREMULTIPLIED` always wins. I have this working and can push it.

The open question is whether `force-*` should exist at all. It covers a file that declares *wrongly*, but if we'd rather a declaration always win I'm happy to drop it — that's closer to what I said above.

PR #1801, opened 2026-09-20
fix(windows): rename versioned media-scanner exe to scanner.exe during packaging

### LINKED ISSUE #1798: issue: v2.5.1 Stable
### Observed Behavior

Hi, marvelous work with 251 kudos to all!

A small ting. I know that some uses the casparcg_auto_restart.bat now it points to a scanner that is named different. scanner.exe not casparcg-scanner-v1.4.0-win32-x64.exe
Not a big problem, but anyway.

if everything is ok, CasparCG 2.5.1 will do the Eurovision Asia Finals in Bankok, November 14.

### Expected behaviour

Start a scanner ;)

### Steps to reproduce

1.
2.
3. ...

### Environment

* Commit: [e.g. ab1234c]
* Server version: [e.g. v2.2]
* Operating system: [e.g. Windows 10]

dedicatedbroadcastsolutions:
## Problem
`casparcg_auto_restart.bat` only checks for and starts `scanner.exe`. The media-scanner release zip (e.g. `casparcg-scanner-v1.4.0-win32-x64.zip`) ships the executable with a versioned filename (`casparcg-scanner-v1.4.0-win32-x64.exe`), which never matches, so the scanner is silently never started when using the auto-restart batch script.

Confirmed by inspecting the actual v1.4.0 Windows release asset — it contains a single file named `casparcg-scanner-v1.4.0-win32-x64.exe`.

## Fix
After `package.bat` copies the media-scanner folder into the server output folder, rename the versioned exe to `scanner.exe` so it matches what `casparcg_auto_restart.bat` expects, regardless of the scanner's release version.

Fixes #1798


dedicatedbroadcastsolutions:
Alternatively we could have some sort of version aware packaging script, but this would get a bit complicated.

dimitry-ishenko:
This slightly OT, but I am working on another solution.

I am re-writing the scanner in python and my idea was to run it as a service. On Linux it can be a user or a system service. On windows, I am planning to use nssm.

It will be completely decoupled from CasparCG Server and will run on its own. Let me know if you are interested.

dedicatedbroadcastsolutions:
This is just a response to fix [issue 1798](https://github.com/CasparCG/server/issues/1798).  I have already made my own replacement for scanner in typescript/react with a UI for broadcast conductor and do not rely on scanner at all for my app at this point except for a legacy server that will be updated soon.  

My implementation is more of a library management program though that allows for user input in what to scan so individual folders can be rescanned.  If there is an interest, I may be able to open source a portion of it if I can get some grants to cover the cost of splitting it out.

Issue #94, opened 2013-04-03
Uptime monitor counter


hummelstrand:
Being able to get a timestamp of the local time the Server was started, and an uptime dd:hh:mm:ss counter since last start (not total running time) via AMCP would be useful.

PR #1805, opened 2026-09-21
feat(amcp,osc): add server uptime via INFO SERVER and OSC

dedicatedbroadcastsolutions:
## Problem
Issue #94 asks for a way to get the server's start timestamp and an uptime counter (dd:hh:mm:ss since last start) via AMCP, with OSC suggested as a longer-term option. Neither exists today - `INFO`/`INFO CONFIG`/`INFO PATHS` only cover channel state, config, and paths, and no OSC path carries anything like it.

## Fix
- `env::start_time()` ([common/env.h](../../src/common/env.h)/[env.cpp](../../src/common/env.cpp)) records the process start time once, at static initialization.
- New AMCP query command `INFO SERVER`, alongside the existing `INFO CONFIG`/`INFO PATHS`, returns:
  ```xml
  <server>
    <start-time>2026-09-21T12:00:00</start-time>
    <uptime>00:03:12:45</uptime>
  </server>
  ```
- Each channel's regular OSC tick now also includes a top-level `uptime` value (seconds since start, as a float) alongside the existing per-channel state, so monitoring/automation clients can track it continuously via OSC subscription instead of having to poll AMCP.

## Testing
Verified the change compiles with no diagnostics.


src/protocol/amcp/AMCPCommandsImpl.cpp
```diff
@@ -1946,6 +1948,38 @@ std::wstring info_paths_command(command_context& ctx)
...
+
+    auto uptime_secs = duration_cast<seconds>(system_clock::now() - caspar::env::start_time()).count();
+    auto days        = uptime_secs / 86400;
+    auto hours       = (uptime_secs % 86400) / 3600;
+    auto minutes     = (uptime_secs % 3600) / 60;
+    auto seconds_rem = uptime_secs % 60;
+
+    std::wstringstream uptime;
+    uptime << std::setfill(L'0') << std::setw(2) << days << L":" << std::setw(2) << hours << L":" << std::setw(2)
+           << minutes << L":" << std::setw(2) << seconds_rem;
```
dimitry-ishenko: we are on c++20 now and this mess can be replaced with simple:
```c++
auto uptime = duration_cast<seconds>(system_clock::now() - caspar::env::start_time());
auto d = duration_cast<days>(uptime);
auto uptime = std::format(L"{:02}:{:%T}", d.count(), uptime - d);
```

src/protocol/amcp/AMCPCommandsImpl.cpp
```diff
@@ -1946,6 +1948,38 @@ std::wstring info_paths_command(command_context& ctx)
...
+
+    auto start_time_str =
+        u16(boost::posix_time::to_iso_extended_string(boost::posix_time::from_time_t(system_clock::to_time_t(caspar::env::start_time()))));
```
dimitry-ishenko: no need to use boost here:
```c++
auto start_time_str = std::format(L"{:%FT%T}", caspar::env::start_time());
```

src/protocol/amcp/AMCPCommandsImpl.cpp
dedicatedbroadcastsolutions: Hmm, looks like I need to review a bunch of my other recent work if that got through like that... Thanks for doing the slop patrol.

src/protocol/amcp/AMCPCommandsImpl.cpp
dedicatedbroadcastsolutions: nice

dedicatedbroadcastsolutions:
Does anyone think we need greater resolution than seconds for this particular metric?

dimitry-ishenko:
> Does anyone think we need greater resolution than seconds for this particular metric?

IMHO seconds is fine

Issue #1544, opened 2024-06-17
Add ports device number on diag


Sidonai-1:
### Description

Sometimes it's a bit misleading seeing only the main device on the Diag window, which I use to do all the troubleshooting.

I wonder if it would be doable to add the port device number next to the main one.

I'm guessing adding all the device names would be too much in some scenarios, so something like this should suffice:

![WhatsApp Image 2024-06-17 at 12 11 17](https://github.com/CasparCG/server/assets/58334180/e2ccf654-6fe8-4804-9ff3-6d6ad1c4a968)

Issue #1361, opened 2021-01-13
Expose all the graph statistics via network not only in the diag window


rubu:
### Description

Due to leaks and sync issues it would be very interesting to obtain the statistics show in the diagnostics window for aggregation and inspection. Currently AMCP has some stats on the GPU usage, OSC exposes rates for channels, but stuff like dropped frames and consume times are not exposed.

### Solution suggestion
I could implement this, but since I don't want to live off a fork all the time, my question is what would be the preferred way, so that this stuff can be accepted via a Pull Request? I see the following options:

1) Add an AMCP command that returns the values from the last tick. Kind of bad that this would mean polling, but it would live off an existing mechanism
2) Add support for this in OSC. I noticed that current OSC implementation does not use proper NTP timestamps but some kind of ticks, which I would need to fix then. And UDP being unrealiable, but again an already existing mechanism
3) A new transport, I guess TCP based, maybe compatible with Prometheus or any other popular monitoring software?

---

### Additional information
If you have some additional information or attachments.

hreinnbeck:
Note: If done via OSC it should be configured in the config file and disabled by default.

Julusian:
I do not like 1. I don't think it would be able to give a good set of data if only returning the last tick due to command queueing.
3 adds quite a bit of complexity, and another library. While the idea of prometheus metrics does appeal to me, but I don't think the pull model is best suited here and not everyone wants to run prometheus.

2 is the most appealing, and is basically what we used to have in 2.1. I don't remember why that went away.
UDP being unreliable is probably not something worth worrying about. Just make sure that casparcg and the receiver are close together on a reliable network and all will be fine. Many people and hardware devices use osc over wifi for sending commands without issue.

With osc being such a common protocol, it will be easy to write a standalone process in any language to convert the osc data into a more useful format such as prometheus

As hreinnbeck says, it should be an opt in thing. Probably opt-in per osc target, as most will not care for the information.

rubu:
@jamesdwise thanks, yeah 2. also did sound as as the best choice to me, so will spare some time and start working.

Issue #1310, opened 2020-06-08
Diag Window Kills Performance


ronag:
Having the diag window significantly affects server performance.

hreinnbeck:
It sure does.  I'd say that diagnostics in general should be configurable.

ronag:
It shouldn't be that bad. It worked well in the past. Unsure what happened. Suspect a live loop somewhere.

jesperstarkar:
It's always killed CPU performance. I also see som 20mbps network traffic. Guess that is OSC data being sent blindly over UDP?

hreinnbeck:
I do have an old patch laying around that made most of diagnostics and all of osc configurable (down to the number of audio channels, if someone wanted that data).

ronag:
> It's always killed CPU performance.

No it hasn't. Was fine back in 2.0 days.

jesperstarkar:
Then we don't agree.

dimitry-ishenko:
Same reason I never have it open during live shows (running on Linux)

dimitry-ishenko:
Wasn't there a proposal at one point to make it into a standalone app?

ronag:
@jesperstarkar 

> Then we don't agree.

Look at the far right of the graphics:

2.3

<img width="404" alt="Screen Shot 2020-06-09 at 00 58 45" src="https://user-images.githubusercontent.com/3065230/84088556-0980ca00-a9ed-11ea-874f-d281201db821.png">

2.0

<img width="408" alt="Screen Shot 2020-06-09 at 01 02 39" src="https://user-images.githubusercontent.com/3065230/84088563-0ede1480-a9ed-11ea-866a-27a6967521c9.png">

baltedewit:
> It's always killed CPU performance

Judging from the graphs that @ronag shared it's a lot more now (not to say it didn't snoop cpu before). I was running performance tests earlier with diag open and it didn't seem to be a big issue for me. If I can find the time I'll try to run some tests as comparison.

Perhaps now is a good time to think about planing https://github.com/CasparCG/server/issues/899 to reduce future maintenance? I know it's not in the same scope as 2.3.1, but still good to bring to the table?

baltedewit:
I've failed to reproduce this running 4x 1080i50 with an ffmpeg producer and decklink consumer. Opening DIAG influences my CPU usage by at most ~5%. This is on an e5-2670 so hardly a very strong CPU.

ronag:
[...]

PR #1807, opened 2026-09-21
feat(diag): browser-based diagnostics viewer over HTTP/SSE

### LINKED ISSUE #899: Replace DIAG window with external app
As suggested. It should be possible to implement DIAG externally based on OSC. Reduces server complexity. I like it!

dedicatedbroadcastsolutions:
Prototype of a lightweight alternative to the SFML DIAG window: a small opt-in HTTP+SSE server that mirrors every `caspar::diagnostics::graph`'s text/values/colors/tags into JSON, with no rendering cost on the server.

- `GET /` - a small `<canvas>`-based viewer (all rendering client-side)
- `GET /snapshot` - one-shot JSON for simple polling clients
- `GET /events` - Server-Sent Events stream for clients that want to subscribe and do their own rendering/processing
- Enabled by default at `configuration.diag.default-port` (4000), matching how OSC's client is always active; `0` disables it
- Tracks graphs by a stable id (not registry position), shows channel/layer context, renders per-tag severity, and boosts the visibility of dropped/late-frame style series over their (often very dark) SFML colors

Closes #899, #1310 (replaces the SFML DIAG window entirely, so it no longer competes for GPU/desktop resources). Related to #1361 (network-exposed graph stats — done over HTTP/SSE here rather than OSC), #1544 (per-port device labeling — not yet implemented, still channel/layer only), #1139 / #1654 (broader websocket control/monitoring proposals — this is a lighter-weight alternative for the monitoring half), #914 (motivation: avoiding desktop-GUI overhead on the server).

**Status: draft.** Core/protocol and the full executable build cleanly (confirmed both isolated and end-to-end), but I have not yet run this against a live playout server or exercised it under real load - want to validate the viewer/SSE behavior for real before marking ready for review.


dimitry-ishenko:
As much as I love the idea of ditching the DIAG window, I don't think bolting a web server onto CasparCG Server is the right direction.

I like the idea of spitting out diagnostics through OSC via an opt-in option. That way, people can use their preferred GUIs to show that data in their preferred way. And those who don't use it won't be bothered by it.

I go by the motto "_the best code is no code at all_", so the more "features" we can strip out of the core Server code, the better. IMHO, of course.

dedicatedbroadcastsolutions:
Ok, I can shift it towards that direction if we want to do it via OSC.  Using a browser with SSE shifts the render off with minimal overhead of serving a basic web page.  

Since X is going away, I am not sure we can call diag replaced without having a UI for it shipped with the server, but I can make a simple UI that is a separate light weight web server executable that can provide that functionality for those who need it while keeping with your motto for the server core.

